PR #17, opened 2021-10-28
Add graphql server

dirkmc:
```
go run gql.go
```
Open http://localhost:8080

Screen Shot 2021-10-28 at 4 46 39 PM


nonsense (approved):
We need to incorporate this directly in the `boost` service when we instantiate a node, but for now this is great!

PR #20, opened 2021-11-05
Refactor storage provider

dirkmc:
Add a struct to keep track of deal execution (instead of passing around parameters)

Depends on https://github.com/filecoin-project/boost/pull/19


storagemarket/provider.go
```diff
@@ -54,29 +57,65 @@ type Provider struct {
...
+		Address:     addr,
+		newDealPS:   newDealPS,
+		fs:          fs,
 		fullnodeApi: fullnodeApi,
-		db:          db.NewDealsDB(sqldb),
+		db:          dealsDB,
+
+		acceptDealsChan:  make(chan acceptDealReq, 128),
+		failedDealsChan:  make(chan failedDealReq, 128),
+		restartDealsChan: make(chan restartReq, 128),
```
nonsense, line 94: @dirkmc @aarshkshah1992 i think this is dangerous - if we have a bug in our code, we won't notice a deadlock, until we hit 128 deals... this channel should be unbuffered IMO as a start... also why 128 and not 1 or 10 or 20?

PR #21, opened 2021-11-05
Add graphql subscriptions and mutations

dirkmc:
Depends on https://github.com/filecoin-project/boost/pull/20


gql/server.go
```diff
@@ -58,16 +44,22 @@ func Serve(listenAddr string, dealsDB *db.DealsDB) error {
...
 	}
 
 	// graphQL handler
 	graphQLHandler := graphqlws.NewHandlerFunc(schema, &relay.Handler{Schema: schema})
 
-	fmt.Printf("graphql server listening on %s\n", listenAddr)
-	http.Handle("/graphql", &corsHandler{graphQLHandler})
+	go func() {
+		listenAddr := fmt.Sprintf(":%d", httpPort)
+		fmt.Printf("Graphql server listening on %s\n", listenAddr)
```
nonsense, line 57: Nit, but I think we should default to using `log` instead of `fmt.Print` as much as we can.

nonsense (approved):

PR #22, opened 2021-11-05
Implement dummy deal API

dirkmc:
Depends on https://github.com/filecoin-project/boost/pull/21


cmd/boost/dummydeal.go
```diff
@@ -21,14 +22,17 @@ var dummydealCmd = &cli.Command{
...
+		log.Debug("Make API call to start dummy deal")
+		rej, err := boostApi.MarketDummyDeal(ctx)
 		if err != nil {
-			return xerrors.Errorf("couldnt get boost identity: %w", err)
+			return xerrors.Errorf("creating dummy deal: %w", err)
 		}
 
-		spew.Dump(res)
+		if rej != nil && rej.Reason != "" {
+			fmt.Printf("Dummy deal rejected: %s\n", rej.Reason)
```
nonsense, line 32: Same comment about `fmt.Print` and `log` here, but at this specific place, I don't care so much, because this will likely change in the coming days when we make more progress, so fine to leave as is.

node/impl/boost.go
```diff
@@ -23,9 +41,9 @@ type BoostAPI struct {
 	//LocalStore  *stores.Local
 	//RemoteStore *stores.Remote
 
-	//// Markets
+	// Markets
+	StorageProvider *storagemarket.Provider `optional:"true"`
```
nonsense, line 45:
```suggestion
	StorageProvider *storagemarket.Provider
```

node/impl/boost.go
```diff
@@ -34,6 +52,10 @@ type BoostAPI struct {
 	//Host              host.Host                         `optional:"true"`
 	//DAGStore          *dagstore.DAGStore                `optional:"true"`
 
+	// TODO: Figure out how to start graphql server without it needing
+	// to be a dependency of another fx object
+	GraphqlServer *gql.Server `optional:"true"`
```
nonsense, line 57:
```suggestion
	GraphqlServer *gql.Server
```

nonsense (approved):

PR #26, opened 2021-11-09
ensure deal handler is created before startup / deal accept completes

dirkmc:


dirkmc:
Closing as this approach is hard to reason about

PR #29, opened 2021-11-10
fix loop closure

nonsense:
This PR is fixing a bug with a value used in goroutine, but not captured, as well as fixing a few linter issues.

### DIFF
--- a/storagemarket/deal_acceptance.go
+++ b/storagemarket/deal_acceptance.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 
 	"github.com/filecoin-project/boost/storagemarket/types"
-	cborutil "github.com/filecoin-project/go-cbor-util"
 	"github.com/filecoin-project/go-fil-markets/shared"
 
 	"github.com/filecoin-project/go-state-types/abi"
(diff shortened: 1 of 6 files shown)


storagemarket/deal_acceptance.go
```diff
@@ -149,18 +148,18 @@ func (p *Provider) validateSignature(tok shared.TipSetToken, deal types.Provider
...
+	//verified, err := p.adapter.VerifySignature(p.ctx, deal.ClientDealProposal.ClientSignature, deal.ClientDealProposal.Proposal.Client, b, tok)
+	//if err != nil {
+	//return fmt.Errorf("error verifying signature: %w", err)
+	//}
+	//if !verified {
+	//return errors.New("could not verify signature")
+	//}
 
-	return nil
+	//return nil
```
nonsense, line 164: Nit, but ideally we should keep the linter happy.

storagemarket/provider.go
```diff
@@ -89,9 +89,9 @@ func NewProvider(repoRoot string, dealsDB *db.DealsDB, fullnodeApi v1api.FullNod
 		fullnodeApi: fullnodeApi,
 		db:          dealsDB,
 
-		acceptDealsChan:  make(chan acceptDealReq, 128),
-		failedDealsChan:  make(chan failedDealReq, 128),
-		restartDealsChan: make(chan restartReq, 128),
+		acceptDealsChan:  make(chan acceptDealReq),
+		failedDealsChan:  make(chan failedDealReq),
+		restartDealsChan: make(chan restartReq),
```
nonsense, line 94: Changing these channels from buffered to unbuffered - the rational is - if we have a bug with respect to the synchronicity of how we consume values from these channels, we'll only notice a bug after contention happens on one of them, which might be at high load - so better make the channel unbuffered, and if it is necessary for them to be buffered, we should reduce 128 to a much lower value.

dirkmc (approved):

PR #31, opened 2021-11-15
adding fullnode and miner devnet to support boost itests and development

nonsense:
This PR is adding a devnet command, so that we can easily start a fullnode and a miner, and get `boost` to connect to the miner in order to run end-to-end itests.

### DIFF
--- a/devnet/README.md
+++ b/devnet/README.md
@@ -5,10 +5,18 @@ Runs a Lotus daemon and miner to be used for boost integration tests.
 To run it locally, install ./cmd/devnet and the lotus binaries into your $PATH, and
 simply run one of the integration tests - it should start devnet automatically.
 
-### Building and installing Lotus
+### Building and installing Lotus-related binaries
 
 	git clone https://github.com/filecoin-project/lotus
 	cd lotus
 	git checkout master # or whichever version
 	make debug
-	sudo make install # or "cp lotus lotus-*" into your $PATH
+	sudo make install
+	install -C ./lotus-seed /usr/local/bin/lotus-seed
+
+### Clean up state directories
+
+Make sure your $LOTUS_PATH and $LOTUS_MINER_PATH are clean before running devnet:
+
+	rm -rf ~/.lotus
+	rm -rf ~/.lotusminer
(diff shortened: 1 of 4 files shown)


devnet/README.md
```diff
@@ -0,0 +1,22 @@
...
+	make debug
+	sudo make install
+	install -C ./lotus-seed /usr/local/bin/lotus-seed
+
+### Clean up state directories
+
+Make sure your $LOTUS_PATH and $LOTUS_MINER_PATH are clean before running devnet:
+
+	rm -rf ~/.lotus
+	rm -rf ~/.lotusminer
```
dirkmc, line 22: I wonder if we should use non-default directories for this so people don't accidentally over-write their local state for their production miner

devnet/devnet.go
```diff
@@ -0,0 +1,175 @@
...
+		// This means deals could sit at StorageDealAwaitingPreCommit or
+		// StorageDealSealing for a while, going past our 10m test timeout.
+		{"sed", "-ri", "-e", "s/#BatchPreCommits\\ =\\ true/BatchPreCommits=false/",
+			filepath.Join(home, ".lotusminer", "config.toml")},
+
+		{"sed", "-ri", "-e", "s/#AggregateCommits\\ =\\ true/AggregateCommits=false/",
+			filepath.Join(home, ".lotusminer", "config.toml")},
+
+		{"lotus-miner", "run", "--nosync"},
+	}
```
dirkmc, line 72: Would it make sense to put all these commands in a script file and just run the script file?

devnet/devnet.go
```diff
@@ -0,0 +1,175 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// The parameter files can be as large as 1GiB.
+	// If this is the first time lotus runs,
+	// and the machine doesn't have particularly fast internet,
+	// we don't want devnet to seemingly stall for many minutes.
+	// Instead, show the download progress explicitly.
+	// fetch-params will exit in about a second if all files are up to date.
+	// The command is also pretty verbose, so reduce its verbosity.
```
dirkmc, line 122: Love it ❤️

dirkmc (approved):

PR #34, opened 2021-11-15
Publish deal

dirkmc:
Implement publish deals part of deal execution

### DIFF
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -34,10 +34,10 @@ type Boost struct {
 	SectorIndexApiInfo string
 	Dealmaking         DealmakingConfig
 	DAGStore           DAGStoreConfig
-	Wallets            WalletConfig
+	Wallets            WalletsConfig
 }
 
-type WalletConfig struct {
+type WalletsConfig struct {
 	// The "owner" address of the miner
 	Miner address.Address
 	// The wallet used to send PublishStorageDeals messages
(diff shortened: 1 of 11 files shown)


storagemarket/provider.go
```diff
@@ -82,19 +80,19 @@ func NewProvider(repoRoot string, dealsDB *db.DealsDB, fullnodeApi v1api.FullNod
 
 	return &Provider{
 		// TODO: pass max transfer duration as a param
-		config:      Config{MaxTransferDuration: 30 * time.Second},
-		Address:     addr,
-		newDealPS:   newDealPS,
-		fs:          fs,
-		fullnodeApi: fullnodeApi,
```
nonsense, line 89: I am surprised we don't need this here, but probably because of all the commented-out code right now.

node/config/types.go
```diff
@@ -32,6 +34,14 @@ type Boost struct {
 	SectorIndexApiInfo string
 	Dealmaking         DealmakingConfig
 	DAGStore           DAGStoreConfig
+	Wallets            WalletConfig
+}
+
+type WalletConfig struct {
```
nonsense, line 40: Nit, but probably this should be called `WalletsConfig` as it has config for 2 wallets.

nonsense (approved):

storagemarket/provider.go
dirkmc: Yep we'll need it when we want to watch deals on chain after they've started sealing

PR #37, opened 2021-11-15
Serve react app from GQL server

dirkmc:
Serve react from the GQL server instead of running a separate server


README.md
```diff
@@ -42,6 +42,22 @@ FULLNODE_API_INFO=/ip4/127.0.0.1/tcp/1234/http boost run
...
 ```
 
+## Web UI
+
+1. Install and build
+
+```
+cd react
+npm install
+npm run build
```
jacobheun, line 52: Is there also a `watch` for development mode? Would be great to have it documented if so.

dirkmc: I added a "Development mode" section describing `npm start`

jacobheun (approved):

PR #40, opened 2021-11-16
HTTP Pull Transport

aarshkshah1992:
### TODO

- [x] Solid e2e tests for the transport.


transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,237 @@
...
+	pub, sub, err := createPubSub(eventbus.NewBus())
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to create pub-sub: %w", err)
+	}
+	// emit event with initial state
+	if err := pub.Emit(types.TransportEvent{
+		Status:         transferstatus.Initiated,
+		NBytesReceived: fileSize,
+	}); err != nil {
+		log.Errorw("failed to publish transport event", "dealUuid", dealInfo.DealUuid, "err", err)
```
dirkmc, line 74: In other places in the code when logging we call the field `id` instead of `dealUuid`. Happy to go with either but let's make sure it's consistent so that someone can grep the logs easily

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,237 @@
...
+		nBytesReceived: fileSize,
+		wg:             wg,
+	}
+
+	t.wg.Add(1)
+	go t.execute(tctx)
+	return transport.NewHandler(tctx, cancel, dealInfo, sub, wg), false, nil
+}
+
+type transfer struct {
```
dirkmc, line 97: transfer and TransportHandler share several of the same fields - any reason not to unite them into one struct?

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,237 @@
...
+	tInfo    *pb.HttpReq
+	dealInfo *types.TransportDealInfo
+	pub      event.Emitter
+	wg       sync.WaitGroup
+
+	nBytesReceived int64
+}
+
+func (t *transfer) execute(ctx context.Context) {
+	defer t.wg.Done()
```
dirkmc, line 107: nit: I'd suggest putting `t.wg.Done()` in a go-routine that gets called on line 86. That way it's easy to see the purpose and lifecycle of the WaitGroup as the `Add` and `Done` methods are called near each other

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+
+	failTransfer := func(err error) {
```
dirkmc, line 109: Instead of explicitly calling the failTransfer method whenever there's an error, I'd suggest a more idiomatic approach would be to
- return error from this method
- in the method that calls this one
  - check if there was an error
  - if so call failTransfer

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (h *httpTransport) Execute(ctx context.Context, transportInfo []byte, dealInfo *types.TransportDealInfo) (th transport.Handler, isComplete bool, err error) {
+
+	// de-serialize transport opaque token
+	tInfo := &pb.HttpReq{}
+	if err := proto.Unmarshal(transportInfo, tInfo); err != nil {
+		return nil, false, fmt.Errorf("failed to de-serialize transport info bytes: %w", err)
+	}
+	// check that the outputFile exists
+	fi, err := os.Stat(dealInfo.OutputFile)
```
dirkmc, line 45: Would it make sense to just create the file if it doesn't exist here?

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// construct request
+	req, err := http.NewRequest("GET", t.tInfo.URL, nil)
+	if err != nil {
+		failTransfer(fmt.Errorf("failed to create http req: %w", err))
+		return
+	}
+
+	// add range req to start reading from the last byte we have in the output file
+	fi, err := os.Stat(t.dealInfo.OutputFile)
```
dirkmc, line 127: We already got the file size on line 54, we can just reuse it here instead of calling stat again

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+func (t *transfer) doHttp(req *http.Request, dst io.Writer, toRead int64) error {
+	// send http request and validate response
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to send  http req: %w", err)
+	}
+	// we should either get back a 200 or a 206 -> anything else means something has gone wrong and we return an error.
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
+		resp.Body.Close() // nolint
+		return fmt.Errorf("http req failed: %d", resp.StatusCode)
```
dirkmc, line 189: Is there anything else we can include in this error message, eg the status text? Probably connecting to the server is going to be the point where we'll have the most errors, so it would be helpful if we can surface as much information as possible to the user about why the transfer fails.

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	defer resp.Body.Close()
```
dirkmc, line 191: nit: move this `defer resp.Body.Close()` to line 186 and remove `defer resp.Body.Close()` on line 188

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+
+	//  start reading the response stream `readBufferSize` at a time using a limit reader so we only read as many bytes as we need to.
```
dirkmc, line 193: Good idea to use a limit reader 👍

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+		// if we read more than zero bytes, write whatever read.
+		if nr > 0 {
+			nw, writeErr := dst.Write(buf[0:nr])
+
+			// if the number of read and written bytes don't match -> something has gone wrong, abort the http req.
+			if nw < 0 || nr != nw {
+				if writeErr != nil {
+					return fmt.Errorf("failed to write to output file: %w", err)
+				}
+				return fmt.Errorf("read-write mismatch, read=%d, written=%d", nr, nw)
```
dirkmc, line 208: nit: make clear this error arises when writing the file (not when reading from the HTTP stream)

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/filecoin-project/boost/transport/types"
+	"github.com/gogo/protobuf/proto"
+
+	"github.com/filecoin-project/boost/transport"
+)
+
+var log = logging.Logger("http-transport")
+
+const (
+	readBufferSize = 4096
```
dirkmc, line 28: For efficiency we may want to make `readBufferSize` bigger than 4096. I'm not sure what the standard is here though.

dirkmc (commented):
I’ve left several specific comments on the PR, and I also have some general comments:

1. Could you please integrate this transport with the existing code in deal_execution.go
I think it will help clarify the design choices and interfaces. For example  `Execute` returns an `isComplete` boolean, but we also have a handler that returns events with an `IsTerminalState` boolean. I’d like to see how this works in the calling code.

2. Can we simplify TransportEvent?
It may be possible to remove the Status field. For example I don’t think we need the Initiated state: if the number of bytes transferred is zero, we know the transfer hasn’t started. Similarly if the Error field is set then we know the transfer has errored, so we may not need the Errored field

3. We are using protobuf to serialize the URL field. There is only one field in that struct so far, so I assume you’re planning to add other fields (eg for auth headers). If that’s the case, I’d suggest using JSON instead of protobuf, because it’s human readable so it’s easier to debug when things go wrong.

transport/httptransport/http_transport.go
aarshkshah1992: @dirkmc This method executes in a go-routine.

transport/httptransport/http_transport.go
aarshkshah1992: I've included the full status string here as well not sure what else we can include. Note that you hit this code only when you actually send a request and get back a status code.

If there's some error connecting to the server, we do return the entire error in the code above this snippet.

transport/httptransport/http_transport.go
aarshkshah1992: I strongly disagree. We want the caller to control the entire lifecycle of the output file and don't want that to get fragmented between the Boost SM & the transport. If down the road, it turns out that we do need this, we can add a config option but let's avoid for now.

dirkmc: Ok sounds good 👍

transport/handler.go
```diff
@@ -0,0 +1,50 @@
...
+		ctx:      ctx,
+		cancel:   cancel,
+		dealInfo: dInfo,
+		Sub:      sub,
+		wg:       wg,
+	}
+}
+
+// Close shuts down the transfer for the given deal. It is the caller's responsibility to call Close after it no longer needs the transfer.
+func (t *TransportHandler) Close() error {
```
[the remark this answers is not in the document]
aarshkshah1992: This needs to satisfy the Transport `Handler` interface and some transports might want to return an error when they are shut down. Let's keep it for now.

transport/httptransport/http_transport.go
dirkmc: Ok great 👍

transport/handler.go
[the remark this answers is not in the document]
dirkmc: Ok makes sense 👍

transport/httptransport/http_transport.go
aarshkshah1992: It is now 64K. Let's iterate if needed later.

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+			Status:          transferstatus.Errored,
+			IsTerminalState: true,
+			Error:           err,
+		}); err != nil {
+			log.Errorw("failed to publish transport error", "dealUuid", t.dealInfo.DealUuid, "err", err)
+		}
+	}
+
+	// construct request
+	req, err := http.NewRequest("GET", t.tInfo.URL, nil)
```
willscott, line 120: can there be a http.Client / http.roundtripper state used for these that's set up for the transfer struct, rather than using the implicit global one?

if we want to support outbound dialing through proxies, custom timeouts, etc we'll want that flexibility

transport/interface.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+
+	"github.com/libp2p/go-libp2p-core/event"
+
+	"github.com/filecoin-project/boost/transport/types"
+)
+
+type Transport interface {
+	Execute(ctx context.Context, transportInfo []byte, dealInfo *types.TransportDealInfo) (th Handler, isComplete bool, err error)
```
willscott, line 12: can we get comments through this interface. it's hard to understand exactly how to plug in other implementations given this interface.

transport/types/types.go
```diff
@@ -0,0 +1,19 @@
+package types
+
+import (
+	"github.com/filecoin-project/boost/transport/types/transferstatus"
+	"github.com/google/uuid"
+)
+
+type TransportDealInfo struct {
```
willscott, line 8: where will this be set up? how is output file name chosen? will the file name need to be sanitized?

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,241 @@
...
+	readBufferSize = 4096
+)
+
+var _ transport.Transport = (*httpTransport)(nil)
+
+type httpTransport struct {
+	// TODO Construct and inject an http client here ?
+}
+
+func (h *httpTransport) Execute(ctx context.Context, transportInfo []byte, dealInfo *types.TransportDealInfo) (th transport.Handler, isComplete bool, err error) {
```
willscott, line 37: what lets us signal which transport to use? this is assuming that transportInfo already knows it's using HTTP transport, so is there signalling a layer up of which transport to use?

storagemarket/deal_execution.go
```diff
@@ -103,33 +113,39 @@ func (p *Provider) transferAndVerify(ctx context.Context, pub event.Emitter, dea
...
+		OutputFile: deal.InboundFilePath,
+		DealUuid:   deal.DealUuid,
+		// TODO Is this the correct size of the actual data without padding ?
+		// I think it's best to explicitly add a "deal size" param to the deal negotiation.
+		DealSize: int64(deal.ClientDealProposal.Proposal.PieceSize),
+	})
 	if err != nil {
 		return fmt.Errorf("failed data transfer: %w", err)
 	}
+	defer handler.Close()
```
dirkmc, line 126: nit: suggest refactoring this loop into a function so you can call `defer handler.Close()` at the beginning of the function. At the moment it's a little tricky to follow why handler.Close() is called in some places and not others (eg on line 143, but not on line 137)

storagemarket/types/deal_state.go
```diff
@@ -50,4 +50,7 @@ type ProviderDealState struct {
 	Checkpoint dealcheckpoints.Checkpoint
 	// set if there's an error
 	Err string
+
+	// NBytesReceived is the number of bytes Received for this deal
+	NBytesReceived int64
```
dirkmc, line 55: We have a ProviderDealInfo object:
```
type ProviderDealInfo struct {
	Deal        *ProviderDealState
	Transferred uint64
}
```

If we want to persist NBytesReceived let's keep it here and get rid of the ProviderDealInfo struct.

If we don't want to persist NBytesReceived let's remove it from here and use ProviderDealInfo instead.

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,235 @@
...
+
+type httpTransport struct {
+	// TODO Construct and inject an http client here ?
+}
+
+func (h *httpTransport) Execute(ctx context.Context, transportInfo []byte, dealInfo *types.TransportDealInfo) (th transport.Handler, err error) {
+	// de-serialize transport opaque token
+	tInfo := &types.HttpRequest{}
+	if err := json.Unmarshal(transportInfo, tInfo); err != nil {
+		return nil, fmt.Errorf("failed to de-serialize transport info bytes: %w", err)
```
dirkmc, line 43: nit: include the bytes as a string in the error message to make it easier to debug

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,235 @@
...
+	if _, err := url.Parse(tInfo.URL); err != nil {
+		return nil, fmt.Errorf("failed to parse url: %w", err)
+	}
+	// do we have more bytes than required already ?
+	fileSize := fi.Size()
+	if fileSize > dealInfo.DealSize {
+		return nil, fmt.Errorf("deal size=%d but file size=%d", dealInfo.DealSize, fileSize)
+	}
+
+	// construct the transfer instance that will act has the transfer handler
```
dirkmc, line 60:
```suggestion
	// construct the transfer instance that will act as the transfer handler
```

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,235 @@
...
+	tctx, cancel := context.WithCancel(ctx)
+	t := &transfer{
+		cancel:         cancel,
+		tInfo:          tInfo,
+		dealInfo:       dealInfo,
+		eventCh:        make(chan types.TransportEvent, 256),
```
dirkmc, line 66: I think the eventCh should either be unbuffered, or we should have a mechanism that only keeps the latest event (over-writes earlier events)

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,235 @@
...
+		nBytesReceived: fileSize,
+	}
+
+	// is the transfer already complete ? we check this by comparing the number of bytes
+	// in the output file with the deal size.
+	if fileSize == dealInfo.DealSize {
+		defer close(t.eventCh)
```
dirkmc, line 73: Should we also call cancel() here? Alternatively, should we call t.Close()?

transport/httptransport/http_transport.go
```diff
@@ -0,0 +1,235 @@
...
+
+		if err := t.emitEvent(tctx, types.TransportEvent{
+			NBytesReceived: fileSize,
+		}, dealInfo.DealUuid); err != nil {
+			return nil, fmt.Errorf("failed to publish transfer completion event, id: %s, err: %w", t.dealInfo.DealUuid, err)
```
dirkmc, line 78: nit: suggest moving the
```fmt.Errorf("failed to publish transfer event, id: %s, err: %w", t.dealInfo.DealUuid, err)```
to `t.emitEvent` so you can just do
```
if err := t.emitEvent(tctx, types.TransportEvent{NBytesReceived: fileSize}, dealInfo.DealUuid); err != nil {
			return nil, err
```

File: transport/httptransport/http_transport.go [hunk 24]
```diff
@@ -0,0 +1,235 @@
...
+	}, t.dealInfo.DealUuid); err != nil {
+		log.Errorw("failed to publish transport event", "id", t.dealInfo.DealUuid, "err", err)
+	}
+
+	return nil
+}
+
+func (t *transfer) doHttp(ctx context.Context, req *http.Request, dst io.Writer, toRead int64) error {
+	// send http request and validate response
+	resp, err := http.DefaultClient.Do(req)
```
dirkmc, line 172: I wonder if there's some way to pass a context here 🤔 
If it takes a long time to connect to the HTTP server (eg because it's behind a firewall and hangs forever) it seems like there's no way to cancel the request.

transport/interface.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/filecoin-project/boost/transport/types"
+)
+
+type Transport interface {
+	Execute(ctx context.Context, transportInfo []byte, dealInfo *types.TransportDealInfo) (th Handler, err error)
+}
+
+type Handler interface {
+	Sub() chan types.TransportEvent
+	Close() error
```
dirkmc, line 15: It looks from the code like we ignore the error return value of `Close() error` so I think we should just remove it

transport/httptransport/http_transport.go
aarshkshah1992: We already do that in:

```
 // init the request with the transfer context
	req = req.WithContext(ctx)
```

transport/interface.go
aarshkshah1992: Hey, we already discussed this in https://github.com/filecoin-project/boost/pull/40#discussion_r750179349

transport/httptransport/http_transport.go
aarshkshah1992: We don't need to call Close as the transfer go-routine hasn't started yet. This check is done when the user calls `Execute` and want to ensure that the file dosen't already have as many bytes as the user needs.

dirkmc: Ok makes sense, I think we still want to call cancel() here though

transport/httptransport/http_transport.go
dirkmc: Ah ok makes sense 👍

transport/interface.go
dirkmc: Yeah I know, but we're not actually using the return value in the calling code, AFAICT. So if we're not using the return value, and we're not returning anything from the existing transport implementation, I think we should just remove it.
If in future we need it for a different transport implementation, we can add it in again.

If you feel strongly about it I'm fine keeping it here.

transport/httptransport/http_transport.go
aarshkshah1992: I disagree. The client should get some leeway in terms of consuming it without blocking the transfer.  I don't think over-writes make sense at this layer as these events are simply append only logs of what happened with the transfer. You can do it at the UX layer if you want.

dirkmc: In this case the process that reads from the channel is the provider, which reads in a for loop that just passes the event to pubsub. The process writing to the channel is a loop which reads from i/o. So I would expect the process writing to the channel to be much slower. In any case even if the process writing to the channel does get blocked momentarily, I don't believe it matters because the i/o it reads from is being buffered by the OS.

storagemarket/provider.go
```diff
@@ -101,6 +103,14 @@ func NewProvider(repoRoot string, dealsDB *db.DealsDB, fullnodeApi v1api.FullNod
 	}, nil
 }
 
+func (p *Provider) NBytesReceived(deal *types.ProviderDealState) (int64, error) {
+	fi, err := os.Stat(deal.InboundFilePath)
+	if err != nil {
+		return 0, fmt.Errorf("failed to stat file: %w", err)
+	}
+	return fi.Size(), nil
+}
```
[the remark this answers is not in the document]
aarshkshah1992: Hey @dirkmc  -> Let's address this in a follow up PR as part of the Boost SM refactor. I understand the requirements here.

transport/httptransport/http_transport.go
aarshkshah1992: Good catch.

transport/httptransport/http_transport.go
aarshkshah1992: I don't have strong opinions here in the absence of having run this in the wild. Happy to make it unbuffered and see how it goes in practise.

transport/interface.go
aarshkshah1992: I'm okay either ways if you feel so strongly about this. Removed.

transport/httptransport/http_transport.go
dirkmc: Ok sounds good 👍

storagemarket/deal_execution.go
```diff
@@ -51,6 +54,15 @@ func (p *Provider) doDeal(deal *types.ProviderDealState) {
...
 	})
 
+	// build in-memory state
+	fi, err := os.Stat(deal.InboundFilePath)
+	if err != nil {
+		err := fmt.Errorf("failed to stat output file: %w", err)
+		p.failDeal(pub, deal, err)
+		return
+	}
+	deal.NBytesReceived = fi.Size()
```
willscott, line 64: it would be great to stream the inbound car through the car-v2 library to build up an index as we're already stream it. that could mean you end up with the file on disk at the end bigger though. can we have this in the interface rather than the file-on-disk?

gql/resolver.go
```diff
@@ -277,10 +283,10 @@ type dealResolver struct {
 	dealsDB     *db.DealsDB
 }
 
-func newDealResolver(deal *types.ProviderDealInfo, dealsDB *db.DealsDB) *dealResolver {
+func newDealResolver(deal *types.ProviderDealState, dealsDB *db.DealsDB) *dealResolver {
 	return &dealResolver{
-		ProviderDealState: *deal.Deal,
-		transferred:       deal.Transferred,
+		ProviderDealState: *deal,
+		transferred:       uint64(deal.NBytesReceived),
```
dirkmc, line 289: In a follow-up PR we can remove the `transferred` field from `dealResolver`

storagemarket/deal_execution.go
```diff
@@ -146,6 +147,30 @@ func (p *Provider) transferAndVerify(ctx context.Context, pub event.Emitter, dea
...
 
+func (p *Provider) waitForTransferFinish(ctx context.Context, handler transport.Handler, pub event.Emitter, deal *types.ProviderDealState) error {
+	defer handler.Close()
+
+loop:
+	for {
+		select {
+		case evt, ok := <-handler.Sub():
+			if !ok {
+				break loop
```
dirkmc, line 158: nit: you can just return here

storagemarket/deal_execution.go
```diff
@@ -146,6 +147,30 @@ func (p *Provider) transferAndVerify(ctx context.Context, pub event.Emitter, dea
...
+			}
+			if evt.Error != nil {
+				return fmt.Errorf("data-transfer failed: %w", evt.Error)
```
dirkmc, line 161: nit: you already moved "data-transfer failed" to the calling code so I think you can just directly return the error here and on line 167 below

dirkmc (approved):

transport/httptransport/http_transport.go
aarshkshah1992: Actually, the consumer i.e. Boost SM here reads these transfer events and then pushes them further upwards after transforming there and could be running into back-pressure there or maybe the transformation takes time. I don't want that to block transfers. Remember, the transfer is running a single go-routine and so the transfer will get blocked if it gets blocked writing to the channel here.  We don't need strict synchronisation between the transport and Boot SM here. These notifications are just the append-only log of what's happening with the transfer, NOT used for synchronisation between the transport and Boost SM. Can look into making this a non-blocking queue if you feel strongly against using a buffered channel here.

aarshkshah1992: I'd like to see how this works in practise before we decide to change this.

storagemarket/deal_execution.go
aarshkshah1992: @willscott If the user already has a CARv2 file, just streaming the file over HTTP should get us the Index too, right ? Ah, but you are saying that the Index is never part of the deal size.

willscott: right. it's probably safer if the storage provider re-validates/generates the index as it pulls it, but the commp will need to be over the carv1 subset, so we need to account for that balance somehow

storagemarket/deal_execution.go
aarshkshah1992: done.

storagemarket/deal_execution.go
aarshkshah1992: done.

transport/httptransport/http_transport.go
aarshkshah1992: Agreed to punt this discussion as Boost evolves.

Issue #33, opened 2021-11-15
Pass wallet addresses to `boost init`


dirkmc:
To run boost the user needs to have two wallets:
- One for publish storage deals messages
- One for pledging collateral

These wallets should be passed as required parameters to `boost init`.
We should check that the wallets are distinct.
We should store the wallet addresses in config.toml (rather than in the datastore) so that it's easier for the user to see and update their configuration.

PR #45, opened 2021-11-16
add two wallets as required when initialising boost

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/33

TODO:
- [x] update README.md for starting `devnet` with exact commands to create a second wallet, send funds from first wallet to second (so that both wallets have funds) and pass these wallets to `boost init`

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,11 @@ rm -rf ~/.lotus ~/.lotusminer
 2. Create Boost repository
 
 ```
-FULLNODE_API_INFO=/ip4/127.0.0.1/tcp/1234/http boost --vv init --api-sector-index=`lotus-miner auth api-info --perm=admin`
+FULLNODE_API_INFO=/ip4/127.0.0.1/tcp/1234/http \
+boost --vv init \
+      --api-sector-index=`lotus-miner auth api-info --perm=admin` \
+      --wallet-publish-storage-deals=`lotus wallet new bls` \
+      --wallet-collateral-pledge=`lotus wallet new bls`
 ```
 
 3. Run Boost service
--- a/cmd/boost/init.go
+++ b/cmd/boost/init.go
@@ -8,6 +8,7 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/boost/node/repo"
+	"github.com/filecoin-project/go-address"
 
 	cliutil "github.com/filecoin-project/boost/cli/util"
 	"github.com/filecoin-project/boost/node/config"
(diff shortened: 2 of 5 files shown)


cmd/boost/init.go
```diff
@@ -35,6 +46,28 @@ var initCmd = &cli.Command{
 
 		log.Debug("Trying to connect to full node RPC")
 
+		var walletPSD address.Address
+		if wPSD := cctx.String("wallet-publish-storage-deals"); wPSD != "" {
+			var err error
+			walletPSD, err = address.NewFromString(wPSD)
+			if err != nil {
+				return err
```
dirkmc, line 54: nit: it would be nice to return a descriptive error here so they know which wallet address failed to parse, eg "failed to parse wallet address publish storage deals '%s': %w"

cmd/boost/init.go
```diff
@@ -35,6 +46,28 @@ var initCmd = &cli.Command{
 
 		log.Debug("Trying to connect to full node RPC")
 
+		var walletPSD address.Address
+		if wPSD := cctx.String("wallet-publish-storage-deals"); wPSD != "" {
```
dirkmc, line 50: nit: This is a required field so `wPSD != ""` should always be true, right?

node/builder.go
```diff
@@ -351,6 +352,15 @@ func ConfigBoost(c interface{}) Option {
 		return Error(xerrors.New("retrieval pricing policy must be either default or external"))
 	}
 
+	walletPSD, err := address.NewFromString(cfg.Wallets.PublishStorageDeals)
+	if err != nil {
+		return Error(err)
```
dirkmc, line 357: Same here, let's be descriptive in the error messages to help with debugging

README.md
```diff
@@ -38,7 +38,11 @@ rm -rf ~/.lotus ~/.lotusminer
 2. Create Boost repository
 
 ```
-FULLNODE_API_INFO=/ip4/127.0.0.1/tcp/1234/http boost --vv init --api-sector-index=`lotus-miner auth api-info --perm=admin`
+FULLNODE_API_INFO=/ip4/127.0.0.1/tcp/1234/http \
+boost --vv init \
+      --api-sector-index=`lotus-miner auth api-info --perm=admin` \
+      --wallet-publish-storage-deals=`lotus wallet new bls` \
+      --wallet-collateral-pledge=`lotus wallet new bls`
```
nonsense, line 45: We'll need to feed these wallets with FIL later on in `boost` in order to attempt deals, but for now this is fine.

dirkmc (approved):

PR #50, opened 2021-11-17
chore: port over some labels from Lotus

jacobheun:
This copies over labels from Lotus to help with organizing a bit better.
I've simplified the area labels and made them contextual to Boost.

I started setting up some MIlestones and wanted to add some additional issues for
tracking based on what we're shooting for in those milestones, so having these would
be nice for helping to organize a bit more.

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -61,13 +61,13 @@
 #
 - name: dif/trivial
   color: b2b7ff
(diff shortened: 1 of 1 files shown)


dirkmc (approved):
There are a couple of references to lotus in the descriptions, but not a biggie

jacobheun:
Nice catch, replaced lotus with Boost

Issue #32, opened 2021-11-15
Update `devnet` to not use default LOTUS environment variables and paths


nonsense:
At the moment if you use the `devnet` binary, or if you run integration tests, they will create artefacts at various places under the $HOME directory of the user.

Ideally we should be creating artefacts under a temporary directory that `devnet` creates and then do a cleanup after we shut the devnet down.

PR #55, opened 2021-11-17
initial dummydeal itest for boost with a devnet

nonsense:
TODO:
- [x] instantiate test boost node connected to fullnode and miner launched by the devnet
- [x] make sure TestDummydeal passes
- [x] add `go test ./...` to our CI
- [ ] get `lotus`, `lotus-seed`, `lotus-miner` in the CI PATH
- [x] fix relative/absolute path for gql
- [ ] set `HOME` to a temporary dir... or fix https://github.com/filecoin-project/boost/issues/32 - so that we avoid having to do `rm -rf ~/.lotus ~/.lotusminer`

### DIFF
--- a/gql/server.go
+++ b/gql/server.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"path/filepath"
+	"runtime"
 
 	logging "github.com/ipfs/go-log/v2"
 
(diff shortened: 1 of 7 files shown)


nonsense:
@dirkmc @aarshkshah1992 marking this as ready for review, as it will take a bit of time to setup CI properly, as we need lotus binaries in CI as well. Let's review the rest in the meantime.

gql/server.go
```diff
@@ -41,7 +41,7 @@ func (s *Server) Serve(ctx context.Context) error {
 	http.HandleFunc("/graphiql", graphiql(httpPort))
 
 	// Init graphQL schema
-	schemaText, err := ioutil.ReadFile("gql/schema.graphql")
+	schemaText, err := ioutil.ReadFile("/Users/nonsense/code/src/github.com/filecoin-project/boost/gql/schema.graphql")
```
nonsense, line 44: We have to see how to address this.

dirkmc: I think you can do something like
```
_, currentDir, _, ok := runtime.Caller(1)
filepath := path.Join(path.Dir(currentDir), "schema.graphql")
```

dirkmc (approved):

Issue #59, opened 2021-11-18
Implement manual dummy deal


jacobheun:
- Update the dummy deal CLI to create a signed deal proposal (currently the deal proposal is not signed)
- Uncomment the code in the deals state manager that verifies the signature

Fixed by #63

PR #63, opened 2021-11-18
Uncomment signature verification from deal execution

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/59

And modify dummy deal to create a proper signature

Note: There is a TODO in here that can be removed once https://github.com/filecoin-project/boost/pull/45 is merged

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -1,0 +1,47 @@
+package api
+
+import (
+	"context"
+
+	"github.com/filecoin-project/lotus/chain/types"
+
+	"github.com/filecoin-project/go-address"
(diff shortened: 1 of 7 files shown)


api/wallet.go
```diff
@@ -0,0 +1,47 @@
+package api
```
dirkmc, line 1: Note: All this wallet code is just copied across from lotus

dirkmc:
Closing this PR because
- it includes a bunch of wallet code that we don't need - we can just call the lotus wallet API
- it creates a wallet for each client in dummy deal - instead we should just implement the client side of deal making

PR #64, opened 2021-11-19
enable go test in CI

nonsense:
This PR is adding `boost` integration tests as part of CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,9 +56,9 @@ commands:
       - restore_cache:
           name: Restore lotus artefacts cache
           keys:
-            - 'lotus-v1.13.1-rc.2'
+            - 'lotus-v1.13.1-rc.22'
           paths:
-            - /go/bin/
+            - /tmp/lotus-artefacts
       - setup_remote_docker:
           version: 19.03.13
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -37,6 +37,46 @@ commands:
...
+          keys:
+            - 'lotus-v1.13.1-rc.22'
+          paths:
+            - /tmp/lotus-artefacts
+      - setup_remote_docker:
+          version: 19.03.13
+      - run:
+          name: copy latest stable lotus binaries
+          command: |
+            docker pull nonsens3/lotus:v1.13.1-rc2
```
nonsense, line 67: We need docs on this, given that there is no existing image we can leverage... it will be up to us to update the version of Lotus periodically.

.circleci/config.yml
```diff
@@ -37,6 +37,46 @@ commands:
...
+            - 'v25-2k-lotus-params'
+          paths:
+            - /var/tmp/filecoin-proof-parameters/
+      - run:  lotus fetch-params 2048
+      - save_cache:
+          name: Save parameters cache
+          key: 'v25-2k-lotus-params'
+          paths:
+            - /var/tmp/filecoin-proof-parameters/
+  lotus-artefacts:
```
nonsense, line 54: This part could use a bit of love. We do load /tmp/lotus-artefacts , but we are also running all the docker part afterwards, which is redundant.

dirkmc: Fair enough but I think this is a great start, so happy to do that later

itests/dummydeal_test.go
```diff
@@ -27,19 +28,37 @@ func init() {
 
 func TestDummydeal(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 
 	done := make(chan struct{})
 	go devnet.Run(ctx, done)
 
-	//TODO: detect properly when devnet (daemon+miner) are ready to serve requests
-	time.Sleep(45 * time.Second)
```
dirkmc, line 35: Great we could get rid of this :)

dirkmc (approved):

PR #65, opened 2021-11-19
Fund Manager

dirkmc:
TODO
- [x] Refactor DB code to accommodate several tables (funds tables as well as deals tables)
- [ ] More tests for FundManager class
- [ ] Add indexes to DB tables
- [ ] Add Fund Manager config to config.toml


db/db.go
```diff
@@ -4,332 +4,29 @@ import (
 	"context"
```
dirkmc, line 4: You can ignore the changes here and in db_deals.go. I just refactored the code into two separate files.

db/funds.go
```diff
@@ -0,0 +1,93 @@
...
+type FundsDB struct {
+	db *sql.DB
+}
+
+func NewFundsDB(db *sql.DB) *FundsDB {
+	return &FundsDB{db: db}
+}
+
+func (f *FundsDB) Tag(ctx context.Context, dealUuid uuid.UUID, collateral abi.TokenAmount, pubMsg abi.TokenAmount) interface{} {
+	qry := "INSERT INTO FundsTagged (DealID, CreatedAt, Collateral, PubMsg) "
```
nonsense, line 31: Nit, but we might want to rename the field from `DealID` to `DealUUID` just so that we avoid confusion with the DealID on chain.

nonsense (approved):

db/funds.go
dirkmc: I think it would be better to call it DealUUID 👍 
We're using DealID in other places in the DB, so I'm going to merge this code for consistency then open a follow-up PR to change the name everywhere

dirkmc: https://github.com/filecoin-project/boost/issues/87

PR #70, opened 2021-11-22
instantiate fullnode RPC API with admin perm for fullnode

nonsense:


pkg/devnet/devnet.go
```diff
@@ -186,3 +186,22 @@ func GetMinerEndpoint(ctx context.Context) (string, error) {
 
 	return ai, nil
 }
+
+func GetFullnodeEndpoint(ctx context.Context) (string, error) {
```
nonsense, line 190: This name, same as `GetMinerEndpoint` are really bad - really do not reflect what the function is doing - we should rename them.

dirkmc (approved):

PR #76, opened 2021-11-24
improve readme

nonsense:


README.md
```diff
@@ -38,7 +38,9 @@ rm -rf ~/.lotus ~/.lotusminer
 2. Create Boost repository
 
 ```
-export FULLNODE_API_INFO=/ip4/127.0.0.1/tcp/1234/http
+export `lotus auth api-info --perm=admin`
```
nonsense, line 41: We actually need a token here, as we are generating a new wallet on the fullnode, hence we need a `write` permission, not just `read`.

PR #79, opened 2021-11-24
Refactor deal transfer cancellation and remove raciness from deal update subscription

aarshkshah1992:
* Ensure all state that will be accessed by the resolver is in place before a deal is sent to the event loop for acceptance. This removes raciness because we can be sure that the state required the resolver for subscription, deal cancellation etc is in place before the deal crosses the async boundary. We need to do this for new deals and deal restarts.
* This PR also  introduces deal transfer cancellation. If the caller tries to cancel a deal that has already finished transferring, we will return an error.


storagemarket/provider.go
```diff
@@ -31,6 +34,8 @@ var log = logging.Logger("boost-provider")
 
 var ErrDealNotFound = fmt.Errorf("deal not found")
 
+var ErrDealHandlerFound = errors.New("deal handler not found")
```
dirkmc, line 37:
```suggestion
var ErrDealHandlerNotFound = errors.New("deal handler not found")
```

storagemarket/provider.go
```diff
@@ -264,22 +297,37 @@ func (p *Provider) SubscribeNewDeals() (event.Subscription, error) {
...
 
-// CancelDeal cancels a deal and any associated data transfer
-func (p *Provider) CancelDeal(ctx context.Context, dealUuid uuid.UUID) error {
-	dh, err := p.dealHandlers.get(dealUuid)
-	if err != nil {
-		if xerrors.Is(err, ErrDealHandlerFound) {
-			return nil
-		}
-		return err
+// CancelDeal cancels the data-tarnsfer associated with a deal.
```
dirkmc, line 308:
```suggestion
// CancelDealTransfer cancels the data-transfer associated with a deal.
```

storagemarket/provider.go
```diff
@@ -264,22 +297,37 @@ func (p *Provider) SubscribeNewDeals() (event.Subscription, error) {
...
+func (p *Provider) CancelDealTransfer(ctx context.Context, dealUuid uuid.UUID) error {
+	dh := p.getDealHandler(dealUuid)
+	if dh == nil {
+		return ErrDealHandlerFound
```
dirkmc, line 312: I'd suggest that we don't return an error when trying to cancel a data transfer for a deal that is not found; just return without doing anything. At this point the user probably doesn't care if it's no longer there, they just want to cancel it.

storagemarket/deal_handler.go
```diff
@@ -2,31 +2,45 @@ package storagemarket
...
-func (d *dealHandler) cancel(ctx context.Context) {
-	d.stop()
+func (d *dealHandler) cancelTransfer(ctx context.Context) error {
+	d.transferCancel()
 	select {
 	case <-ctx.Done():
-		return
-	case <-d.stopped:
-		return
+		return ctx.Err()
```
dirkmc, line 38: I'd suggest that if the context is already cancelled, it effectively means the transfer has already been cancelled so we can just return nil here

storagemarket/deal_execution.go
```diff
@@ -80,21 +63,32 @@ func (p *Provider) execDeal(ctx context.Context, pub event.Emitter, deal *types.
 
 	// Transfer Data
 	if deal.Checkpoint < dealcheckpoints.Transferred {
-		if err := p.transferAndVerify(ctx, pub, deal); err != nil {
+		if err := p.transferAndVerify(dh.transferCtx, pub, deal); err != nil {
+			dh.transferWriteOnce.Do(func() {
+				dh.transferredCancelled <- transferCancelResp{}
+				close(dh.transferredCancelled)
+			})
```
dirkmc, line 70: This code operates on variables that are internal to the deal handler struct, so I suggest putting it in a method on the dealHandler struct. Something like `dh.transferComplete()`

storagemarket/deal_handler.go
```diff
@@ -2,31 +2,45 @@ package storagemarket
...
+	case tc, ok := <-d.transferredCancelled:
+		if !ok {
+			return errors.New("cannot cancel transfer anymore")
+		}
+		return tc.err
```
dirkmc, line 43: IIUC there are two ways that an error can be returned here: because the channel has already been closed, or because transferCancelled returns a struct with a non-nil error field, which is also because the transfer can no longer be cancelled. It's quite hard to follow how we end up in these states. Can we simplify?

storagemarket/deal_handler.go
```diff
@@ -2,31 +2,45 @@ package storagemarket
...
 	"github.com/filecoin-project/boost/storagemarket/types"
 	"github.com/libp2p/go-eventbus"
 	"github.com/libp2p/go-libp2p-core/event"
-
-	"github.com/google/uuid"
 )
 
+type transferCancelResp struct {
+	err error // err=nil means that the transfer cancellation was successful.
+}
```
dirkmc, line 18: I don't think we need this struct: can just use a channel of errors

storagemarket/provider.go
```diff
@@ -142,6 +148,14 @@ func (p *Provider) ExecuteDeal(dp *types.ClientDealParams) (pi *api.ProviderDeal
...
 	}
 
+	// cleanup all state allocated here if deal is NOT accepted.
+	defer func() {
+		if pi != nil || err != nil {
+			_ = os.Remove(ds.InboundFilePath)
+			p.cleanupDealHandler(ds.DealUuid)
+		}
+	}()
+
```
dirkmc, line 158: Why not put this block after the InboundFilePath and deal handler have been created?
It's a little tricky to follow where the return variables get set. I'd suggest trying to restructure the code so that you don't have to depend on reading the value of return variables in a defer statement.

storagemarket/provider_loop.go
```diff
@@ -44,7 +46,7 @@ func (p *Provider) loop() {
 			go func() {
```
dirkmc, line 46: If there's an error in any of the restart synchronization logic, we need to call `p.cleanupDealHandler(restartReq.deal.DealUuid)`
We're also calling cleanup if the context times out trying to put the restartReq into `restartDealsChan`. I wonder if there's some way to keep all the cleanup logic in one place 🤔 

Maybe it would be simplest to remove `restartDealsChan` for now, and only add it once we have restart synchronization logic. Otherwise we're adding code complexity for something we may not actually need.

dirkmc (commented):
I need to study the code very carefully to figure out if the deal handler is going to get cleaned up in every scenario. I think we need to find a mechanism that makes it clearer 🤔

PR #81, opened 2021-11-25
use temporary homedir for filecoin devnet

nonsense:
This should make our integration tests a bit more robust - so far we've always had to clean up the default locations for the lotus-miner and lotus fullnode repository directories. With a temp dir, we don't have to do that, and just leave it to the OS to cleanup on reboot.

---

The `devnet` cmd is still using the user's HOME dir.

### DIFF
--- a/itests/dummydeal_test.go
+++ b/itests/dummydeal_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"math/rand"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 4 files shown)


itests/dummydeal_test.go
```diff
@@ -65,8 +66,13 @@ func TestDummydeal(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	tempHome, err := ioutil.TempDir(os.TempDir(), "boost-devnet-")
```
dirkmc, line 69: nit: you can also use `t.TmpDir()`

PR #84, opened 2021-11-26
Speed up test by reducing message confidence and preseal sectors

dirkmc:


itests/dummydeal_test.go
```diff
@@ -50,6 +51,10 @@ import (
 
 var log = logging.Logger("boosttest")
 
+func init() {
+	build.MessageConfidence = 1
+}
```
nonsense, line 56: Does that have an effect... I was under the impression we need to adjust this on the actual devnet.

PR #98, opened 2021-12-01
Basic mpool introspection

dirkmc:
Just mocks out the mpool messages for now, although the code is in place (but commented out) for fetching local pending messages from lotus.

Screen Shot 2021-12-01 at 5 43 06 PM

### DIFF
--- a/gql/resolver.go
+++ b/gql/resolver.go
@@ -4,18 +4,15 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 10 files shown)


nonsense:
Provided the addresses that are used are used solely for boost, I think this is fine. Given that this is a requirement, I think this screen will be useful.

nonsense (approved):

PR #103, opened 2021-12-03
initial storage manager ; local filter for mpool ; various fixes

nonsense:
This PR includes:
- initial Storage Manager implementation
- local filter for mpool
- updated Makefile with builds for both production and debug (2KiB/8MiB sectors)
- fixed dummydeal cmdline tool with correct `StartEpoch` and `EndEpoch`
- `maxDealCollateralMultiplier` hardcoded to 2, instead of 0, which is certainly not correct
- renamed `New` state to `Accepted` 

TODO:

- [x] add tests for MaxStagingDealsBytes != 0 and make sure we reject deals if we hit the limit
- [x] make sure we untag Staging storage, once a deal is passed to the sealer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,6 @@ ifneq ($(strip $(LDFLAGS)),)
 endif
 
 GOFLAGS+=-ldflags="$(ldflags)"
-GOFLAGS+=--tags=debug
 
 
 ## FFI
@@ -61,6 +60,8 @@ build/.update-modules:
 ## MAIN BINARIES
 
 CLEAN+=build/.update-modules
+
+debug: GOFLAGS+=-tags=debug
 
 deps: $(BUILD_DEPS)
 .PHONY: deps
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ build/.update-modules:
 CLEAN+=build/.update-modules
 
 debug: GOFLAGS+=-tags=debug
+debug: build
 
 deps: $(BUILD_DEPS)
 .PHONY: deps
--- a/db/storage.go
+++ b/db/storage.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
 	"github.com/google/uuid"
 )
@@ -14,7 +13,7 @@ import (
 type StorageLog struct {
 	DealUUID  uuid.UUID
 	CreatedAt time.Time
-	PieceSize abi.PaddedPieceSize
+	PieceSize uint64
 	Text      string
 }
 
@@ -26,7 +25,7 @@ func NewStorageDB(db *sql.DB) *StorageDB {
 	return &StorageDB{db: db}
 }
 
-func (s *StorageDB) Tag(ctx context.Context, dealUuid uuid.UUID, pieceSize abi.PaddedPieceSize) error {
+func (s *StorageDB) Tag(ctx context.Context, dealUuid uuid.UUID, pieceSize uint64) error {
 	qry := "INSERT INTO StorageTagged (DealUUID, CreatedAt, PieceSize) "
 	qry += "VALUES (?, ?, ?)"
 	values := []interface{}{dealUuid, time.Now(), fmt.Sprintf("%d", pieceSize)}
@@ -34,7 +33,7 @@ func (s *StorageDB) Tag(ctx context.Context, dealUuid uuid.UUID, pieceSize abi.P
 	return err
 }
 
-func (s *StorageDB) Untag(ctx context.Context, dealUuid uuid.UUID) (abi.PaddedPieceSize, error) {
+func (s *StorageDB) Untag(ctx context.Context, dealUuid uuid.UUID) (uint64, error) {
 	qry := "SELECT PieceSize FROM StorageTagged WHERE DealUUID = ?"
 	row := s.db.QueryRowContext(ctx, qry, dealUuid)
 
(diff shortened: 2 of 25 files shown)


storagemarket/deal_execution.go
```diff
@@ -319,6 +319,15 @@ func (p *Provider) addPiece(ctx context.Context, pub event.Emitter, deal *types.
 
 	p.addDealLog(deal.DealUuid, "Deal handed off to sealer successfully")
 
+	// Now that the deal has been published, we no longer need to have funds
+	// tagged as being for this deal (the publish message moves collateral
+	// from the storage market actor escrow balance to the locked balance)
```
dirkmc, line 324: 🤔

storagemarket/deal_execution.go
```diff
@@ -352,11 +361,17 @@ func (p *Provider) failDeal(pub event.Emitter, deal *types.ProviderDealState, er
...
 func (p *Provider) cleanupDeal(ctx context.Context, deal *types.ProviderDealState) {
 	_ = os.Remove(deal.InboundFilePath)
-	// ...
-	//cleanup resources here
+
+	// clean up tagged funds
 	err := p.fundManager.UntagFunds(ctx, deal.DealUuid)
 	if err != nil {
-		log.Errorw("untagging funds", "id", deal.DealUuid, "err", err)
+		log.Errorw("untagging funds", "uuid", deal.DealUuid, "err", err)
```
dirkmc, line 368: I think we've used "id" everywhere for dealUUID - fine changing it to uuid, but if so let's do it everywhere (and in a separate PR)

storagemarket/provider_loop.go
```diff
@@ -72,11 +72,22 @@ func (p *Provider) loop() {
 				continue
 			}
 
+			// Tag the storage required for the deal in the staging area
+			err = p.storageManager.Tag(p.ctx, deal.DealUuid, deal.ClientDealProposal.Proposal.PieceSize)
+			if err != nil {
+				go writeDealResp(false, &api.ProviderDealRejectionInfo{
```
dirkmc, line 78: Do we need to untag funds here as well?

itests/dummydeal_test.go
```diff
@@ -105,29 +107,61 @@ func TestDummydeal(t *testing.T) {
 	f.start()
 
 	// Create a CAR file
-	randomFilepath, err := testutil.CreateRandomFile(5, 2000000)
+	tempdir := os.TempDir()
```
dirkmc, line 110: If you use `t.TempDir()` it will automatically get cleaned up:
```
// TempDir returns a temporary directory for the test to use.
// The directory is automatically removed by Cleanup when the test and
// all its subtests complete.
// Each subsequent call to t.TempDir returns a unique directory;
// if the directory creation fails, TempDir terminates the test by calling Fatal.
```

db/storage.go
```diff
@@ -0,0 +1,148 @@
...
+
+type StorageDB struct {
+	db *sql.DB
+}
+
+func NewStorageDB(db *sql.DB) *StorageDB {
+	return &StorageDB{db: db}
+}
+
+func (s *StorageDB) Tag(ctx context.Context, dealUuid uuid.UUID, pieceSize abi.PaddedPieceSize) error {
```
dirkmc, line 29: Would it make sense to use uint64 instead of abi.PaddedPieceSize at the DB layer?

dirkmc (commented):
Looks great so far - just a couple of comments and suggestions.

db/storage.go
```diff
@@ -0,0 +1,148 @@
...
+	ps := big.NewIntUnsigned(uint64(pieceSize))
+
+	qry := "INSERT INTO StorageTagged (DealUUID, CreatedAt, PieceSize) "
+	qry += "VALUES (?, ?, ?)"
+	values := []interface{}{dealUuid, time.Now(), ps.String()}
```
dirkmc, line 34:
```suggestion
	values := []interface{}{dealUuid, time.Now(), fmt.Sprintf("%d", pieceSize)}
```

db/storage.go
```diff
@@ -0,0 +1,148 @@
...
+	for _, l := range logs {
+		if l.CreatedAt.IsZero() {
+			l.CreatedAt = now
+		}
+
+		ps := big.NewIntUnsigned(uint64(l.PieceSize))
+
+		qry := "INSERT INTO StorageLogs (DealUUID, CreatedAt, PieceSize, LogText) "
+		qry += "VALUES (?, ?, ?, ?)"
+		values := []interface{}{l.DealUUID, l.CreatedAt, ps.String(), l.Text}
```
dirkmc, line 71:
```suggestion
		values := []interface{}{l.DealUUID, l.CreatedAt, fmt.Sprintf("%d", l.pieceSize), l.Text}
```

db/storage.go
```diff
@@ -0,0 +1,148 @@
...
+	values := []interface{}{dealUuid, time.Now(), ps.String()}
+	_, err := s.db.ExecContext(ctx, qry, values...)
+	return err
+}
+
+func (s *StorageDB) Untag(ctx context.Context, dealUuid uuid.UUID) (abi.PaddedPieceSize, error) {
+	qry := "SELECT PieceSize FROM StorageTagged WHERE DealUUID = ?"
+	row := s.db.QueryRowContext(ctx, qry, dealUuid)
+
+	ps := &bigIntFieldDef{f: new(big.Int)}
```
dirkmc, line 43: I wonder if we can find a simpler way to convert from string to uint64 🤔

Makefile
```diff
@@ -62,6 +61,9 @@ build/.update-modules:
 
 CLEAN+=build/.update-modules
 
+debug: GOFLAGS+=-tags=debug
```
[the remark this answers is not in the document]
nonsense: No, it's fine as is. It gets translated to:

```
go build  -ldflags="-X=github.com/filecoin-project/boost/build.CurrentCommit=+git.d7b3bd0" -tags=debug -o boost ./cmd/boost
```

And `make build` to:

```
go build  -ldflags="-X=github.com/filecoin-project/boost/build.CurrentCommit=+git.d7b3bd0" -o boost ./cmd/boost
```

itests/dummydeal_test.go
nonsense: OK, removing the `t.Cleanup` function then!

db/storage.go
nonsense: Yes, actually it is simpler.

db/storage.go
nonsense: We can use `strconv.FormatUint64(val, 10)` or `big.Int`, I don't think it makes much difference.

storagemarket/provider_loop.go
nonsense: Yes, we do. I am not happy with the current pattern, but maybe this won't get more complex so is good enough for now - https://github.com/filecoin-project/boost/pull/103/commits/bd7616124504bbfca6b34870e2f9c0484ad9a330

dirkmc (approved):

PR #116, opened 2021-12-07
Refactor deal cleanup and cancellation and remove deal updates subscription raciness

aarshkshah1992:
- Ensure deal handler is created before deal enters the event loop.
- Deal Transfer cancellation. 
- Easy to reason about deal state cleanup irrespective of whether deal succeeds or fails.


db/funds.go
```diff
@@ -45,7 +45,7 @@ func (f *FundsDB) Untag(ctx context.Context, dealUuid uuid.UUID) (abi.TokenAmoun
 	err := row.Scan(&collat.marshalled, &pubMsg.marshalled)
 	if err != nil {
 		if err == sql.ErrNoRows {
-			return abi.NewTokenAmount(0), nil
+			return abi.NewTokenAmount(0), ErrDealNotFound
```
dirkmc, line 48: I don't think we need to return an error if the deal has already been untagged. It should make the calling code simpler if we just return nil.

storagemarket/deal_execution.go
```diff
@@ -28,49 +29,44 @@ import (
...
 
-func (p *Provider) doDeal(deal *types.ProviderDealState) {
+func (p *Provider) doDeal(deal *types.ProviderDealState, dh *dealHandler) {
 	// Set up pubsub for deal updates
-	bus := eventbus.NewBus()
-	pub, err := bus.Emitter(&types.ProviderDealState{}, eventbus.Stateful)
+	pub, err := dh.bus.Emitter(&types.ProviderDealState{}, eventbus.Stateful)
 	if err != nil {
 		err := fmt.Errorf("failed to create event emitter: %w", err)
+		p.cleanupDeal(deal)
```
dirkmc, line 37: It seems like p.cleanupDeal(deal) gets called on every code path - would it make sense to just have a single `defer p.cleanupDeal(deal)` at the start of the function?

storagemarket/deal_execution.go
```diff
@@ -350,29 +341,37 @@ func (p *Provider) failDeal(pub event.Emitter, deal *types.ProviderDealState, er
...
+	// submit req to event loop to untag tagged funds and storage space
+	select {
+	case p.finishedDealsChan <- finishedDealsReq{deal: deal, done: done}:
+	case <-p.ctx.Done():
 	}
 
-	p.dealHandlers.del(deal.DealUuid)
+	// wait for event loop to finish cleanup and return before we return from here
+	// so caller is guaranteed that all resources associated with the deal have been cleanedup before
+	// taking further action.
```
dirkmc, line 370: Nice 👍

storagemarket/deal_handler.go
```diff
@@ -37,3 +36,27 @@ func (d *dealHandler) subscribeUpdates() (event.Subscription, error) {
...
+
+	dh.transferCancel()
+
+	select {
+	case err := <-dh.transferDone:
+		dh.transferFinished = true
+		dh.transferErr = err
+		return err
+	case <-dh.providerCtx.Done():
+		return dh.providerCtx.Err()
```
dirkmc, line 56: If the provider context has been shutdown, the deal will get cancelled anyway, effectively cancelling the transfer. So in this case I think we should return nil, as the cancel action has succeeded.

storagemarket/deal_handler.go
```diff
@@ -37,3 +36,27 @@ func (d *dealHandler) subscribeUpdates() (event.Subscription, error) {
...
+	}
+}
+
+func (dh *dealHandler) close() {
```
dirkmc, line 60: nit: we don't really need this close method, you can just rename `transferCancel` to `close`

storagemarket/provider.go
```diff
@@ -144,42 +149,53 @@ func (p *Provider) ExecuteDeal(dp *types.ClientDealParams) (pi *api.ProviderDeal
...
+	cleanup := func() {
+		if tmpFile != nil {
+			_ = os.Remove(string(tmpFile.OsPath()))
+		}
+		if dh != nil {
+			dh.close()
+			p.delDealHandler(dp.DealUUID)
+		}
+	}
+
```
dirkmc, line 171: This is much easier for me to follow 👍

storagemarket/provider_loop.go
```diff
@@ -97,16 +84,25 @@ func (p *Provider) loop() {
...
-			go writeDealResp(true, nil, nil)
+			dealReq.rsp <- acceptDealResp{true, nil, nil}
 			log.Infow("deal execution started", "id", deal.DealUuid)
 
-		case failedDeal := <-p.failedDealsChan:
-			log.Errorw("deal failed", "id", failedDeal.st.DealUuid, "err", failedDeal.err)
-			// Release storage space , funds, shared resources etc etc.
+		case finishedDeal := <-p.finishedDealsChan:
+			deal := finishedDeal.deal
+			log.Errorw("deal finished", "id", deal.DealUuid)
```
dirkmc, line 95: Should this be logged at Info level?

storagemarket/provider_loop.go
```diff
@@ -34,61 +38,44 @@ func (p *Provider) loop() {
...
+			// setup cleanup function
+			cleanup := func() {
+				errf := p.fundManager.UntagFunds(p.ctx, deal.DealUuid)
+				if errf != nil && !xerrors.Is(errf, db.ErrDealNotFound) {
+					log.Errorw("untagging funds", "id", deal.DealUuid, "err", errf)
+				}
+
+				errs := p.storageManager.Untag(p.ctx, deal.DealUuid)
+				if errs != nil && !xerrors.Is(errf, db.ErrDealNotFound) {
+					log.Errorw("untagging storage", "id", deal.DealUuid, "err", errs)
```
dirkmc, line 50: This structure is much clearer for me to follow 👍

storagemarket/provider_loop.go
```diff
@@ -97,16 +84,25 @@ func (p *Provider) loop() {
...
+			log.Errorw("deal finished", "id", deal.DealUuid)
+			errf := p.fundManager.UntagFunds(p.ctx, deal.DealUuid)
+			if errf != nil {
+				log.Errorw("untagging funds", "id", deal.DealUuid, "err", errf)
+			}
+
+			errs := p.storageManager.Untag(p.ctx, deal.DealUuid)
+			if errs != nil {
+				log.Errorw("untagging storage", "id", deal.DealUuid, "err", errs)
+			}
```
dirkmc, line 104: Can this code be extracted to a common function?

db/db.go
```diff
@@ -3,11 +3,14 @@ package db
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"io/ioutil"
 	"path"
 	"runtime"
 )
 
+var ErrDealNotFound = errors.New("deal not found")
```
dirkmc, line 12: If this error is being used for storage and funds as well as deals, let's call it ErrNotFound

db/funds_test.go
```diff
@@ -27,7 +27,7 @@ func TestFundsDB(t *testing.T) {
 
 	dealUUID := uuid.New()
 	amt, err := db.Untag(ctx, dealUUID)
-	req.NoError(err)
+	req.Contains(err.Error(), "not found")
```
dirkmc, line 30: nit: use xerrors.Is(...)

storagemarket/deal_execution.go
```diff
@@ -81,10 +77,17 @@ func (p *Provider) execDeal(ctx context.Context, pub event.Emitter, deal *types.
...
+			dh.tdOnce.Do(func() {
+				close(dh.transferDone)
+			})
 			return fmt.Errorf("failed data transfer: %w", err)
 		}
 	}
+	dh.tdOnce.Do(func() {
+		dh.transferDone <- errors.New("transfer already complete")
+		close(dh.transferDone)
+	})
```
dirkmc, line 90: The code for handling the complete state of a transfer is spread around functions in this file. Would it be possible to add a function to the dealHandler class and call it from here, so as not to expose dealHandler internals?
eg `dh.completed(err)`

dirkmc (commented):
This is a nice improvement - the logic is tighter and the code is easier to follow 👍 

I left a few more suggestions for simplification.

storagemarket/deal_execution.go
aarshkshah1992: @dirkmc It's a bit tricky -> for eg: we need to ensure we've cleaned up a deal before calling `failDeal`. 
Also, we going-ahead we will probably return from this function after the deal has expired/slashed but we wanna clean up the deal state as soon as it's been handed over for sealing.

I tried doing this and the. code looks hard to reason about. Let's keep it this way for now -> it's really easy to see what's happening.

db/funds.go
aarshkshah1992: @dirkmc Trying to untag a deal that does NOT exist means something has gone wrong somewhere. If the caller does NOT care about it, it is the calling code's responsibility to handle it. However, this layer should NOT ignore that error and I think most callers should log it so we can debug what went wrong.

storagemarket/deal_handler.go
aarshkshah1992: @dirkmc This method will be called whenever a deal is cleaned up just so we can ensure all resources associated with a deal handler are reclaimed. I think it's good to keep this method and encapsulate all cleanup inside it rather then exposing the context cancellation to all the cleanup callsites.

storagemarket/provider_loop.go
aarshkshah1992: @dirkmc I am thinking of refactoring the storage & fund manager to a "resource manager". Will discuss this offline and do this change in a follow up PR as part of it.

storagemarket/provider_loop.go
aarshkshah1992: done.

storagemarket/deal_handler.go
aarshkshah1992: done.

storagemarket/deal_execution.go
aarshkshah1992: done.

db/db.go
```diff
@@ -3,11 +3,14 @@ package db
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"io/ioutil"
 	"path"
 	"runtime"
 )
 
+var ErrNotFound = errors.New("deal not found")
```
dirkmc, line 12:
```suggestion
var ErrNotFound = errors.New("not found")
```

storagemarket/deal_execution.go
dirkmc: Ok

storagemarket/deal_execution.go
```diff
@@ -81,10 +77,13 @@ func (p *Provider) execDeal(ctx context.Context, pub event.Emitter, deal *types.
 
 	// Transfer Data
 	if deal.Checkpoint < dealcheckpoints.Transferred {
-		if err := p.transferAndVerify(ctx, pub, deal); err != nil {
+		if err := p.transferAndVerify(dh.transferCtx, pub, deal); err != nil {
+			dh.transferCancelled(nil)
```
dirkmc, line 81: The transfer is not actually cancelled here right? Maybe this method should be called `transferCompleted`

storagemarket/deal_handler.go
dirkmc: Ok makes sense

storagemarket/provider_loop.go
dirkmc: Ok sounds good

storagemarket/deal_handler.go
```diff
@@ -37,3 +37,35 @@ func (d *dealHandler) subscribeUpdates() (event.Subscription, error) {
...
+		return dh.transferErr
+	}
+
+	dh.transferCancel()
+
+	select {
+	case err := <-dh.transferDone:
+		dh.transferFinished = true
+		dh.transferErr = err
+		return err
```
dirkmc, line 55: I don't think this will work if cancel is called twice before the transfer completes.
It's a little edge-casey but worth thinking about.

db/funds_test.go
aarshkshah1992: done.

db/db.go
aarshkshah1992: done.

storagemarket/deal_handler.go
aarshkshah1992: @dirkmc The locking on line 42-43 should take care of that.

storagemarket/deal_execution.go
aarshkshah1992: This method call is to inform the deal handler that a transfer isn't going to go through. What we are saying here is that the transfer not going through is successful so that the deal handler can return a nil error in response to a transfer cancellation call. Open to suggestions here.

storagemarket/deal_handler.go
dirkmc: Ah yep makes sense 👍

storagemarket/deal_execution.go
dirkmc: Ok

dirkmc (approved):

PR #136, opened 2021-12-09
Use correct payload size for a CARv1 file

aarshkshah1992:
For a CARv1 file, using the CARv2 reader over it returns an empty header. We shouldn't use this header to determine the payload size of a CARv1 file and should instead check the size of the file.

### DIFF
--- a/storagemarket/deal_execution.go
+++ b/storagemarket/deal_execution.go
@@ -281,9 +281,21 @@ func (p *Provider) addPiece(ctx context.Context, pub event.Emitter, deal *types.
 		return fmt.Errorf("failed to open CARv2 file: %w", err)
 	}
 
+	var size uint64
+	switch v2r.Version {
+	case 1:
+		st, err := os.Stat(deal.InboundFilePath)
+		if err != nil {
+			return fmt.Errorf("failed to stat CARv1 file: %w", err)
+		}
+		size = uint64(st.Size())
+	case 2:
+		size = v2r.Header.DataSize
+	}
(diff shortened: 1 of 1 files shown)


mvdan (approved):
This kind of version switching is what we imagined some advanced users might do, when @masih and myself designed the API :)

The separate stat calls are perhaps slightly unfortunate, as go-car/v2 doesn't currently expose the size of the entire file. We do have access to it via https://pkg.go.dev/golang.org/x/exp/mmap#ReaderAt.Len, but it's not exposed via go-car/v2. Could be a really easy API addition if you think it's worthwhile in terms of it being relatively common and avoiding extra syscalls.

mvdan:
Or even, one could imagine a `DataSize` method alongside `DataReader`, which would do the equivalent of what you want here - removing the need for the version switch entirely.

nonsense (approved):

aarshkshah1992:
@mvdan Yeah, having that would be great. The Datasize field should account for CARv1s. Let me know if you guys raise a PR for it.

mvdan:
Will do - feel free to leave a TODO for it here as a reminder, as I don't want to block you.

masih:
> We do have access to it via https://pkg.go.dev/golang.org/x/exp/mmap#ReaderAt.Len

@mvdan I think what we have in CARv2 is `io.ReaderAt` which does not have `Len` function. Can you think of an easy way to get length from just an `io.ReaderAt`?

PR #158, opened 2021-12-10
Add a libp2p interface for making storage deals

dirkmc:
TODO:
- [x] Provider libp2p interface
- [x] Client libp2p interface
- [x] Integration test


storagemarket/lp2pimpl/net.go
```diff
@@ -0,0 +1,153 @@
...
+// RetryParameters changes the default parameters around connection reopening
+func RetryParameters(minDuration time.Duration, maxDuration time.Duration, attempts float64, backoffFactor float64) Option {
+	return func(n *dealNet) {
+		n.retryStream.SetOptions(shared.RetryParameters(minDuration, maxDuration, attempts, backoffFactor))
+	}
+}
+
+type dealNet struct {
+	host        host.Host
+	retryStream *shared.RetryStream
```
aarshkshah1992, line 39: This creates a dependency on `go-fil-markets`. Are we okay with that ?

storagemarket/lp2pimpl/net.go
```diff
@@ -0,0 +1,153 @@
...
+func (n *dealNet) NewDealStream(ctx context.Context, id peer.ID) (*DealStream, error) {
+	s, err := n.retryStream.OpenStream(ctx, id, []protocol.ID{DealProtocolID})
+	if err != nil {
+		return nil, err
+	}
+	buffered := bufio.NewReaderSize(s, 16)
+	return &DealStream{p: id, rw: s, buffered: buffered, host: n.host}, nil
+}
+
+func (n *dealNet) ID() peer.ID {
```
aarshkshah1992, line 53: Remove all the four methods below unless we need them.

storagemarket/lp2pimpl/net.go
```diff
@@ -0,0 +1,153 @@
...
+		log.Warnw("reading storage deal proposal from stream", "id", proposal.DealUUID, "err", err)
+		return
+	}
+
+	log.Infow("new libp2p deal proposal", "id", proposal.DealUUID, "client-peer", s.Conn().RemotePeer())
+
+	// Start executing the deal.
+	// Note: This method just waits for the deal to be accepted, it doesn't
+	// wait for deal execution to complete.
+	res, err := n.prov.ExecuteDeal(&proposal)
```
aarshkshah1992, line 133: Check for a nil value of `res` here.

storagemarket/lp2pimpl/deal_stream.go
```diff
@@ -0,0 +1,53 @@
...
+
+	cborutil "github.com/filecoin-project/go-cbor-util"
+)
+
+// DealStream reads and writes deal proposals to a libp2p stream
+type DealStream struct {
+	p        peer.ID
+	host     host.Host
+	rw       mux.MuxedStream
+	buffered *bufio.Reader
```
aarshkshah1992, line 20: You can get away without having this var as the stream is also an io.Reader. If you wanna use a buffered reader over the stream, increase the size to 4K.

storagemarket/lp2pimpl/deal_stream.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/mux"
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	cborutil "github.com/filecoin-project/go-cbor-util"
+)
+
+// DealStream reads and writes deal proposals to a libp2p stream
+type DealStream struct {
```
aarshkshah1992, line 16: This struct can be breaken down into two structs, one for the client and one for the provider.

storagemarket/lp2pimpl/deal_stream.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (d *DealStream) WriteDealResponse(dr types.DealResponse) error {
+	return cborutil.WriteCborRPC(d.rw, &dr)
+}
+
+func (d *DealStream) Close() error {
+	return d.rw.Close()
+}
+
+func (d *DealStream) RemotePeer() peer.ID {
```
aarshkshah1992, line 51: Do we need this method ?

storagemarket/lp2pimpl/net.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+)
+
+var log = logging.Logger("boost-net")
+
+const DealProtocolID = "/fil/storage/mk/1.2.0"
+
+// TagPriority is the priority of deal streams in the connection manager
+const TagPriority = 100
```
aarshkshah1992, line 25: Remove this var

storagemarket/lp2pimpl/net.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	for _, option := range options {
+		option(&c.dealNet)
+	}
+	return c
+}
+
+// DealProvider listens for incoming deal proposals over libp2p
+type DealProvider struct {
+	dealNet
```
aarshkshah1992, line 89: DealProvider dosen't need the dealNet.

storagemarket/lp2pimpl/deal_stream.go
```diff
@@ -0,0 +1,53 @@
...
+		return types.DealResponse{}, err
+	}
+	return dr, nil
+}
+
+func (d *DealStream) WriteDealResponse(dr types.DealResponse) error {
+	return cborutil.WriteCborRPC(d.rw, &dr)
+}
+
+func (d *DealStream) Close() error {
```
aarshkshah1992, line 47: Make sure the client closes the stream after the deal making is done.

storagemarket/lp2pimpl/net.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func (n *DealProvider) Stop() {
+	n.host.RemoveStreamHandler(DealProtocolID)
+}
+
+// Called when the client opens a libp2p stream with a new deal proposal
+func (n *DealProvider) handleNewDealStream(s network.Stream) {
+	defer s.Close()
+
```
aarshkshah1992, line 118: Set a read deadline on the stream here so we don't indefinitely hang if the client dosen't send anything.

aarshkshah1992: @dirkmc Make sure the client also sets a read deadline on the stream before reading back the deal response.

storagemarket/lp2pimpl/deal_stream.go
aarshkshah1992: The client opens a deal stream -> writes a proposal -> reads a response. 

The provider simply handles incoming deal streams, reads the proposal and writes the response.

aarshkshah1992 (approved):

PR #161, opened 2021-12-12
storage resolver: fix pending underflow

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/159

---

Screenshot 2021-12-12 at 20 47 14

Screenshot 2021-12-12 at 20 46 59

---

This PR is fixing the `pending` overflow, and introduces the following semantics in `boost` with respect to deals storage and boost staging area:

1. `Staging` - when a deal data is successfully transferred, but still stored on `boost` side, it is marked as `Staging`.
2. `Transferred` - bytes transferred for deals that are not yet fully transferred from client to boost
3. `Pending` - bytes to be transferred to `boost`.
4. `Sealing` (not visible in UI yet) - deals that are passed from `boost` to a `sealer` node.

We have to decide how/if we want to visualise storage for deals that are already passed to a sealer node for inclusion in sectors - probably yes, but this is certainly not part of the `markets/boost staging area` storage.

### DIFF
--- a/cmd/boost/main.go
+++ b/cmd/boost/main.go
@@ -21,6 +21,7 @@ func main() {
 	_ = logging.SetLogLevel("provider", "INFO")
 	_ = logging.SetLogLevel("gql", "INFO")
 	_ = logging.SetLogLevel("boost-provider", "INFO")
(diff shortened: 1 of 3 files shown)


dirkmc (approved):
Nice 👍

PR #164, opened 2021-12-15
Deal resumption and funds untagging after publish

### LINKED ISSUE #112: Add support to retry failed deal states that are not terminal

### LINKED ISSUE #126: Deal state retry / resumption

### LINKED ISSUE #162: Untag funds by sending message event loop after publish is complete
- After publish is complete, send a message to the event loop to untag both collateral and publish funds for the deal.
- However, keep the untagging logic around in the cleanup code as well for failed deals.
- Untagging funds for a given deal id should be an idempotent operation.

aarshkshah1992:
Closes #126 .
Closes #112 .
Closes #162 .

### DIFF
--- a/storagemarket/deal_execution.go
+++ b/storagemarket/deal_execution.go
@@ -62,7 +62,7 @@ func (p *Provider) doDeal(deal *types.ProviderDealState, dh *dealHandler) {
 			p.failDeal(pub, deal, err)
 		} else {
 			// TODO For now, we will get recoverable errors only when the process is gracefully shutdown and
-			// the provider context gets cancelled which in turn cancels the transfers.
+			// the provider context gets cancelled.
 			// However, down the road, other stages of deal making will start returning
 			// recoverable errors as well and we will have to build the DB/UX/resumption support for it.
 
@@ -95,7 +95,6 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter, 
 	if deal.Checkpoint < dealcheckpoints.Transferred {
 		if err := p.transferAndVerify(dh.transferCtx, pub, deal); err != nil {
 			dh.transferCancelled(nil)
-
 			// if the transfer failed because of context cancellation and the context was not
 			// cancelled because of the user explicitly cancelling the transfer, this is a recoverable error.
 			if xerrors.Is(err, context.Canceled) && !dh.TransferCancelledByUser() {
@@ -104,7 +103,6 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter, 
 					err:         fmt.Errorf("execDeal failed data transfer: %w", err),
 				}
 			}
-
 			return &dealMakingError{
 				err: fmt.Errorf("execDeal failed data transfer: %w", err),
 			}
@@ -120,6 +118,20 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter, 
 				err: fmt.Errorf("failed to publish deal: %w", err),
 			}
 		}
+	}
+	// Now that the deal has been published, we no longer need to have funds
+	// tagged as being for this deal (the publish message moves collateral
+	// from the storage market actor escrow balance to the locked balance)
+	presp := make(chan struct{}, 1)
+	select {
+	case p.dealPublishedChan <- publishDealReq{deal: deal, done: presp}:
+	case <-ctx.Done():
+		return &dealMakingError{recoverable: true, err: ctx.Err()}
+	}
+	select {
+	case <-presp:
+	case <-ctx.Done():
+		return &dealMakingError{recoverable: true, err: ctx.Err()}
 	}
 
 	// AddPiece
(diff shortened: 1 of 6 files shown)


storagemarket/deal_execution.go
```diff
@@ -66,20 +82,30 @@ func (p *Provider) doDeal(deal *types.ProviderDealState, dh *dealHandler) {
 	// will fail, we can look into it when we implement deal completion.
 }
 
-func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter, deal *types.ProviderDealState, dh *dealHandler) error {
+func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter, deal *types.ProviderDealState, dh *dealHandler) *dealMakingError {
 	// publish "new deal" event
+	// TODO Do we need to fire this event for resumed deals ?
```
aarshkshah1992, line 87: @dirkmc Will the UI be able to handle this event getting fired again on a deal resumption on restart ?

storagemarket/deal_execution.go
```diff
@@ -50,9 +55,20 @@ func (p *Provider) doDeal(deal *types.ProviderDealState, dh *dealHandler) {
...
+	if derr := p.execDealUptoAddPiece(dh.providerCtx, pub, deal, dh); derr != nil {
+		// If the error is NOT recoverable, fail the deal and cleanup state.
+		if !derr.recoverable {
+			p.cleanupDeal(deal)
+			p.failDeal(pub, deal, err)
+		} else {
+			// TODO For now, we will get recoverable errors only when the process is gracefully shutdown and
+			// the provider context gets cancelled.
+			// However, down the road, other stages of deal making will start returning
+			// recoverable errors as well and we will have to build the DB/UX/resumption support for it.
```
dirkmc, line 67: 👌

storagemarket/deal_execution.go
dirkmc: Yes it should do 👍

storagemarket/deal_execution.go
```diff
@@ -66,20 +82,30 @@ func (p *Provider) doDeal(deal *types.ProviderDealState, dh *dealHandler) {
...
 	if deal.Checkpoint < dealcheckpoints.Transferred {
 		if err := p.transferAndVerify(dh.transferCtx, pub, deal); err != nil {
 			dh.transferCancelled(nil)
-			return fmt.Errorf("execDeal failed data transfer: %w", err)
+			// if the transfer failed because of context cancellation and the context was not
+			// cancelled because of the user explicitly cancelling the transfer, this is a recoverable error.
+			if xerrors.Is(err, context.Canceled) && !dh.TransferCancelledByUser() {
+				return &dealMakingError{
+					recoverable: true,
+					err:         fmt.Errorf("execDeal failed data transfer: %w", err),
```
dirkmc, line 103: This message is going to show up in the UI so I'd suggest we give as much detail as possible to help them figure out why the transfer failed:
```suggestion
					err:         fmt.Errorf("data transfer to client %s (peer ID %s) failed after %d bytes with error: %w", err),
```
(It's probably not a bad idea to include all this info in the logs as well)

storagemarket/deal_execution.go
```diff
@@ -88,14 +114,32 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter,
...
 		}
 	}
+	// Now that the deal has been published, we no longer need to have funds
+	// tagged as being for this deal (the publish message moves collateral
+	// from the storage market actor escrow balance to the locked balance)
+	presp := make(chan struct{}, 1)
+	select {
+	case p.publishedDealChan <- publishDealReq{deal: deal, done: presp}:
+	case <-ctx.Done():
+		return &dealMakingError{recoverable: true, err: ctx.Err()}
```
dirkmc, line 129: I don't think the context cancelled error is going to be very clear to the user in the UI, maybe for context cancelled we can have a message like "the deal was paused in the Publishing state because Boost was shut down"

storagemarket/deal_execution.go
```diff
@@ -88,14 +114,32 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter,
...
-			return fmt.Errorf("failed to publish deal: %w", err)
+			return &dealMakingError{
+				err: fmt.Errorf("failed to publish deal: %w", err),
+			}
 		}
 	}
+	// Now that the deal has been published, we no longer need to have funds
+	// tagged as being for this deal (the publish message moves collateral
+	// from the storage market actor escrow balance to the locked balance)
+	presp := make(chan struct{}, 1)
```
dirkmc, line 125: nit: suggest moving this block of code to a function - as a reader of the code I need to check carefully to make sure it doesn't modify any variables that will affect subsequent code

storagemarket/deal_handler.go
```diff
@@ -38,7 +41,12 @@ func (d *dealHandler) subscribeUpdates() (event.Subscription, error) {
 	return sub, nil
 }
 
+func (dh *dealHandler) TransferCancelledByUser() bool {
```
dirkmc, line 44: Should we also have a method that sets transferCancelledByUser to false? eg when the user clicks the "Restart" button in the UI.

aarshkshah1992: I think once a transfer is cancelled, best to fail the deal permanently for now and add resumption support here once there is a use case to resume user cancelled transfers.

dirkmc: Ok makes sense 👍

storagemarket/deal_execution.go
aarshkshah1992: done.

storagemarket/deal_execution.go
aarshkshah1992: done.

storagemarket/deal_execution.go
aarshkshah1992: done.

storagemarket/deal_execution.go
aarshkshah1992: removed this TODO

PR #168, opened 2021-12-16
Implement http tansfer resumption

aarshkshah1992:
Resume http transfers post disconnect.

### DIFF
--- a/transport/httptransport/http_transport.go
+++ b/transport/httptransport/http_transport.go
@@ -9,6 +9,11 @@ import (
 	"net/url"
 	"os"
 	"sync"
+	"time"
+
+	"golang.org/x/xerrors"
+
+	"github.com/jpillora/backoff"
 
 	"github.com/google/uuid"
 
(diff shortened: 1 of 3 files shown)


transport/httptransport/http_transport.go
```diff
@@ -121,41 +158,79 @@ func (t *transfer) emitEvent(ctx context.Context, evt types.TransportEvent, id u
...
-		// TODO: resumption
-		return fmt.Errorf("failed to execute http transfer: %w", err)
+		// backoff-retry transfer if max number of attempts haven't been exhausted
+		nAttempts := b.Attempt() + 1
+		if nAttempts >= t.h.maxReconnectAttempts {
+			return fmt.Errorf("could not finish transfer even after %d attempts, lastErr: %w", maxReconnectAttempts, err)
+		}
+		duration := b.Duration()
+		bt := time.NewTimer(duration)
+		fmt.Println(duration)
```
dirkmc, line 211: Suggest adding log lines for each attempt (and remove the fmt.Println())

transport/httptransport/http_transport.go
```diff
@@ -61,6 +96,7 @@ func (h *httpTransport) Execute(ctx context.Context, transportInfo []byte, dealI
 	tctx, cancel := context.WithCancel(ctx)
 	t := &transfer{
 		cancel:         cancel,
+		h:              h,
```
dirkmc, line 99: nit: can we just store the variables on a backoff.Backoff object, and pass that object here instead of passing the httpTransport pointer?

aarshkshah1992: done.

transport/httptransport/http_transport.go
aarshkshah1992: We will do it as part of the logging work.

transport/httptransport/http_transport.go
```diff
@@ -121,41 +166,72 @@ func (t *transfer) emitEvent(ctx context.Context, evt types.TransportEvent, id u
...
+		}
+		defer of.Close()
+
+		// start the http transfer
+		remaining := t.dealInfo.DealSize - t.nBytesReceived
+		if err = t.doHttp(ctx, req, of, remaining); err == nil {
+			// if there's no error, transfer was successful
+			break
+		}
+		_ = of.Close()
```
dirkmc, line 200: Do we need of.Close() here? It seems like it's also being called in the defer statement above on line 192
Also that defer will only get called when the function exits, I'm not sure if that's what we want?
It may be easier to do with a defer inside a function called from within the for loop.

transport/httptransport/http_transport_test.go
```diff
@@ -166,11 +170,69 @@ func TestTransferCancellation(t *testing.T) {
...
+var ConnContextKey = &contextKey{"http-conn"}
+
+func SaveConnInContext(ctx context.Context, c net.Conn) context.Context {
+	return context.WithValue(ctx, ConnContextKey, c)
+}
+func GetConn(r *http.Request) net.Conn {
+	return r.Context().Value(ConnContextKey).(net.Conn)
+}
+
+func TestTransferResumption(t *testing.T) {
```
dirkmc, line 191: This test looks great 💯

transport/httptransport/http_transport.go
aarshkshah1992: Hey, we need to close before doing the resumption just to be safe.

aarshkshah1992: I'm not sure stating on an already open file will work out well.

PR #191, opened 2022-01-13
compare boost.api.MinerAPIVersion0 with remote api version; use lotus…

nonsense:
… v1.14.0-rc1 binaries in tests


.circleci/config.yml
```diff
@@ -56,16 +56,16 @@ commands:
...
           paths:
             - /tmp/lotus-artefacts
       - setup_remote_docker:
           version: 19.03.13
       - run:
           name: copy latest stable lotus binaries
           command: |
-            docker pull nonsens3/lotus:v1.13.1-rc2
-            id=$(docker create nonsens3/lotus:v1.13.1-rc2)
+            docker pull nonsens3/lotus:v1.14.0-rc1
```
nonsense, line 67: We need to add the debug version of Lotus binaries as part of the build system of Lotus at some point. At the moment we are building only the production versions (i.e. `make lotus-miner`), whereas for our tests we need `make debug` in a Dockerfile target.

For now I am building these manually on localhost and pushing to my repo for the `boost` tests.

dirkmc (approved):

PR #193, opened 2022-01-17
Boost SM test harness and some integration tests

aarshkshah1992:
-> Wrapped some of the code related to publishing a deal, waiting for deal publish confirmation and sealing a deal inside an interface to be able to mock it in the tests.

-> Test harness to write integration & unit tests for the Boost SM.

-> Tests for making single/multiple-concurrent deals with strong state assertions.


storagemarket/helper.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/filecoin-project/go-fil-markets/storagemarket"
+	"github.com/filecoin-project/go-state-types/exitcode"
+	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/build"
+	market2 "github.com/filecoin-project/specs-actors/v2/actors/builtin/market"
+	"github.com/ipfs/go-cid"
+	"golang.org/x/xerrors"
+)
+
+type MinerHelper struct {
```
dirkmc, line 21: nit: can we use a more descriptive name than MinerHelper? IIUC the only method that gets mocked out is WaitForPublishDeals, so maybe it should be combined with the DealPublish interface, or it could be called something like ChainDealManager

storagemarket/provider.go
```diff
@@ -182,10 +179,12 @@ func (p *Provider) ExecuteDeal(dp *types.DealParams, clientPeer peer.ID) (pi *ap
 		Transfer:           dp.Transfer,
 	}
 	// validate the deal proposal
-	if err := p.validateDealProposal(ds); err != nil {
-		return &api.ProviderDealRejectionInfo{
-			Reason: fmt.Sprintf("failed validation: %s", err),
-		}, nil
+	if !testMode {
```
dirkmc, line 182: Can we add a signature to the deal proposal so we can avoid this testMode hack? (There's an example of how to do it in the dummy deal test)

storagemarket/provider_test.go
```diff
@@ -0,0 +1,552 @@
...
+	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSimpleDealHappy(t *testing.T) {
+	ctx := context.Background()
+	testMode = true
+	defer func() {
+		testMode = false
+	}()
```
dirkmc, line 61: I left a comment suggesting how we can avoid using a testMode variable.
If we can't avoid testMode, I'd suggest adding it as a field on Provider, so it gets set up automatically in the harness instead of needing to add this code to each test.

storagemarket/provider_test.go
```diff
@@ -0,0 +1,552 @@
...
+
+	// setup the provider test harness
+	// TODO Do not hardcode these values
+	minPublish := abi.NewTokenAmount(100)
+	harness := NewHarness(t, ctx, 10000000000, minPublish)
```
dirkmc, line 66: I'd suggest having default values for these, that can be overridden with harness options

storagemarket/provider_test.go
```diff
@@ -0,0 +1,552 @@
...
+	defer harness.Stop()
+	// start the provider test harness
+	harness.Start(t, ctx)
+
+	// build the deal proposal
+	dp, carV2FilePath := harness.mkDealProposal(t, harness.NormalServer.URL, 1)
+
+	// setup mock publish & add-piece expectations
+	so := harness.MinerStub.ForDeal(dp).SetupAllBlocking().Output()
+	harness.setupWalletBalances(t, big.NewInt(300), big.NewInt(500), 1000)
```
dirkmc, line 76: I suggest having harness defaults for the wallet balances that can be overridden with options

storagemarket/provider_test.go
```diff
@@ -0,0 +1,552 @@
...
+
+	if publishCid == nil {
+		require.Empty(t, dbState.PublishCID)
+	} else {
+		require.EqualValues(t, publishCid, dbState.PublishCID)
+	}
+}
+
+type ProviderHarness struct {
+	GokMockCtrl    *gomock.Controller
```
dirkmc, line 393:
```suggestion
	GoMockCtrl    *gomock.Controller
```

storagemarket/provider.go
```diff
@@ -94,19 +88,21 @@ type Provider struct {
...
+	pieceAdder                  types.PieceAdder
 	maxDealCollateralMultiplier uint64
 
 	fullnodeApi v1api.FullNode
 
-	dhsMu sync.RWMutex
-	dhs   map[uuid.UUID]*dealHandler
+	dhsMu  sync.RWMutex
+	dhs    map[uuid.UUID]*dealHandler
+	helper types.MinerHelper
```
dirkmc, line 101: nit: helper isn't protected by dhsMu so let's move it out from under the ["mutex hat"](https://dmitri.shuralyov.com/idiomatic-go#mutex-hat)

dirkmc (commented):
I left a few minor suggestions.

In general the code looks great: clean, well structured and well presented, nice work! 👍

storagemarket/provider_test.go
aarshkshah1992: It's hard to avoid that variable for now because navigating the call hierarchy of the validator will need some grunt work (see what we do in `go-fil-markets` right now, we basically mock out the entire call hierarchy). For now, I am moving this variable as a field on the Provider as you have suggested.

aarshkshah1992: As a part of writing the tests, I have enlisted a testcase for all the validation logic as well.

storagemarket/provider_test.go
aarshkshah1992: done.

storagemarket/provider_test.go
aarshkshah1992: I want these values to be explicit, so people think about it when writing tests. Keeping this hidden will then make you spend time debugging tests wondering why things are failing when you have too many deals.

storagemarket/provider.go
aarshkshah1992: Hey, please see https://github.com/filecoin-project/boost/pull/193#discussion_r786684209 but FWIW, I've move this var as a field on the Provider for now.

storagemarket/deal_handler_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"errors"
+	"testing"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCancellationSimple(t *testing.T) {
```
nonsense, line 13: @aarshkshah1992 could we add some godoc around these tests to explain what they are testing.

There is also godoc missing in the `dealHandler` so it is not clear who/why calls `dh.cancel` and `dh.transferCancelled`.

storagemarket/deal_handler_test.go
```diff
@@ -0,0 +1,64 @@
...
+		transferDone:   make(chan error, 1),
+	}
+
+	require.False(t, dh.TransferCancelledByUser())
+	dh.transferCancelled(nil)
+	require.NoError(t, dh.cancel())
+	require.True(t, dh.TransferCancelledByUser())
+}
+
+func TestCancellationByCancel(t *testing.T) {
```
nonsense, line 28: Same comment here - can we add godoc on what this method is testing?

storagemarket/deal_publisher.go
```diff
@@ -117,7 +117,7 @@ func NewDealPublisher(publishMsgCfg PublishMsgConfig) func(lc fx.Lifecycle, full
 	}
 }
 
-func newDealPublisher(
+func NewDealPublisherSimple(
```
nonsense, line 120: Why are we exporting this? I think it should stay unexported.

storagemarket/provider.go
```diff
@@ -182,10 +180,12 @@ func (p *Provider) ExecuteDeal(dp *types.DealParams, clientPeer peer.ID) (pi *ap
 		Transfer:           dp.Transfer,
 	}
 	// validate the deal proposal
-	if err := p.validateDealProposal(ds); err != nil {
-		return &api.ProviderDealRejectionInfo{
-			Reason: fmt.Sprintf("failed validation: %s", err),
-		}, nil
+	if !p.testMode {
```
nonsense, line 183: If we added interfaces for PieceAdder and ChainDealManager, maybe add an interface for the validator as well?

storagemarket/provider_test.go
```diff
@@ -0,0 +1,563 @@
...
+	// start the provider test harness
+	harness.Start(t, ctx)
+
+	// setup wallet balances
+	harness.setupWalletBalances(t, big.NewInt(300), big.NewInt(500), 1000)
+
+	var errGrp errgroup.Group
+	var testDeals []*testDealInfo
+	for i := 0; i < nDeals; i++ {
+		i := i
```
nonsense, line 116:
```suggestion
```

testutil/httptestfileservers.go
```diff
@@ -0,0 +1,21 @@
+package testutil
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func HttpTestFileServer(t *testing.T, dir string) (*httptest.Server, error) {
```
nonsense, line 9: Maybe change the signature of these 2 methods to not return an error and not have to check for it in tests?

storagemarket/provider_test.go
```diff
@@ -0,0 +1,563 @@
...
+	h.NormalServer.Start()
+	require.NoError(t, h.Provider.Start(ctx))
+}
+
+func (h *ProviderHarness) Stop() {
+	h.GoMockCtrl.Finish()
+	h.NormalServer.Close()
+}
+
+func (h *ProviderHarness) mkDealProposal(t *testing.T, serverURL string, seed int) (dp *types.DealParams, carV2FilePath string) {
```
nonsense, line 499: Nit: I suggest `newDealProposal` as a better name

```suggestion
func (h *ProviderHarness) newDealProposal(t *testing.T, serverURL string, seed int) (dp *types.DealParams, carV2FilePath string) {
```

storagemarket/provider_test.go
```diff
@@ -0,0 +1,563 @@
...
+
+// configures the max bytes allocated to the staging area
+func MaxStagingDealBytes(maxBytes uint64) HarnessOpt {
+	return func(h *ProviderHarness) {
+		h.MaxStagingDealBytes = maxBytes
+	}
+}
+
+func NewHarness(t *testing.T, ctx context.Context, opts ...HarnessOpt) *ProviderHarness {
+	// tempDir
```
nonsense, line 411:
```suggestion
```

storagemarket/provider_test.go
```diff
@@ -0,0 +1,563 @@
...
+
+	Provider *Provider
+
+	// http test servers
+	NormalServer *httptest.Server
+}
+
+type HarnessOpt func(h *ProviderHarness)
+
+// configures the min publish balance for each deal
```
nonsense, line 396:
```suggestion
// MinPublishFee configures the min publish balance for each deal
```

storagemarket/provider.go
dirkmc: Ok makes sense

storagemarket/provider_test.go
dirkmc: Ok makes sense

dirkmc (approved):

storagemarket/deal_handler_test.go
aarshkshah1992: done.

storagemarket/deal_handler_test.go
aarshkshah1992: done.

storagemarket/provider.go
aarshkshah1992: Hey yes, have added it as a TODO on the issue we have for writing tests. This will be in a follow-up PR.

storagemarket/provider_test.go
```diff
@@ -0,0 +1,563 @@
...
+	"github.com/filecoin-project/specs-actors/actors/builtin/market"
+
+	"github.com/filecoin-project/boost/db"
+	"github.com/filecoin-project/boost/fundmanager"
+	"github.com/filecoin-project/boost/storagemanager"
+	"github.com/filecoin-project/boost/storagemarket/types"
+	"github.com/filecoin-project/go-address"
+	lotusmocks "github.com/filecoin-project/lotus/api/mocks"
+
+	"testing"
```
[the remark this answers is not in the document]
aarshkshah1992: done.

storagemarket/provider_test.go
aarshkshah1992: done.

testutil/httptestfileservers.go
aarshkshah1992: Hey, doing that removes the ability to call this from go-routines. I realised while writing the tests that it's always good for harness helpers to return an error so we can call it from go-routines and then assert on the error by having the error group go-routines return the error.

storagemarket/provider_test.go
aarshkshah1992: done.

storagemarket/provider_test.go
aarshkshah1992: done.

storagemarket/provider_test.go
aarshkshah1992: done.

storagemarket/provider_test.go
```diff
@@ -0,0 +1,563 @@
...
+type HarnessOpt func(h *ProviderHarness)
+
+// configures the min publish balance for each deal
+func MinPublishFee(minPublishBal abi.TokenAmount) HarnessOpt {
+	return func(h *ProviderHarness) {
+		h.MinPublishFees = minPublishBal
+	}
+}
+
+// configures the max bytes allocated to the staging area
```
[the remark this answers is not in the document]
aarshkshah1992: done.

storagemarket/deal_publisher.go
aarshkshah1992: Hey it's for the tests so we don't have to deal with the lifecycle stuff when writing tests for the Boost SM.

storagemarket/provider_test.go
```diff
@@ -0,0 +1,563 @@
...
+			}
+			if i%2 == 0 {
+				if err := harness.waitForCheckpoint(sub, dealcheckpoints.AddedPiece); err != nil {
+					return err
+				}
+			} else {
+				if err := harness.waitForCheckpoint(sub, dealcheckpoints.Published); err != nil {
+					return err
+				}
+			}
```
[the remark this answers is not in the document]
aarshkshah1992: done.

aarshkshah1992:
@nonsense Merging as I have addressed your review.

storagemarket/deal_publisher.go
nonsense: `NewDealPublisherSimple` is used only in the same package. If we don't need a struct or func exported, it really shouldn't be exported.

testutil/httptestfileservers.go
nonsense: WDYM? There is no error here, we always return `nil`. Why do you need the `nil` error to call this from a goroutine?

PR #201, opened 2022-01-27
initial deal acceptance filter with sealing pipeline state

nonsense:
This PR is adding the storage deal user cmd acceptance filter to Boost in a backward compatible format to how it currently works in Lotus.

Users can add configuration to `config.toml` at `Dealmaking` to configure:

```
[Dealmaking]
  ConsiderOnlineStorageDeals = true
  ConsiderOfflineStorageDeals = true
  ConsiderOnlineRetrievalDeals = true
  ConsiderOfflineRetrievalDeals = true
  ConsiderVerifiedStorageDeals = true
  ConsiderUnverifiedStorageDeals = true
  PieceCidBlocklist = []
```

as well as the `user cmd` at:

```
[Dealmaking]
  Filter = ""
```

---

`SealingPipelineState` struct is added to the JSON blob that is passed on to the Dealmaking.Filter cmd, which contains a summary of the current state of the sealing pipeline. We will probably want to add a few more params to the JSON blob, so that filters such as CID gravity can avoid having to make a lot of API calls to Lotus and immediately act on information in the input JSON blob.


gql/resolver_sealingpipeline.go
```diff
@@ -124,12 +127,16 @@ type worker struct {
...
+type SealingPipelineState struct {
 	WaitDeals    waitDeals
 	SectorStates sectorStates
 	Workers      []*worker
 }
 
+type sealingPipelineResolver struct {
+	SealingPipelineState
+}
+
```
dirkmc, line 139: nit: do we actually need the sealingPipelineResolver? Can we just return a sealingPipelineState object from `func (r *resolver) SealingPipeline`?

node/builder.go
```diff
@@ -404,12 +406,40 @@ func ConfigBoost(c interface{}) Option {
...
 		// Sealing Pipeline State API
-		Override(new(sealingpipeline.State), From(new(modules.MinerStorageService))),
+		Override(new(sealingpipeline.API), From(new(modules.MinerStorageService))),
 
 		Override(new(*storagemarket.Provider), modules.NewStorageMarketProvider(walletMiner)),
 
 		// GraphQL server
 		Override(new(*gql.Server), modules.NewGraphqlServer),
+
+		// Config (todo: get a real property system)
```
dirkmc, line 416: What do you mean by a "real property system"?

node/modules/storageminer.go
```diff
@@ -568,3 +569,90 @@ func NewGraphqlServer(lc fx.Lifecycle, prov *storagemarket.Provider, dealsDB *db
 
 	return server
 }
+
+func BasicDealFilter(cfg config.DealmakingConfig, userCmd dtypes.StorageDealFilter) func(onlineOk dtypes.ConsiderOnlineStorageDealsConfigFunc,
```
dirkmc, line 573: nit: suggest putting this in a separate file so that storageminer.go doesn't get too bloated

storagemarket/provider_loop.go
```diff
@@ -42,6 +43,41 @@ func (p *Provider) loop() {
 			deal := dealReq.deal
 			log.Infow("process accept deal request", "id", deal.DealUuid)
 
+			// get current sealing pipeline status
```
dirkmc, line 46: Refactoring suggestion:
In several places in this case block we have code that
- checks for an error
- creates ProviderDealRejectionInfo with the error
- sends an acceptDealResp on the response channel
- `continue`s

It may be easier to follow the logic if we can make it look more like the traditional go form of `if err != nil { return err }`. So for example if we put all the logic in this case block into a function something like `processDealRequest(dealReq) error` then we can do
```
	err := processDealRequest(dealReq)
	if err != nil {
		ri := &api.ProviderDealRejectionInfo{Accepted: false, Reason: err.Error()}
		dealReq.rsp <- acceptDealResp{ri: ri, err: nil}
		continue
	}
```

nonsense: Good idea, but we actually need both `err` and `reason`. Have a look at the refactor and let me know if this is what you had in mind.

node/builder.go
nonsense: Copied over from Lotus. The idea is that we don't want a new set of overrides for every single config, but rather a dynamic property system, where we can add properties and hot-reload / change them, without having to statically define every single one of them.

dirkmc: Ok makes sense - could we update the comment to clarify 🙏

nonsense: I removed the comment altogether. We will probably get rid of these dynamic configs soon.

dirkmc (commented):
This is a lot easier for me to follow 👍 

A couple more suggestions:
1. In the code that calls `processDealRequest` check if there's both an error and a reason returned. That way we can both log the error locally eg "database down" and return something human-readable to the client "could not process your request due to a server error".
2. In `processDealRequest` return explicit values instead of having named return values. That way it's easier to follow when we're returning just an error, just a reason or both an error and a reason.

storagemarket/provider_loop.go
```diff
@@ -34,12 +34,11 @@ type publishDealReq struct {
...
 }
 
-func (p *Provider) processDealRequest(deal *types.ProviderDealState) (ok bool, reason string, rerr error) {
+func (p *Provider) processDealRequest(deal *types.ProviderDealState) (bool, string, error) {
 	// get current sealing pipeline status
 	status, err := sealingpipeline.GetStatus(p.ctx, p.fullnodeApi, p.sps)
 	if err != nil {
-		reason = err.Error()
-		return
+		return false, err.Error(), nil
```
dirkmc, line 41:
```suggestion
		return false, "server error", err
```

storagemarket/provider_loop.go
```diff
@@ -52,12 +51,11 @@ func (p *Provider) processDealRequest(deal *types.ProviderDealState) (ok bool, r
 	}
 	accept, reason, err := p.df(p.ctx, params)
 	if err != nil {
-		reason = err.Error()
-		return
+		return false, err.Error(), nil
```
dirkmc, line 54:
```suggestion
		return false, "deal filter error", err
```

storagemarket/provider_loop.go
```diff
@@ -78,17 +76,15 @@ func (p *Provider) processDealRequest(deal *types.ProviderDealState) (ok bool, r
 	if err != nil {
 		cleanup()
 
-		reason = err.Error()
-		return
+		return false, err.Error(), nil
```
dirkmc, line 79:
```suggestion
		return false, "server error", err
```

storagemarket/provider_loop.go
```diff
@@ -78,17 +76,15 @@ func (p *Provider) processDealRequest(deal *types.ProviderDealState) (ok bool, r
...
 	}
 
 	// tag the storage required for the deal in the staging area
 	err = p.storageManager.Tag(p.ctx, deal.DealUuid, deal.Transfer.Size)
 	if err != nil {
 		cleanup()
 
-		reason = err.Error()
-		return
+		return false, err.Error(), nil
```
dirkmc, line 87:
```suggestion
		return false, "server error", err
```

storagemarket/provider_loop.go
```diff
@@ -99,14 +95,12 @@ func (p *Provider) processDealRequest(deal *types.ProviderDealState) (ok bool, r
 	if err != nil {
 		cleanup()
 
-		rerr = fmt.Errorf("failed to insert deal in db: %w", err)
-		return
+		return false, "", fmt.Errorf("failed to insert deal in db: %w", err)
```
dirkmc, line 98:
```suggestion
		return false, "server error", fmt.Errorf("failed to insert deal in db: %w", err)
```

dirkmc (approved):

storagemarket/provider_loop.go
gmanon256: @dirkmc leaving this as `return false, err.Error(), nil` as at the moment we don't differentiate between expected conditions and errors within the `StorageManager`.

This is similar to `lack of funds` in the `FundManager`, although while we can assume that `lack of funds` is a condition which should not happen, and we can return a generic `server error` to the client in this case, I don't think we should do that when the miner has its staging area overloaded -- in this case, this is not really a `server error`, but a miner-operating-at-capacity condition which should be communicated back to the client.

nonsense: @dirkmc leaving this as return false, err.Error(), nil as at the moment we don't differentiate between expected conditions and errors within the StorageManager.

This is similar to lack of funds in the FundManager, although while we can assume that lack of funds is a condition which should not happen, and we can return a generic server error to the client in this case, I don't think we should do that when the miner has its staging area overloaded -- in this case, this is not really a server error, but a miner-operating-at-capacity condition which should be communicated back to the client.

PR #202, opened 2022-01-28
Dont fail deal on graceful shutdown of Boost

aarshkshah1992:
All context cancellation errors should be marked as recoverable.


dirkmc (approved):
Approving as the logic looks correct.
Is it possible to handle this logic all in the same place (eg the where we call execDealUptoAddPiece()) instead of repeating it through the code?

aarshkshah1992:
@dirkmc  I am not sure we wanna do that yet and leave it to each checkpoint to determine what's a recoverable error and what's not. 

For example, if it later turns out that you can't recover from a context cancelled error in the `AddPiece` state, it will become a non-recoverable error. 

For data-transfer, in additional to context cancellation, we also need to verify that the transfer was not cancelled by the user  before deeming an error to be recoverable.

PR #205, opened 2022-01-31
Test for deal completion on Boost process resumption

aarshkshah1992:
- Test for deal completion on Boost process resumption for each checkpoint
- Miner Stub should respect the context.
- Add harness capability to shutdown the existing provider and instantiate a new provieder.


storagemarket/deal_execution.go
```diff
@@ -396,6 +396,10 @@ func (p *Provider) addPiece(ctx context.Context, pub event.Emitter, deal *types.
 
 	// Add the piece to a sector
 	packingInfo, packingErr := p.AddPieceToSector(p.ctx, *deal, paddedReader)
+	if xerrors.Is(packingErr, context.Canceled) {
+		_ = v2r.Close()
+		return packingErr
+	}
```
dirkmc, line 402: It's a little confusing for me to follow the logic on lines 399 - 411. Is there a way we can combine it into less if statements? Let's add some comments as well to explain what's going on.

itests/dummydeal_test.go
```diff
@@ -133,6 +133,7 @@ func TestDummydeal(t *testing.T) {
 	res, err := f.makeDummyDeal(dealUuid, carFilepath, rootCid, server.URL+"/"+filepath.Base(carFilepath))
 	require.NoError(t, err)
 	require.True(t, res.Accepted)
+	fmt.Println(res.Reason)
```
dirkmc, line 136:
```suggestion
```

storagemarket/provider_resumption_test.go
```diff
@@ -0,0 +1,113 @@
...
+			waitForAndAssertBeforeResumeF: func(t *testing.T, h *ProviderHarness, td *testDeal) {
+				td.waitForAndAssert(t, ctx, dealcheckpoints.PublishConfirmed)
+				h.EventuallyAssertStorageFundState(t, ctx, td.params.Transfer.Size, abi.NewTokenAmount(0), abi.NewTokenAmount(0))
+			},
+		},
+	}
+
+	for name, tc := range tcs {
+		t.Run(name, func(t *testing.T) {
+			// setup the provider test harness
```
dirkmc, line 77: Nice 👍 
This code + comments are super clear

storagemarket/provider_test.go
```diff
@@ -49,15 +51,14 @@ func TestSimpleDealHappy(t *testing.T) {
 
 	// setup the provider test harness
 	harness := NewHarness(t, ctx)
-	defer harness.Stop()
```
dirkmc, line 52: I'm curious why did you change defer harness.Stop() everywhere in this file? My understanding is that if there's a test failure, the test execution will immediately stop. But the defer will cause harness.Stop() to be called even if test execution aborts.

storagemarket/smtestutil/mocks.go
```diff
@@ -94,27 +99,31 @@ func (mb *MinerStubBuilder) SetupAllBlocking() *MinerStubBuilder {
...
-			<-ch
+			select {
+			case <-ctx.Done():
+				return cid.Undef, ctx.Err()
+			case <-ch:
+			}
+
+		}
+		if ctx.Err() != nil {
+			return cid.Undef, ctx.Err()
```
dirkmc, line 121: I don't think you need to check `if ctx.Err() != nil` because you're already selecting against `<-ctx.Done()` which does the same thing.

storagemarket/deal_execution.go
aarshkshah1992: simplified.

storagemarket/smtestutil/mocks.go
aarshkshah1992: Yeah but that select is guarded by an if.

storagemarket/deal_execution.go
dirkmc: I don't see any change - do you still need to push a commit?

storagemarket/provider_test.go
aarshkshah1992: Hmm, I don't remember why I did this. I agree, it's wrong. I've put it back now.

itests/dummydeal_test.go
aarshkshah1992: done.

storagemarket/deal_execution.go
aarshkshah1992: pushed now.

dirkmc (approved):

PR #231, opened 2022-02-14
integrate lotus-markets into boost

nonsense:
This PR is merging dependencies of Lotus markets with Boost, so that we run both protocols within one process.

We are re-using the same repository (under `~/.boost` by default).

The idea here is to provide an upgrade path for storage providers where they can replace their existing `markets` process with the new `boost` process, and in the process:
- be backward compatible with Lotus markets
- support the new storage deal protocols enabled by Boost

---

FOLLOW UP TODO:
- [ ] include itests for Lotus markets storage and retrieval deals

---

TODO:
- [x] refactor `RepoType` in https://github.com/filecoin-project/lotus
- [ ] review all Boost configs and remove unused values from config.Boost
- [x] introduce both storage deal filters (both markets and boost)
- [x] introduce retrieval deal filter (backward-compat with markets)


go.mod
```diff
@@ -4,29 +4,31 @@ go 1.16
 
 replace github.com/filecoin-project/filecoin-ffi => ./extern/filecoin-ffi
 
+//replace github.com/filecoin-project/lotus => ../lotus
+
+//replace github.com/ipfs/go-datastore => ../../ipfs/go-datastore
+
```
dirkmc, line 10: Let's remove these debugging lines

node/config/types.go
```diff
@@ -29,11 +31,16 @@ type Backup struct {
 type Boost struct {
 	Common
 
+	Storage            sectorstorage.SealerConfig // TODO: make sure this is set on Boost
```
dirkmc, line 34: Has this TODO been taken care of?

dirkmc (approved):

PR #238, opened 2022-02-16
Fixes based on integration testing

aarshkshah1992:
- Don't log client signature and transfer params other than the URL.
- Use a non-cancellable context for important db updates such as checkpoints and logging.
- Better logging for transfer progress.
- Better logging for funds tagging and untagging.
- Remove unnecessary chatter from deal resumption logs.

### DIFF
--- a/fundmanager/fundmanager.go
+++ b/fundmanager/fundmanager.go
@@ -57,14 +57,14 @@ func New(cfg Config) func(api v1api.FullNode, sqldb *sql.DB) *FundManager {
 }
 
 type TagFundsResp struct {
-	ForCollat  abi.TokenAmount
-	ForPublish abi.TokenAmount
+	Collateral     abi.TokenAmount
+	PublishMessage abi.TokenAmount
 
-	TotalTaggedCollat  abi.TokenAmount
-	TotalTaggedPublish abi.TokenAmount
+	TotalCollateral     abi.TokenAmount
+	TotalPublishMessage abi.TokenAmount
 
-	RemainingCollat  abi.TokenAmount
-	RemainingPublish abi.TokenAmount
+	AvailableCollateral     abi.TokenAmount
+	AvailablePublishMessage abi.TokenAmount
 }
 
 // TagFunds tags funds for deal collateral and for the publish storage
@@ -112,14 +112,14 @@ func (m *FundManager) TagFunds(ctx context.Context, dealUuid uuid.UUID, proposal
 	}
 
 	return &TagFundsResp{
-		ForCollat:  dealCollateral,
-		ForPublish: m.cfg.PubMsgBalMin,
+		Collateral:     dealCollateral,
+		PublishMessage: m.cfg.PubMsgBalMin,
 
-		TotalTaggedPublish: big.Add(tagged.PubMsg, m.cfg.PubMsgBalMin),
-		TotalTaggedCollat:  big.Add(tagged.Collateral, dealCollateral),
+		TotalPublishMessage: big.Add(tagged.PubMsg, m.cfg.PubMsgBalMin),
+		TotalCollateral:     big.Add(tagged.Collateral, dealCollateral),
 
-		RemainingPublish: big.Sub(availForPubMsg, m.cfg.PubMsgBalMin),
-		RemainingCollat:  big.Sub(availForDealCollat, dealCollateral),
+		AvailablePublishMessage: big.Sub(availForPubMsg, m.cfg.PubMsgBalMin),
+		AvailableCollateral:     big.Sub(availForDealCollat, dealCollateral),
 	}, nil
 }
 
@@ -141,7 +141,7 @@ func (m *FundManager) totalTagged(ctx context.Context) (*db.TotalTagged, error) 
 // UntagFunds untags funds that were associated (tagged) with a deal.
 // It's called when it's no longer necessary to prevent the funds from being
 // used for a different deal (eg because the deal failed / was published)
-func (m *FundManager) UntagFunds(ctx context.Context, dealUuid uuid.UUID) (pub, collat abi.TokenAmount, err error) {
+func (m *FundManager) UntagFunds(ctx context.Context, dealUuid uuid.UUID) (collat, pub abi.TokenAmount, err error) {
 	untaggedCollat, untaggedPublish, err := m.db.Untag(ctx, dealUuid)
 	if err != nil {
 		return abi.NewTokenAmount(0), abi.NewTokenAmount(0), fmt.Errorf("persisting untag funds for deal to DB: %w", err)
@@ -160,7 +160,7 @@ func (m *FundManager) UntagFunds(ctx context.Context, dealUuid uuid.UUID) (pub, 
 	}
 
 	log.Infow("untag", "id", dealUuid, "amount", tot)
-	return untaggedPublish, untaggedCollat, nil
+	return untaggedCollat, untaggedPublish, nil
 }
 
 func (m *FundManager) persistTagged(ctx context.Context, dealUuid uuid.UUID, dealCollateral abi.TokenAmount, pubMsgBal abi.TokenAmount) error {
--- a/fundmanager/fundmanager.go
+++ b/fundmanager/fundmanager.go
@@ -56,32 +56,43 @@ func New(cfg Config) func(api v1api.FullNode, sqldb *sql.DB) *FundManager {
 	}
 }
 
+type TagFundsResp struct {
+	ForCollat  abi.TokenAmount
+	ForPublish abi.TokenAmount
+
+	TotalTaggedCollat  abi.TokenAmount
+	TotalTaggedPublish abi.TokenAmount
+
+	RemainingCollat  abi.TokenAmount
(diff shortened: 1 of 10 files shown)


fundmanager/fundmanager.go
```diff
@@ -56,51 +56,71 @@ func New(cfg Config) func(api v1api.FullNode, sqldb *sql.DB) *FundManager {
...
 
+type TagFundsResp struct {
+	ForCollat  abi.TokenAmount
+	ForPublish abi.TokenAmount
+
+	TotalTaggedCollat  abi.TokenAmount
+	TotalTaggedPublish abi.TokenAmount
+
+	RemainingCollat  abi.TokenAmount
+	RemainingPublish abi.TokenAmount
```
dirkmc, line 67: nit: I suggest renaming these for consistency with how they're used in the code base:
ForCollat -> Collateral
ForPublish -> PublishMessage
TotalTaggedCollat -> TotalCollateral
TotalTaggedPublish -> TotalPublishMessage
RemainingCollat -> AvailableCollateral
RemainingPublish -> AvailablePublishMessage

fundmanager/fundmanager.go
```diff
@@ -121,24 +141,26 @@ func (m *FundManager) totalTagged(ctx context.Context) (*db.TotalTagged, error)
...
 	err = m.db.InsertLog(ctx, fundsLog)
 	if err != nil {
-		return fmt.Errorf("persisting untag funds log to DB: %w", err)
+		return abi.NewTokenAmount(0), abi.NewTokenAmount(0), fmt.Errorf("persisting untag funds log to DB: %w", err)
 	}
 
-	log.Infow("untag", "id", dealUuid, "amount", untaggedAmt)
-	return nil
+	log.Infow("untag", "id", dealUuid, "amount", tot)
+	return untaggedPublish, untaggedCollat, nil
```
dirkmc, line 163: Let's swap the order of these return values so it matches FundsDB.Untag

storagemarket/deal_execution.go
```diff
@@ -242,12 +252,30 @@ func (p *Provider) transferAndVerify(ctx context.Context, pub event.Emitter, dea
...
 
 	p.dealLogger.Infow(deal.DealUuid, "commP matched successfully: deal-data verified")
-	return p.updateCheckpoint(ctx, pub, deal, dealcheckpoints.Transferred)
+	return p.updateCheckpoint(pub, deal, dealcheckpoints.Transferred)
 }
 
 func (p *Provider) waitForTransferFinish(ctx context.Context, handler transport.Handler, pub event.Emitter, deal *types.ProviderDealState) error {
 	defer handler.Close()
 	defer p.transfers.complete(deal.DealUuid)
+	tlog := make(map[int]struct{})
```
dirkmc, line 261: nit: we don't really need to use a map here, or loop through all the percentages.

We can just keep track of the last percent output, and compare against it each time:
```
pct := (100 * received) / int64(deal.Transfer.Size)
outputPct := pct % 10
if outputPct != lastOutputPct {
    lastOutputPct = outputPct
    log(outputPct)
}
```

storagemarket/deal_execution.go
```diff
@@ -495,10 +528,11 @@ func (p *Provider) fireEventDealUpdate(pub event.Emitter, deal *types.ProviderDe
 	}
 }
 
-func (p *Provider) updateCheckpoint(ctx context.Context, pub event.Emitter, deal *types.ProviderDealState, ckpt dealcheckpoints.Checkpoint) error {
+func (p *Provider) updateCheckpoint(pub event.Emitter, deal *types.ProviderDealState, ckpt dealcheckpoints.Checkpoint) error {
 	prev := deal.Checkpoint
 	deal.Checkpoint = ckpt
-	if err := p.dealsDB.Update(ctx, deal); err != nil {
+	// we don't want a graceful shutdown to mess with db updates
```
dirkmc, line 534:
```suggestion
	// we don't want a graceful shutdown to mess with db updates so pass a background context
```

storagemarket/deal_execution.go
```diff
@@ -495,10 +528,11 @@ func (p *Provider) fireEventDealUpdate(pub event.Emitter, deal *types.ProviderDe
...
+	if err := p.dealsDB.Update(context.TODO(), deal); err != nil {
```
dirkmc, line 535:
```suggestion
	if err := p.dealsDB.Update(context.Background(), deal); err != nil {
```
https://www.programming-books.io/essential/go/context.todo-vs.context.background-d5224e27ff724a33a79cb4e03a5eb333.html

storagemarket/logs/log.go
```diff
@@ -75,7 +72,8 @@ func (d *DealLogger) updateLogDB(dealId uuid.UUID, msg string, level string, kvs
 		LogParams: string(jsn),
 		Subsystem: d.subsystem,
 	}
-	if err := d.logsDB.InsertLog(d.ctx, l); err != nil {
+	// we don't want context cancellations to mess up our logging
+	if err := d.logsDB.InsertLog(context.TODO(), l); err != nil {
```
dirkmc, line 76:
```suggestion
	if err := d.logsDB.InsertLog(context.Background(), l); err != nil {
```

storagemarket/provider_loop.go
```diff
@@ -59,9 +59,12 @@ func (p *Provider) processDealRequest(deal *types.ProviderDealState) (bool, stri
 	}
 
 	cleanup := func() {
-		errf := p.fundManager.UntagFunds(p.ctx, deal.DealUuid)
+		pub, collat, errf := p.fundManager.UntagFunds(p.ctx, deal.DealUuid)
 		if errf != nil && !xerrors.Is(errf, db.ErrNotFound) {
 			p.dealLogger.LogError(deal.DealUuid, "failed to untag funds during deal cleanup", err)
+		} else {
+			p.dealLogger.Infow(deal.DealUuid, "untagged funds for deal cleanup", "untagged publish", pub, "untagged collateral", collat,
+				"err", errf)
```
dirkmc, line 67: I think we may want to log something different (or maybe just log nothing at all) if errf is db.ErrNotFound

storagemarket/provider_loop.go
```diff
@@ -72,12 +75,19 @@ func (p *Provider) processDealRequest(deal *types.ProviderDealState) (bool, stri
...
 
 		return false, "server error", fmt.Errorf("failed to tag funds for deal: %w", err)
 	}
+	p.dealLogger.Infow(deal.DealUuid, "tagged funds for deal",
+		"tagged for deal publish", trsp.ForPublish,
+		"tagged for deal collateral", trsp.ForCollat,
+		"total tagged for publish", trsp.TotalTaggedPublish,
+		"total tagged for collateral", trsp.TotalTaggedCollat,
+		"total remaining for publish", trsp.RemainingPublish,
+		"total remaining for collateral", trsp.RemainingCollat)
```
dirkmc, line 90:
```suggestion
		"total available for publish", trsp.RemainingPublish,
		"total available for collateral", trsp.RemainingCollat)
```

storagemarket/provider_loop.go
```diff
@@ -128,26 +138,31 @@ func (p *Provider) loop() {
...
 		case finishedDeal := <-p.finishedDealChan:
 			deal := finishedDeal.deal
 			p.dealLogger.Infow(deal.DealUuid, "deal finished")
-			errf := p.fundManager.UntagFunds(p.ctx, deal.DealUuid)
+			pub, collat, errf := p.fundManager.UntagFunds(p.ctx, deal.DealUuid)
 			if errf != nil && !xerrors.Is(errf, db.ErrNotFound) {
 				p.dealLogger.LogError(deal.DealUuid, "failed to untag funds", errf)
+			} else {
+				p.dealLogger.Infow(deal.DealUuid, "untagged funds for deal as deal finished", "untagged publish", pub, "untagged collateral", collat,
+					"err", errf)
```
dirkmc, line 165: Same here: maybe we want to log something different / not log if errf is db.ErrNotFound

storagemarket/deal_execution.go
```diff
@@ -47,14 +49,21 @@ func (p *Provider) doDeal(deal *types.ProviderDealState, dh *dealHandler) {
 		return
 	}
 
-	p.dealLogger.Infow(deal.DealUuid, "deal execution initiated", "deal state", *deal)
+	if isRestart {
+		p.dealLogger.Infow(deal.DealUuid, "resuming deal on boost restart")
+	}
```
dirkmc, line 54: Instead of adding the `isRestart` parameter to this method, can we just add this log line in the calling method if it's a restart?

storagemarket/provider_loop.go
aarshkshah1992: Well, it's be nice to see if it was not found in the db vs there was no error and I don't wanna add one more branch here.

storagemarket/deal_execution.go
aarshkshah1992: Good idea

aarshkshah1992: This dosen't work because it logs the output every one percent. We wanna log only every X percent so we will have to use a map.

PR #239, opened 2022-02-16
ordered close

aarshkshah1992:
- Shut down deal acceptance -> shut down event loop -> shut down all deal go-routines -> shut down any pending db op if any.

### DIFF
--- a/db/funds.go
+++ b/db/funds.go
@@ -36,7 +36,7 @@ func (f *FundsDB) Tag(ctx context.Context, dealUuid uuid.UUID, collateral abi.To
 	return err
 }
 
-func (f *FundsDB) Untag(ctx context.Context, dealUuid uuid.UUID) (abi.TokenAmount, error) {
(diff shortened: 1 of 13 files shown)


dirkmc:
It's a little hard to follow where each context is used - I think it would be easier if we pass the context as a parameter to methods that use the same context everywhere, for example validateDealProposal and processDealRequest

aarshkshah1992:
Well, if user cancels the context, they want to shut down. Now, we are saying that

- We will allow you to shutdown only after we've flushed out all ongoing DB writes.
- The DB call needs a context so it can giveup on DB ops that take time (remote DB server etc)
- What if the DB op takes time and that's contending with the user's wish to shutdown ?
- Either we simply disobey the user's shutdown req and block till db op eventually completes -> context background OR
- We allow user to specify (with a default config) how much to wait before giving up on DB writes ?

PR #263, opened 2022-02-21
Parallelize dummydeal test vs other tests

dirkmc:


aarshkshah1992:
@dirkmc We had an interesting race in the tests where

-> we execute a deal -> it finishes before we subscribe for updates -> deal handler gets cleaned- up -> we subscribe for updates. 

I have pushed commits to fix that.

PR #268, opened 2022-02-21
markets v1 deal integration test

nonsense:
This PR is adding `TestMarketsV1Deal` - a happy path test for the markets v1 functionality that Boost inherits from Lotus.


pkg/devnet/devnet.go
```diff
@@ -128,6 +128,9 @@ func runLotusMiner(ctx context.Context, home string) {
 		{"sed", "-Ei", "-e", "s/#AggregateCommits\\ =\\ true/AggregateCommits=false/",
 			filepath.Join(home, ".lotusminer", "config.toml")},
 
+		{"sed", "-Ei", "-e", "s/#EnableMarkets\\ =\\ true/EnableMarkets=false/",
```
nonsense, line 131: @dirkmc @aarshkshah1992 I think it makes sense for the `devnet` in Boost to always disable markets on Lotus.

itests/markets_v1_deal_test.go
```diff
@@ -0,0 +1,102 @@
...
+	// Create a new markets v1 deal
+	dp := f.DefaultMarketsV1DealParams()
+	dp.Data.Root = res.Root
+
+	dealProposalCid, err := f.fullNode.ClientStartDeal(ctx, &dp)
+	require.NoError(t, err)
+
+	f.WaitDealSealed(ctx, dealProposalCid)
+
+	//TODO: confirm retrieval works
```
nonsense, line 80: This is for a follow-up PR.

itests/markets_v1_deal_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	file, err := os.CreateTemp(t.TempDir(), "sourcefile.dat")
+	require.NoError(t, err)
+
+	n, err := io.Copy(file, source)
+	require.NoError(t, err)
+	require.EqualValues(t, n, size)
+
+	return file.Name()
+}
```
dirkmc, line 102: We have a function to do this in testutil/car.go

.circleci/config.yml
```diff
@@ -118,7 +118,7 @@ jobs:
         default: golang
       go-test-flags:
         type: string
-        default: "-v --tags=debug -timeout 4m"
+        default: "-v --tags=debug -timeout 15m"
```
dirkmc, line 121: Do we still need this 15m timeout?

gql/server.go
```diff
@@ -62,25 +66,41 @@ func (s *Server) Serve(ctx context.Context) error {
...
+
+func (s *Server) Stop(ctx context.Context) error {
+	if err := s.srv.Shutdown(ctx); err != nil {
+		return err
+	}
+
+	s.wg.Wait()
+
+	return nil
+}
```
dirkmc, line 106: Thanks for adding this 🙏

dirkmc (approved):

.circleci/config.yml
nonsense: It certainly needs to be >4m, I will adjust with 2min slack and check how long these tests take (since we have more integration tests now, and we'll be adding retrieval soon).

nonsense: Successful run took 12min. so will leave this at 15min. for now.

PR #281, opened 2022-02-23
Integrate dagstore and piecestore

aarshkshah1992:
So Markets can serve retrievals for Boost storage deals.


storagemarket/deal_execution.go
```diff
@@ -186,6 +190,21 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter,
 		p.dealLogger.Infow(deal.DealUuid, "deal has already been handed over to the sealing subsystem")
 	}
 
+	// Index deal in DAGStore and Announce deal
+	if deal.Checkpoint < dealcheckpoints.IndexedAndAnnounced {
+		if err := p.indexAndAnnounce(ctx, pub, deal); err != nil {
+			// any error here is always a recoverable error as this step is completely idempotent
```
dirkmc, line 196: It looks like there's the possibility for a non-recoverable error also - for example adding a deal to the piece store (eg if there's a problem connecting to the datastore)

aarshkshah1992: It should never fail unless something has temporarily gone wrong.

itests/dummydeal_test.go
```diff
@@ -147,9 +147,10 @@ func TestDummydeal(t *testing.T) {
 	require.Equal(t, "cannot accept piece of size 2254421, on top of already allocated 2254421 bytes, because it would exceed max staging area size 4000000", res2.Reason)
 	log.Debugw("got response from MarketDummyDeal for failing deal", "res2", spew.Sdump(res2))
 
-	// Wait for the deal to be added to a sector
+	// Wait for the deal to be added to a sector and be cleanedup so space is made
 	err = f.waitForDealAddedToSector(dealUuid)
 	require.NoError(t, err)
+	time.Sleep(100 * time.Millisecond)
```
aarshkshah1992, line 153: @dirkmc @nonsense We need this so the previous deal gets time be to cleanedup and space to be released which is done _after_ the deal finishes and we only wait for deal to finish here.

aarshkshah1992: This is why the itest was flaky at times as it sent the next deal before this one was cleanedup and so the next deal ran into not enough space issues.

nonsense (approved):

node/builder.go
```diff
@@ -369,6 +369,7 @@ func ConfigBoost(c interface{}) Option {
 	if !ok {
 		return Error(xerrors.Errorf("invalid config from repo, got: %T", c))
 	}
+	cfg.Storage.ParallelFetchLimit = 10
```
nonsense, line 372: @aarshkshah1992 maybe add this to the itest, rather than here, if this is used only for tests.

aarshkshah1992: You mean before instantiating the itest ?

nonsense: while instantiating Boost in tests, you can set any config you need. since this config is necessary for tests, and not production code, it can live in the test `framework`.

PR #296, opened 2022-02-28
Use legacy deal publisher

dirkmc:
https://user-images.githubusercontent.com/169124/156156795-8fca3ecc-2eb7-4470-b49a-f00602e8c63a.mov


gql/resolver.go
```diff
@@ -31,23 +33,22 @@ type dealListResolver struct {
 // resolver translates from a request for a graphql field to the data for
 // that field
 type resolver struct {
+	cfg        *config.Boost
```
nonsense, line 36: Looks like we are only using `r.cfg.LotusDealmaking.MaxDealsPerPublishMsg` -- maybe pass only that value, rather than the whole `config.Boost` ?

nonsense (approved):

PR #297, opened 2022-02-28
Deal Status V2 protocol

aarshkshah1992:
Implementation of the Deal Status V2 protocol in Boost.


storagemarket/lp2pimpl/net.go
```diff
@@ -146,3 +186,44 @@ func (p *DealProvider) handleNewDealStream(s network.Stream) {
...
+		return
+	}
+	log.Debugw("received deal status request", "id", req.DealUUID, "client-peer", s.Conn().RemotePeer())
+
+	// Set a deadline on writing to the stream so it doesn't hang
+	_ = s.SetWriteDeadline(time.Now().Add(providerWriteDeadline))
+	defer s.SetWriteDeadline(time.Time{}) // nolint
+
+	pds, err := p.prov.Deal(p.ctx, req.DealUUID)
+	if err != nil && err == storagemarket.ErrDealNotFound {
```
dirkmc, line 209: I suggest changing this to
```suggestion
	if err != nil && xerrors.Is(err, storagemarket.ErrDealNotFound) {
```
in case we change the Deal function to return something like `fmt.Errorf("some err: %w", ErrDealNotFound)`

PR #303, opened 2022-03-02
Add fields to deal status response

dirkmc:
- Add a signature to the deal status request and check the signature
- Add several fields to the deal status response to match deal status v1 protocol


nonsense (approved):
lgtm

PR #309, opened 2022-03-07
libp2p server: wait for client to close connection before firing complete

dirkmc:


transport/httptransport/libp2p_server.go
```diff
@@ -225,15 +228,49 @@ func (s *Libp2pCarServer) sendCar(r *http.Request, w http.ResponseWriter, val *A
...
+	return err
+}
+
+// waitForClientClose waits for the client to close the connection
+func waitForClientClose(ctx context.Context, r *http.Request) error {
+	streamClosed := make(chan error, 1)
+	go func() {
+		// Block until Read returns an EOF, which means the connection has
+		// been closed
+		_, err := r.Body.Read(make([]byte, 1024))
```
aarshkshah1992, line 258: What if the server simply dosen't write anything ? Does this read call request the request context so we don;t leak these go routines ?

dirkmc: Discussed sync: when the http handler returns, it will close the stream, so the Read call with return with an io.EOF

Issue #233, opened 2022-02-15
Pprof support in Boost


aarshkshah1992:
https://github.com/filecoin-project/boost/pull/229 improves the introspection story for Boost. We need to add support for pprof now.

PR #310, opened 2022-03-07
add pprof to boost

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/233

### DIFF
--- a/cmd/boost/run.go
+++ b/cmd/boost/run.go
@@ -2,9 +2,6 @@ package main
 
 import (
 	"fmt"
-	"os"
-	"runtime"
-	"runtime/pprof"
 
 	"github.com/filecoin-project/boost/api"
 	"github.com/filecoin-project/boost/node"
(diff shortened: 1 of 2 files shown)


cmd/boost/run.go
```diff
@@ -22,6 +28,39 @@ var runCmd = &cli.Command{
...
+		}
+
+		if memprofile := cctx.String("pprof.mem"); memprofile != "" {
+			f, err := os.Create(memprofile)
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+			runtime.GC() // get up-to-date statistics
+			if err := pprof.WriteHeapProfile(f); err != nil {
```
dirkmc, line 50: I think this will write a heap profile immediately, which may not be that useful. Probably we want to see a heap profile at a particular point in time, which we can get through the pprof web service.

Maybe we don't need the pprof.cpu and pprof.mem flags at all and we can just rely on the pprof web service?

nonsense: I agree, maybe I went too far with the cpu and mem profiles. Let's just have the web service for now, and add more later if we need it.

dirkmc (approved):

PR #312, opened 2022-03-08
Protect libp2p connection for lifetime of data transfer

dirkmc:


transport/httptransport/http_transport.go
```diff
@@ -134,17 +134,36 @@ func (h *httpTransport) Execute(ctx context.Context, transportInfo []byte, dealI
 		dl:                   h.dl,
 	}
 
+	cleanupFns := []func(){
```
aarshkshah1992, line 137: javascript for the win

PR #313, opened 2022-03-08
Index Provider integration

aarshkshah1992:
Should be able to announce indices and serve indices for Boost deals and legacy markets deals.
Markets PR at https://github.com/filecoin-project/go-fil-markets/pull/691.

### TODO

- [x] Boost should start after markets so the index provider callback isn't overwritten.
- [ ] Test Market deals announcement on Sofia Miner.
- [ ] Test Boost deal announcement on Sofia Miner.
- [ ] Test the `AnnounceAllDeals` CLI on Sofia Miner.


cmd/boost/index.go
```diff
@@ -0,0 +1,55 @@
...
+		&cli.BoolFlag{
+			Name:        "color",
+			Usage:       "use color in display output",
+			DefaultText: "depends on output being a TTY",
+		},
+	},
+	Action: func(cctx *cli.Context) error {
+		if cctx.IsSet("color") {
+			color.NoColor = !cctx.Bool("color")
+		}
```
dirkmc, line 33: This should probably be a flag at the global level (not per command).
I'd say let's add it in a later PR if we need it.

node/builder.go
```diff
@@ -576,3 +590,16 @@ func (f boost) Config() interface{} {
...
+func (boost) APIFlags() []string {
+	return []string{"boost-api-url"}
+}
+
+func (boost) RepoFlags() []string {
+	return []string{"boost-repo"}
+}
+
+func (boost) APIInfoEnvVars() (primary string, fallbacks []string, deprecated []string) {
+	// TODO remove deprecated deprecation period
```
dirkmc, line 603: What does this comment mean?

node/config/def.go
```diff
@@ -148,6 +148,13 @@ func DefaultBoost() *Boost {
 			MaxConcurrencyStorageCalls: 100,
 			GCInterval:                 lotus_config.Duration(1 * time.Minute),
 		},
+		IndexProvider: lotus_config.IndexProviderConfig{
+			Enable:               false,
```
dirkmc, line 152: Do we want index announcement to be disabled by default?

storagemarket/provider.go
```diff
@@ -321,6 +325,9 @@ func (p *Provider) Start() ([]*dealHandler, error) {
 	}
 	log.Infow("db initialized")
 
+	// start the index provider
+	p.ip.Start()
+
```
dirkmc, line 330: Why are we starting the index provider from inside the provider?

storagemarket/types/deal_state.go
```diff
@@ -39,6 +39,8 @@ type ProviderDealState struct {
 	ChainDealID abi.DealID
 	PublishCID  *cid.Cid
 
+	IndexerAnnouncementCID *cid.Cid
```
dirkmc, line 42: Why do we need the `IndexerAnnouncementCID` on the deal state?

indexprovider/wrapper.go
```diff
@@ -0,0 +1,136 @@
...
+	return &Wrapper{
+		dealsDB:     dealsDB,
+		legacyProv:  legacyProv,
+		prov:        prov,
+		dagStore:    dagStore,
+		meshCreator: meshCreator,
+	}
+}
+
+func (w *Wrapper) IndexerAnnounceAllDeals(ctx context.Context) error {
```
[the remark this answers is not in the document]
aarshkshah1992: We can do it or we can do it as we're currently doing it. I don't have a strong preference.

node/builder.go
aarshkshah1992: I just copied this code from Lotus.

storagemarket/types/deal_state.go
aarshkshah1992: Actually, just logging it should be good enough for debugging as we will have it in the deal log. Removing this.

storagemarket/deal_execution.go
```diff
@@ -503,6 +503,14 @@ func (p *Provider) indexAndAnnounce(ctx context.Context, pub event.Emitter, deal
 	}
 	p.dealLogger.Infow(deal.DealUuid, "deal successfully registered in dagstore")
 
+	// announce to the network indexer
+	annCid, err := p.ip.AnnounceBoostDeal(ctx, deal)
+	if err != nil {
+		return fmt.Errorf("failed to annoucne deal to index provider: %w", err)
```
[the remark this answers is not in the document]
aarshkshah1992: Fixed this. Also, yes, it will be retried because any error returned by this function is always a recoverable error.

indexprovider/wrapper.go
```diff
@@ -0,0 +1,136 @@
...
+		if _, err := w.AnnounceBoostDeal(ctx, d); err != nil {
+			merr = multierror.Append(merr, err)
+			log.Errorw("failed to announce boost deal to Index provider", "dealId", d.DealUuid, "err", err)
+			continue
+		}
+		shards[d.ClientDealProposal.Proposal.PieceCID.String()] = struct{}{}
+		nSuccess++
+	}
+
+	log.Infow("finished announcing boost deals to index provider", "number of deals", nSuccess, "number of shards", shards)
```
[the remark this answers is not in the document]
aarshkshah1992: Not sure what help that will give but we can.

node/modules/storageminer.go
```diff
@@ -463,39 +465,41 @@ func NewLogsDB(logsSqlDB *LogSqlDB) *db.LogsDB {
 	return db.NewLogsDB(logsSqlDB.db)
 }
 
+func HandleBoostDeals(lc fx.Lifecycle, h host.Host, prov *storagemarket.Provider, a v1api.FullNode) {
```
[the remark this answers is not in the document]
aarshkshah1992: This is done.

storagemarket/types/deal_state.go
aarshkshah1992: This is done.

indexprovider/wrapper.go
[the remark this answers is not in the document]
dirkmc: Where possible I'd prefer to keep knowledge of the two different data sources inside the wrapper, rather than leaking it into the CLI layer

storagemarket/provider.go
dirkmc: It's a bit weird because we use the index provider externally to the storage market provider (eg to announce all deals), but we start the index provider inside the storage market provider.

I'd suggest that we call index provider start at the modules layer, maybe right after we call storage market provider start.

node/config/def.go
aarshkshah1992: So, we enabled it by default in Lotus RC but disable it by default in Lotus master. 

I'm doing what Lotus RC does by default and enabling it for now. We can disable later if we need to.

indexprovider/wrapper.go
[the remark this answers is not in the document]
aarshkshah1992: Man, moving all that markets code here (which has intimate knowledge of various market states and what they mean) will add mantainence burden here and is something I'd like to avoid. Remember, the plan is to eventually deprecate markets.

storagemarket/provider.go
aarshkshah1992: Done.

node/builder.go
aarshkshah1992: Done.

dirkmc (approved):

aarshkshah1992:
Subsumbed by https://github.com/filecoin-project/boost/pull/319

PR #319, opened 2022-03-09
announce indices for deals (rebased on main)

nonsense:
Should be able to announce indices and serve indices for Boost deals and legacy markets deals.
Markets PR at https://github.com/filecoin-project/go-fil-markets/pull/691.

### TODO

- [x] Boost should start after markets so the index provider callback isn't overwritten.
- [x] Test Market deals announcement on Sofia Miner.
- [x] Test Boost deal announcement on Sofia Miner.
- [x] Test the `AnnounceAllDeals` CLI on Sofia Miner.


cmd/boost/index.go
```diff
@@ -0,0 +1,50 @@
...
+			Name:        "color",
+			Usage:       "use color in display output",
+			DefaultText: "depends on output being a TTY",
+		},
+	},
+	Action: func(cctx *cli.Context) error {
+		ctx := lcli.ReqContext(cctx)
+
+		// announce markets deals
+		nodeApi, closer, err := lcli.GetMarketsAPI(cctx)
```
nonsense, line 33: @aarshkshah1992 I think this should be `GetBoostAPI`.

cmd/boost/index.go
```diff
@@ -0,0 +1,50 @@
...
+		if err != nil {
+			return err
+		}
+		defer closer()
+		if err := nodeApi.IndexerAnnounceAllDeals(ctx); err != nil {
+			fmt.Printf("some errors while announcing legacy markets deals: %s", err)
+		}
```
nonsense, line 40: @aarshkshah1992 this doesn't make sense to me. We will never have a `markets` and `boost` deployment at the same time. I guess we should remove this.

indexprovider/wrapper.go
```diff
@@ -0,0 +1,136 @@
...
+		dealsDB:     dealsDB,
+		legacyProv:  legacyProv,
+		prov:        prov,
+		dagStore:    dagStore,
+		meshCreator: meshCreator,
+	}
+}
+
+func (w *Wrapper) IndexerAnnounceAllDeals(ctx context.Context) error {
+	log.Info("will announce all Boost deals to Indexer")
```
nonsense, line 44: This is not very informative, maybe downgrade to `Debug` ?

cmd/boost/index.go
```diff
@@ -0,0 +1,39 @@
...
+
+var indexProvAnnounceAllCmd = &cli.Command{
+	Name:  "announce-all",
+	Usage: "Announce all active deals to indexers so they can download the indices",
+	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:        "color",
+			Usage:       "use color in display output",
+			DefaultText: "depends on output being a TTY",
+		},
```
nonsense, line 25:
```suggestion
```

indexprovider/wrapper.go
```diff
@@ -0,0 +1,142 @@
...
+		legacyProv:  legacyProv,
+		prov:        prov,
+		dagStore:    dagStore,
+		meshCreator: meshCreator,
+	}
+}
+
+func (w *Wrapper) IndexerAnnounceAllDeals(ctx context.Context) error {
+	log.Info("will announce all Markets deals to Indexer")
+	err := w.legacyProv.AnnounceAllDealsToIndexer(ctx)
```
nonsense, line 45: We don't seem to be doing anything with that error, maybe check it and have a WARN log message if it is != nil ?

cmd/boost/index.go
aarshkshah1992: This is done.

storagemarket/deal_execution.go
```diff
@@ -498,11 +502,18 @@ func (p *Provider) indexAndAnnounce(ctx context.Context, pub event.Emitter, deal
...
+	if err := stores.RegisterShardSync(ctx, p.dagst, pc, deal.InboundFilePath, true); err != nil && !xerrors.Is(err, dagstore.ErrShardExists) {
 		return fmt.Errorf("failed to register deal with dagstore: %w", err)
 	}
 	p.dealLogger.Infow(deal.DealUuid, "deal successfully registered in dagstore")
 
+	// announce to the network indexer
+	annCid, err := p.ip.AnnounceBoostDeal(ctx, deal)
+	if err != nil && !xerrors.Is(err, provider.ErrAlreadyAdvertised) {
+		return fmt.Errorf("failed to announce deal to index provider: %w", err)
+	}
```
nonsense, line 514: Shouldn't we check for err != nil (i.e. when the err is provider.ErrAlreadyAdvertised), before saying that deal is successfully announced?

node/modules/storageminer.go
```diff
@@ -463,39 +465,56 @@ func NewLogsDB(logsSqlDB *LogSqlDB) *db.LogsDB {
 	return db.NewLogsDB(logsSqlDB.db)
 }
 
+// We start Boost in this way because we want to ensure Boost is started after the legacy markets Provider.
+// We want to do this so that the index provider callback registered by Boost can overwrite the index
+// provider callback registered by legacy markets as legacy markets does NOT know about Boost storage deals.
```
dirkmc, line 470: Is this comment still correct? It looks like we're starting the index provider in `HandleIndexProvider`

dirkmc (approved):

indexprovider/wrapper.go
aarshkshah1992: It's just once per that CLI command. Harmless but nice to have when going through logs

storagemarket/deal_execution.go
aarshkshah1992: Hey Anton,  if the error is `provider.ErrAlreadyAdvertised`, we have nothing to worry about and can consider announcement to be successful given how the index provider stuff works.

node/modules/storageminer.go
aarshkshah1992: Removed.

aarshkshah1992: And added a comment to `HandleIndexProvider`.

storagemarket/deal_execution.go
nonsense: Returning both a result and an error is certainly not idiomatic behaviour and somewhat confusing - this is why I mentioned the comment. We basically end up with an error and with the result (`annCid`), which is really not what one expects unless they know the full stack.

cmd/boost/index.go
nonsense: `color` is not used, so we can remove it.

storagemarket/deal_execution.go
aarshkshah1992: I see what you mean. Have fixed this.

PR #321, opened 2022-03-09
only fire a completed event on the server once the client has closed the stream

dirkmc:


transport/httptransport/libp2p_server.go
```diff
@@ -223,6 +233,10 @@ func (s *Libp2pCarServer) sendCar(r *http.Request, w http.ResponseWriter, val *A
 		err = e
 	}}
 
+	// Get a channel that will be closed when the client closes the connection
+	stream := getConn(r).(gostream.Stream)
+	closeCh := s.streamMonitor.getCloseChan(stream.ID())
```
aarshkshah1992, line 238: This is racy .

If there's a disconnection while serving the file below, you will deem success on a stream that the client isn't gonna close anymore or has already closed because that is from the previous connection.

aarshkshah1992: I'm really leaning towards simply enquiring the deal status to know that it has finished transferring rather than doing the state management gymnastics here.

dirkmc: If there's a disconnect during the transfer, the err will be non-nil and we will return the err on line 249

aarshkshah1992 (approved):

PR #322, opened 2022-03-10
CLI command to migrate from legacy markets repo to boost repo

dirkmc:
`boost migrate --import-markets-repo=~/.markets-repo`


cmd/boost/init.go
```diff
@@ -55,156 +63,576 @@ var initCmd = &cli.Command{
...
+		log.Debug("Creating boost config")
+		var cerr error
+		err = lr.SetConfig(func(raw interface{}) {
+			rcfg, ok := raw.(*config.Boost)
+			if !ok {
+				cerr = xerrors.New("expected boost config")
+				return
+			}
+
+			asi, err := checkApiInfo(cctx.Context, cctx.String("api-sector-index"))
```
nonsense, line 91: Not sure if this context will be canceled on SIGTERM and all other signals. I think we should be using the ReqContext everywhere, as it is terminated on these signals.

cmd/boost/init.go
```diff
@@ -55,156 +63,576 @@ var initCmd = &cli.Command{
...
 		}
+		defer mktsRepo.Close() //nolint:errcheck
 
-		log.Debug("Checking full node sync status")
-
-		if err := lcli.SyncWait(ctx, &v0api.WrapperV1Full{FullNode: api}, false); err != nil {
-			return xerrors.Errorf("sync wait: %w", err)
+		// Migrate datastore keys
+		log.Info("Migrating datastore keys")
+		err = migrateMarketsDatastore(cctx.Context, ds, mktsRepo)
```
nonsense, line 195:
```suggestion
		err = migrateMarketsDatastore(ctx, ds, mktsRepo)
```

nonsense (approved):

PR #330, opened 2022-03-15
offline deal support in Boost

aarshkshah1992:
Closes https://github.com/filecoin-project/boost/issues/197

### TODO
- [x] Unit tests.
- [x] E2E Test.

### DIFF
--- a/cmd/boost/deals.go
+++ b/cmd/boost/deals.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"fmt"
+
+	bcli "github.com/filecoin-project/boost/cli"
+	"github.com/google/uuid"
+	"github.com/urfave/cli/v2"
+)
+
+var dealCmd = &cli.Command{
+	Name:  "deal",
+	Usage: "Manage Boost deals",
+	Subcommands: []*cli.Command{
+		makeOfflineDealWithData,
+	},
+}
+
+var makeOfflineDealWithData = &cli.Command{
+	Name:  "offline-deal-with-data",
+	Usage: "Make offline deal with data",
+	Flags: []cli.Flag{&cli.StringFlag{
+		Name:  "dealUuid",
+		Usage: "uuid of the offline deal",
+	},
+
+		&cli.StringFlag{
+			Name:  "filepath",
+			Usage: "path of the file containing the offline deal data",
+		},
+	},
+
+	Action: func(cctx *cli.Context) error {
+		if cctx.NArg() != 2 {
+			return fmt.Errorf("must specify deal uuid and deal data file path")
+		}
+
+		napi, closer, err := bcli.GetBoostAPI(cctx)
+		if err != nil {
+			return err
+		}
+		defer closer()
+
+		filePath := cctx.String("filepath")
+		id := cctx.String("dealUuid")
+
+		dealUuid, err := uuid.Parse(id)
+		if err != nil {
+			return fmt.Errorf("failed to parse deal uuid")
+		}
+		rej, err := napi.MakeOfflineDealWithData(dealUuid, filePath)
+		if err != nil {
+			return fmt.Errorf("failed to execute offline deal: %w", err)
+		}
+		if rej != nil && rej.Reason != "" {
+			return fmt.Errorf("offline deal %s rejected: %s\n", dealUuid, rej.Reason)
+		}
+		fmt.Println("\n offline deal is being executed")
+		return nil
+	},
+}
--- a/cmd/boost/deals.go
+++ b/cmd/boost/deals.go
@@ -17,24 +17,22 @@ var dealCmd = &cli.Command{
 }
 
 var makeOfflineDealWithData = &cli.Command{
-	Name:  "offline-deal-with-data",
+	Name:  "offline-deal",
 	Usage: "Make offline deal with data",
 	Flags: []cli.Flag{&cli.StringFlag{
-		Name:  "dealUuid",
-		Usage: "uuid of the offline deal",
+		Name:     "deal-uuid",
+		Usage:    "uuid of the offline deal",
+		Required: true,
 	},
 
 		&cli.StringFlag{
-			Name:  "filepath",
-			Usage: "path of the file containing the offline deal data",
+			Name:     "filepath",
+			Usage:    "path of the file containing the offline deal data",
+			Required: true,
 		},
 	},
 
 	Action: func(cctx *cli.Context) error {
-		if cctx.NArg() != 2 {
-			return fmt.Errorf("must specify deal uuid and deal data file path")
-		}
-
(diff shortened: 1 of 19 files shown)


cmd/boost/deals.go
```diff
@@ -0,0 +1,61 @@
...
+var dealCmd = &cli.Command{
+	Name:  "deal",
+	Usage: "Manage Boost deals",
+	Subcommands: []*cli.Command{
+		makeOfflineDealWithData,
+	},
+}
+
+var makeOfflineDealWithData = &cli.Command{
+	Name:  "offline-deal-with-data",
```
dirkmc, line 20: For simplicity maybe let's just call the command "offline-deal"

cmd/boost/deals.go
```diff
@@ -0,0 +1,61 @@
...
+	Usage: "Make offline deal with data",
+	Flags: []cli.Flag{&cli.StringFlag{
+		Name:  "dealUuid",
```
dirkmc, line 23: I think conventionally we use snake-case for param names, ie "deal-uuid"

cmd/boost/deals.go
```diff
@@ -0,0 +1,61 @@
...
+		&cli.StringFlag{
+			Name:  "filepath",
+			Usage: "path of the file containing the offline deal data",
+		},
+	},
+
+	Action: func(cctx *cli.Context) error {
+		if cctx.NArg() != 2 {
+			return fmt.Errorf("must specify deal uuid and deal data file path")
+		}
```
dirkmc, line 36: Instead of doing this I think you can set `Required: true` on the flags

storagemarket/provider.go
```diff
@@ -202,6 +202,51 @@ func (p *Provider) GetAsk() *storagemarket.StorageAsk {
 	}
 }
 
+func (p *Provider) MakeOfflineDealWithData(dealUuid uuid.UUID, filePath string) (pi *api.ProviderDealRejectionInfo, handler *dealHandler, err error) {
+	p.dealLogger.Infow(dealUuid, "execute offline deal called", "filepath", filePath)
+
+	// db should already have a deal with this uuid as the deal proposal should have been agreed before hand
+	ds, err := p.dealsDB.ByID(p.ctx, dealUuid)
+	if err != nil {
+		return nil, nil, fmt.Errorf("no pre-existing deal proposal for offline deal: %s", err)
```
dirkmc, line 211:
```suggestion
		return nil, nil, fmt.Errorf("no pre-existing deal proposal for offline deal: %w", err)
```

storagemarket/provider.go
```diff
@@ -202,6 +202,51 @@ func (p *Provider) GetAsk() *storagemarket.StorageAsk {
...
+	}
+	ds.InboundFilePath = filePath
+
+	// setup clean-up code
+	var dh *dealHandler
+	cleanup := func() {
+		if dh != nil {
+			dh.close()
+			p.delDealHandler(dealUuid)
+		}
```
dirkmc, line 224: nit: it looks like it's not possible for dh to be nil, so I don't think you need the `if dh != nil`
I'd suggest moving the cleanup function below `dh = p.mkAndInsertDealHandler(dealUuid)` so you don't have to declare the dh variable and initialize it in two separate steps

storagemarket/types/types.go
```diff
@@ -66,9 +66,10 @@ type DealStatus struct {
 
 type DealParams struct {
 	DealUUID           uuid.UUID
+	IsOffline          bool
 	ClientDealProposal market.ClientDealProposal
 	DealDataRoot       cid.Cid
-	Transfer           Transfer
+	Transfer           Transfer // Transfer params will be a no-op of this is an offline deal
```
dirkmc, line 72:
```suggestion
	Transfer           Transfer // Transfer params will be a no-op if this is an offline deal
```

aarshkshah1992: Done.

storagemarket/provider.go
aarshkshah1992: Done.

cmd/boost/deals.go
aarshkshah1992: Done.

cmd/boost/deals.go
aarshkshah1992: Done.

cmd/boost/deals.go
aarshkshah1992: Done.

cmd/boost/deals.go
```diff
@@ -0,0 +1,61 @@
...
+		dealUuid, err := uuid.Parse(id)
+		if err != nil {
+			return fmt.Errorf("failed to parse deal uuid")
+		}
+		rej, err := napi.MakeOfflineDealWithData(dealUuid, filePath)
+		if err != nil {
+			return fmt.Errorf("failed to execute offline deal: %w", err)
+		}
+		if rej != nil && rej.Reason != "" {
+			return fmt.Errorf("offline deal %s rejected: %s\n", dealUuid, rej.Reason)
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

storagemarket/provider_loop.go
```diff
@@ -115,6 +124,29 @@ func (p *Provider) processDealRequest(deal *types.ProviderDealState) (bool, stri
...
 
+func (p *Provider) processOfflineDealRequest(deal *types.ProviderDealState) (bool, string, error) {
+	// tag funds and start executing deal
+	cleanup := func() {
+		collat, pub, errf := p.fundManager.UntagFunds(p.ctx, deal.DealUuid)
+		if errf != nil && !xerrors.Is(errf, db.ErrNotFound) {
+			p.dealLogger.LogError(deal.DealUuid, "failed to untag funds during deal cleanup", errf)
+		} else if errf != nil {
+			p.dealLogger.Infow(deal.DealUuid, "untagged funds for deal cleanup", "untagged publish", pub, "untagged collateral", collat,
+				"err", errf)
```
dirkmc, line 135:
```suggestion
				)
```

storagemarket/types/types.go
```diff
@@ -66,9 +66,10 @@ type DealStatus struct {
 
 type DealParams struct {
 	DealUUID           uuid.UUID
+	IsOffline          bool
 	ClientDealProposal market.ClientDealProposal
 	DealDataRoot       cid.Cid
-	Transfer           Transfer
+	Transfer           Transfer // Transfer params will be a no-op if this is an offline deal
```
dirkmc, line 72:
```suggestion
	Transfer           Transfer // Transfer params will be the zero value if this is an offline deal
```

storagemarket/provider.go
aarshkshah1992: Will do it as part of https://github.com/filecoin-project/boost/pull/338 that will go in after this.

PR #336, opened 2022-03-16
Show storage space used by legacy deals

dirkmc:
Screen Shot 2022-03-16 at 9 15 11 AM

### DIFF
--- a/storagemanager/storagemanager.go
+++ b/storagemanager/storagemanager.go
@@ -15,7 +15,7 @@ import (
 var log = logging.Logger("storagemanager")
 
 var (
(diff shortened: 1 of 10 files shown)


storagemanager/storagemanager.go
```diff
@@ -15,7 +15,7 @@ import (
 var log = logging.Logger("storagemanager")
 
 var (
-	StagingAreaDirName = "deal-staging"
+	StagingAreaDirName = "deal-downloads"
```
nonsense, line 18: Good catch!

nonsense (approved):

PR #345, opened 2022-03-18
propose deal cmd

nonsense:


cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,299 @@
...
+			Name:     "piece-size",
+			Usage:    "size of the CAR file as a padded piece",
+			Required: true,
+		},
+		&cli.Uint64Flag{
+			Name:     "car-size",
+			Usage:    "size of the CAR file",
+			Required: true,
+		},
+		&cli.StringFlag{
```
nonsense, line 60: Technically `payload-cid` is not required, as we just pass it in the `label` field... so shall we mark it as `optional` ?

.circleci/config.yml
```diff
@@ -56,16 +56,16 @@ commands:
       - restore_cache:
           name: Restore lotus artefacts cache
           keys:
```
nonsense, line 58: Every time APIs change on Lotus, we have to update the commit that our CI points to, so that the `devnet`, `boost` and `lotus` are compatible.

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,299 @@
...
+		},
+		&cli.IntFlag{
+			Name:  "provider-collateral",
+			Usage: "",
+			Value: 0,
+		},
+		&cli.IntFlag{
+			Name:  "days",
+			Usage: "how many days should this deal be active for",
+			Value: 181,
```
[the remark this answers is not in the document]
nonsense: 180

cmd/boost/deal_cmd.go
dirkmc: I think let's mark it as required - even though it's not required technically it's become a kind of defacto standard

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,343 @@
...
+		},
+		&cli.StringFlag{
+			Name:     "http-url",
+			Usage:    "http url to CAR file",
+			Required: true,
+		},
+		&cli.StringSliceFlag{
+			Name:  "http-headers",
+			Usage: "http headers to be passed with the request",
+		},
```
dirkmc, line 50: Suggest adding a description with an example of the format for the http-headers param

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,343 @@
...
+		&cli.StringFlag{
+			Name:     "provider",
+			Usage:    "provider address",
```
dirkmc, line 53: Let's make it clear the `provider` parameters is a filecoin address, rather than a multiaddr. Maybe also a description with an example.

cli/node/node.go
```diff
@@ -0,0 +1,161 @@
...
+func Setup(ctx context.Context, cfgdir string) (*Node, error) {
+	peerkey, err := loadOrInitPeerKey(keyPath(cfgdir))
+	if err != nil {
+		return nil, err
+	}
+
+	bwc := metrics.NewBandwidthCounter()
+
+	h, err := libp2p.New(
+		libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/6766"),
```
dirkmc, line 46: I suggest we just pick any free port, as we're not expecting anyone to dial this address, it's only used for outgoing connections.
```suggestion
		libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/0"),
```

cli/node/node.go
```diff
@@ -0,0 +1,161 @@
...
+		libp2p.Identity(peerkey),
+		libp2p.BandwidthReporter(bwc),
```
dirkmc, line 48: We probably don't need the bandwidth reporter

cli/node/node.go
```diff
@@ -0,0 +1,161 @@
...
+	dht "github.com/libp2p/go-libp2p-kad-dht"
+)
+
+type Node struct {
+	Host host.Host
+
+	Datastore  datastore.Batching
+	DHT        *dht.IpfsDHT
+	Blockstore blockstore.Blockstore
+	Bitswap    *bitswap.Bitswap
```
dirkmc, line 32: I don't think we need any of these

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,344 @@
...
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:     "payload-cid",
+			Usage:    "root CID of the CAR file",
+			Required: true,
+		},
+		&cli.IntFlag{
+			Name:  "start-epoch",
+			Usage: "",
```
dirkmc, line 78: Let's explain what the start epoch is used for

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,344 @@
...
+		},
+		&cli.IntFlag{
+			Name:  "duration",
+			Usage: "epochs for this deal to be active for",
```
dirkmc, line 82:
```suggestion
			Usage: "duration of the deal in epochs",
```

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,344 @@
...
+			Value: 518400, // default is 2880 * 180 == 180 days
+		},
+		&cli.IntFlag{
+			Name:  "provider-collateral",
+			Usage: "provider collateral required from storage miner",
```
dirkmc, line 87:
```suggestion
			Usage: "deal collateral that storage miner must put in escrow",
```

I suggest we make this a string flag and use the ParseFIL method that lotus uses.

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,344 @@
...
+		},
+		&cli.Int64Flag{
+			Name:  "storage-price-per-epoch",
+			Usage: "",
```
dirkmc, line 91: Let's describe what this is, and use parseFIL

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,344 @@
...
+			Value: 1,
+		},
+		&cli.BoolFlag{
+			Name:  "verified",
+			Usage: "whether deal should be verified or not",
```
dirkmc, line 96:
```suggestion
			Usage: "whether the deal funds should come from verified client data-cap",
```

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,344 @@
...
+		maddr, err := address.NewFromString(cctx.String("provider"))
+		if err != nil {
+			return err
+		}
+		minfo, err := api.StateMinerInfo(ctx, maddr, chain_types.EmptyTSK)
+		if err != nil {
+			return err
+		}
+		if minfo.PeerId == nil {
+			return fmt.Errorf("could not find peer ID for stoarge provider %s", maddr)
```
dirkmc, line 136: I actually prefer the way you phrased it in the error message below about no multiaddresses:
```suggestion
			return fmt.Errorf("storage provider %s has no peer ID set on-chain", maddr)
```

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,344 @@
...
+		return nil, fmt.Errorf("wallet sign failed: %w", err)
+	}
+
+	return &market.ClientDealProposal{
+		Proposal:        proposal,
+		ClientSignature: *sig,
+	}, nil
+}
+
+func doRpc(ctx context.Context, s inet.Stream, req interface{}, resp interface{}) error {
```
dirkmc, line 328: If the context is cancelled will the `doRpc` function return? Just want to make sure people can ctrl-c if the command hangs while trying to contact the SP

cmd/boostx/utils_cmd.go
```diff
@@ -0,0 +1,272 @@
...
+		commp, err := w.Sum()
+		if err != nil {
+			return xerrors.Errorf("computing commP failed: %w", err)
+		}
+
+		encoder := cidenc.Encoder{Base: multibase.MustNewEncoder(multibase.Base32)}
+
+		fmt.Println("CommP CID: ", encoder.Encode(commp.PieceCID))
+		fmt.Println("Piece size: ", types.NewInt(uint64(commp.PieceSize.Unpadded().Padded())))
+		return nil
```
dirkmc, line 173: Suggest we also output the CAR file root here because they need it for the deal command

nonsense: I am doing that in the `generate-car` command.

cmd/boost/deal_cmd.go
nonsense: I think this is done now, will test it out.

cmd/boost/deal_cmd.go
```diff
@@ -0,0 +1,348 @@
...
+	errc := make(chan error)
+	go func() {
+		if err := cborutil.WriteCborRPC(s, req); err != nil {
+			errc <- fmt.Errorf("failed to send request: %w", err)
+		}
+
+		if err := cborutil.ReadCborRPC(s, resp); err != nil {
+			errc <- fmt.Errorf("failed to read response: %w", err)
```
dirkmc, line 336:
```suggestion
			errc <- fmt.Errorf("failed to read response: %w", err)
			return
```

PR #382, opened 2022-03-29
add default gateway-url with a flag

nonsense:
This PR is adding a default gateway CLI argument, so that we don't have to set an environment variable pointing to a Filecoin gateway.


jacobheun:
Should we close this out, it's 5months old?

PR #383, opened 2022-03-29
fix offline deals

dirkmc:


cmd/boost/deal_cmd.go
```diff
@@ -39,9 +39,11 @@ var dealCmd = &cli.Command{
...
 		},
 		&cli.StringFlag{
-			Name:     "http-url",
-			Usage:    "http url to CAR file",
-			Required: true,
+			Name:  "http-url",
+			Usage: "http url to CAR file",
+		},
+		&cli.BoolFlag{
+			Name: "offline",
```
aarshkshah1992, line 46: The usage doc says "make an online deal with Boost". Don't you think best to not overload this command and best top have a separate command for offline deals ?

storagemarket/deal_execution.go
```diff
@@ -148,6 +157,13 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter,
 			return &dealMakingError{err: fmt.Errorf("error when matching commP for imported data for offline deal: %w", err)}
 		}
 		p.dealLogger.Infow(deal.DealUuid, "commp matched successfully for imported data for offline deal")
+
+		// update checkpoint
+		if err := p.updateCheckpoint(pub, deal, dealcheckpoints.Transferred); err != nil {
+			return &dealMakingError{
+				err: fmt.Errorf("failed to update checkpoint: %w", err),
```
aarshkshah1992, line 164: what happens if we don't do this ?

storagemarket/provider.go
```diff
@@ -271,65 +284,80 @@ func (p *Provider) ExecuteDeal(dp *types.DealParams, clientPeer peer.ID) (pi *ap
...
 		return &api.ProviderDealRejectionInfo{Accepted: true}, nil, nil
 	}
 
-	// setup clean-up code
-	var tmpFile filestore.File
-	var dh *dealHandler
-	cleanup := func() {
-		if tmpFile != nil {
-			_ = os.Remove(string(tmpFile.OsPath()))
+	dh, _ := p.mkAndInsertDealHandler(dp.DealUUID)
```
aarshkshah1992, line 315: Why do we need the refactoring from here to the end of the function ? What problem does it solve ? Dosen't have anything to do with an offline deal.

storagemarket/provider.go
```diff
@@ -204,40 +204,51 @@ func (p *Provider) GetAsk() *storagemarket.StorageAsk {
...
 	ds.InboundFilePath = filePath
 
-	// setup clean-up code
-	var dh *dealHandler
+	// create new deal handler
+	dh, isNew := p.mkAndInsertDealHandler(dealUuid)
 	cleanup := func() {
-		if dh != nil {
+		// if the deal handler didn't already exist, clean it up on error
+		if isNew {
```
aarshkshah1992, line 233: I'm not sure I understand why we need this. Let's discuss.

storagemarket/provider.go
```diff
@@ -271,65 +284,80 @@ func (p *Provider) ExecuteDeal(dp *types.DealParams, clientPeer peer.ID) (pi *ap
 		}
 	}
 
-	// if it is an offline deal  just persist it in the DB here and return
+	// if it is an offline deal just persist it in the DB here and return
 	if dp.IsOffline {
+		p.dealLogger.Infow(dp.DealUUID, "offline deal accepted, waiting for data import")
```
aarshkshah1992, line 289: Can we put this if block code in a separate function for offline deals ?

aarshkshah1992: Actually, I think best to introduce an `ExecuteOfflineDeal` function which the new deal lib2p stream handler calls when it sees an offline deal to reduce the complexity of this function.

We can then rename this function to `ExecuteOnlineDeal`.

storagemarket/provider.go
aarshkshah1992: Please can we get rid of the refactoring below. If it adds some benefit, we can add it in a separate PR for refactoring code. But, let's not increase the scope of this offline deal PR if it's okay with you.

aarshkshah1992 (commented):
Left some comments.

storagemarket/deal_execution.go
dirkmc: Then we would run commp each time boost restarts

storagemarket/provider.go
dirkmc: When the deal is proposed we create a deal handler and publish a "new deal" event. Then when the deal is imported we need to get the same deal handler to publish progress events.

aarshkshah1992 (approved):
lgtm

PR #384, opened 2022-03-30
Fix restart after add piece

aarshkshah1992:
Upon deal resumption, we should expect the inbound file to exist  only if it hasn't already been handed over to the sealer.


aarshkshah1992:
itest failure seems unrelated . Looks like the legacy deal itest is failing with:

`2022-03-30T05:28:20.015Z	ERROR	dt-impl	impl/receiver.go:135	received error message on data transfer: unexpected EOF
`

dirkmc (approved):

PR #387, opened 2022-03-30
Cleanup completed deals on restart

aarshkshah1992:
There's a edge case where we wont cleanup a deal on resumption if it has already been marked completed even though Boost restarted after the deal was previously marked as completed but not cleaned up.

### DIFF
--- a/storagemarket/provider.go
+++ b/storagemarket/provider.go
@@ -387,8 +387,8 @@ func (p *Provider) Start() ([]*dealHandler, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to list completed deals: %w", err)
 	}
-	for _, ds := range finished {
-		p.cleanupDeal(ds)
+	for i := range finished {
+		p.cleanupDealOnRestart(finished[i])
 	}
 	log.Info("finished cleaning up completed deals")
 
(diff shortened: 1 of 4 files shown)


storagemarket/provider.go
```diff
@@ -382,33 +382,48 @@ func (p *Provider) Start() ([]*dealHandler, error) {
...
 	// restart all active deals
 	pds, err := p.dealsDB.ListActive(p.ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to list active deals: %w", err)
 	}
 
-	var dhs []*dealHandler
+	// cleanup all deals that have finished successfully
+	for i := range pds {
+		deal := pds[i]
```
nonsense, line 403:
```suggestion
	for _, deal := range pds {
```

storagemarket/provider.go
```diff
@@ -382,33 +382,48 @@ func (p *Provider) Start() ([]*dealHandler, error) {
...
+		if deal.Checkpoint >= dealcheckpoints.IndexedAndAnnounced {
+			// cleanup if cleanup didn't finish before we restarted
+			p.cleanupDealOnRestart(deal)
+		}
+	}
 
+	// resume all in-progress deals
+	var dhs []*dealHandler
 	for _, ds := range pds {
 		d := ds
```
nonsense, line 414:
```suggestion
	for _, d := range pds {
		d := d
```

nonsense (approved):

storagemarket/provider.go
aarshkshah1992: I think we need the former because deal is a pointer and we shouldn't be looping using a pointer variable, no ?

nonsense: Why not? These two are literally the same, no?

aarshkshah1992: @nonsense Look at the first answer here:
https://stackoverflow.com/questions/48826460/using-pointers-in-a-for-loop

I myself have run into this problem before and therefor stay away from using pointer vars in for loops .

PR #389, opened 2022-03-30
Default dagstore dir

aarshkshah1992:


nonsense:
This lgtm, but why do we need this directory to be configurable? We have a few other directories in a repo, and none of them are configurable (and really don't have to be):

```
dagstore  datastore  deal-staging  incoming  journal  keystore  kvlog
```

aarshkshah1992:
@nonsense  Mostly for backward compat with existing  markets functionality and dosen't hurt to make this configuable.

dirkmc (approved):

PR #395, opened 2022-03-31
Update UI when deal sealing state changes

dirkmc:


nonsense (approved):

storagemarket/provider_test.go
```diff
@@ -686,6 +682,9 @@ func NewHarness(t *testing.T, ctx context.Context, opts ...harnessOpt) *Provider
 
 	ph.MockSealingPipelineAPI.EXPECT().SectorsSummary(gomock.Any()).Return(sealingpipelineStatus, nil).AnyTimes()
```
aarshkshah1992, line 684: Please can we add verification for getting an event for atleast one of the sealing states here ?

aarshkshah1992 (commented):
some comments.

storagemarket/deal_execution.go
```diff
@@ -596,6 +596,67 @@ func (p *Provider) indexAndAnnounce(ctx context.Context, pub event.Emitter, deal
 	return p.updateCheckpoint(pub, deal, dealcheckpoints.IndexedAndAnnounced)
 }
 
+// fireSealingUpdateEvents periodically checks the sealing status of the deal
+// and fires events for each change
+func (p *Provider) fireSealingUpdateEvents(pub event.Emitter, dealUuid uuid.UUID, sectorNum abi.SectorNumber) {
+	var lastSealingState lapi.SectorState
```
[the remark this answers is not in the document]
dirkmc: They should see it immediately - when they retrieve the deal it will already show the sealing state.
Also when fireSealingUpdateEvents is called it immediately checks the status (see line 621)

storagemarket/provider_test.go
dirkmc: When the deal has been added to a piece, the deal go routine starts polling the sealer for sealing state updates. I've modified the sealing API mock to return the "Proving" state. When the code polling the sealer gets sealing state "Proving" it stops polling and returns. So given that the tests finish correctly, that verifies that the event code path is working correctly.

Let me know if you think we need other tests beyond this one.

storagemarket/deal_execution.go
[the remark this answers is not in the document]
aarshkshah1992: But shouldn't we see an event for that sealing state update to `Proving` ?

storagemarket/provider_test.go
aarshkshah1992: But shouldn't we see an event for that sealing state update to `Proving` ?

storagemarket/deal_execution.go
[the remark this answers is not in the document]
aarshkshah1992: Oh nvm, I see `checkstatus` also publishes the event.

storagemarket/provider_test.go
dirkmc: I added a test to watch for an event to be fired with sealing state `Proving`

PR #396, opened 2022-03-31
boost deal-status cmd

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/346

This PR is adding a deal status command, for example:

```
boost deal-status --provider=f0127896 --deal-uuid=0adf9397-b772-4726-924b-f140cc3ab0b4
```

---

It is also modifying the DealStatus protocol and adding a few more fields.

It is also modifying the CI configuration to include a check for CBOR generation.

---

TODO:
- [x] 1. handle offline deals in output of command (and remove TODO)
- [x] 2. decide how to handle `wallet.Wallet` - can we use the `lotus` Wallet, or we really need the updated wallet with interface that doesn't require api.Meta
- [x] 3. decide how to handle `messagesigner.MessageSigner` - same as 2.


api/wallet.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/filecoin-project/go-state-types/crypto"
+	"github.com/filecoin-project/lotus/chain/types"
+)
+
+type Wallet interface {
+	WalletNew(context.Context, types.KeyType) (address.Address, error) //perm:admin
+	WalletHas(context.Context, address.Address) (bool, error)          //perm:admin
+	WalletList(context.Context) ([]address.Address, error)             //perm:admin
+
+	WalletSign(ctx context.Context, signer address.Address, toSign []byte) (*crypto.Signature, error) //perm:admin
```
nonsense, line 16: The reason I am including `Wallet interface` here is:
1. LocalWallet in Lotus uses the following signature:
```
func (w *LocalWallet) WalletSign(ctx context.Context, addr address.Address, msg []byte, meta api.MsgMeta) (*crypto.Signature, error)
```

2. FullNode API uses a different signature:
```
	// WalletSign signs the given bytes using the given address.
	WalletSign(context.Context, address.Address, []byte) (*crypto.Signature, error) //perm:sign
```

Ideally I'd like for us to be able to use the `DealClient` both with a FullNode and with a Gateway (i.e. only local Wallet), hence the import and modification of a few dependencies.

cmd/boost/deal_status_cmd.go
```diff
@@ -0,0 +1,178 @@
...
+			Addrs: maddrs,
+		}
+
+		if err := n.Host.Connect(ctx, *addrInfo); err != nil {
+			return fmt.Errorf("failed to connect to peer %s: %w", addrInfo.ID, err)
+		}
+
+		log.Debugw("storage provider on-chain address", "addr", minerAddr)
+
+		retryOpts := lp2pimpl.RetryParameters(time.Second, 1*time.Minute, 5, 5)
```
dirkmc, line 117: I suggest not retrying - if it fails to connect let's just return an error

cmd/boost/deal_status_cmd.go
```diff
@@ -0,0 +1,178 @@
...
+		msg += fmt.Sprintf("  publish cid: %s\n", resp.DealStatus.PublishCid)
+		msg += fmt.Sprintf("  chain deal id: %d\n", resp.DealStatus.ChainDealID)
+		fmt.Println(msg)
+
+		return nil
+	},
+}
+
+// statusMessage is based on dealResolver.Message
+func statusMessage(resp *types.DealStatusResponse) string {
```
dirkmc, line 147: Let's move this to common code between the resolver and here

cmd/boost/deal_status_cmd.go
```diff
@@ -0,0 +1,178 @@
...
+			if err != nil {
+				return fmt.Errorf("storage provider %s had invalid multiaddrs in their info: %w", maddr, err)
+			}
+			maddrs = append(maddrs, ma)
+		}
+		if len(maddrs) == 0 {
+			return fmt.Errorf("storage provider %s has no multiaddrs set on-chain", maddr)
+		}
+
+		log.Debugw("found storage provider", "id", *minfo.PeerId, "multiaddr", maddrs)
```
dirkmc, line 98: We can probably share a bunch of this code with the deal command

storagemarket/lp2pimpl/net.go
```diff
@@ -288,5 +291,7 @@ func (p *DealProvider) getDealStatus(req types.DealStatusRequest) types.DealStat
 			PublishCid:        pds.PublishCID,
 			ChainDealID:       pds.ChainDealID,
 		},
+		Transfer:       pds.Transfer,
```
dirkmc, line 294: The transfer may contain sensitive information, eg the Authorization header, so we may not want to include it here

cmd/boost/deal_status_cmd.go
nonsense: The data model is completely different, even if the logic is the same, so this might be hard, but will look into it.

dirkmc: Ok not a big deal, if it doesn't make sense let's leave it as-is

storagemarket/lp2pimpl/net.go
nonsense: Replaced just with TransferSize, so that we display progress in the command output.

dirkmc (approved):

PR #400, opened 2022-04-01
Downgrade from v1 to v0 api

dirkmc:


dirkmc:
Closing for now as before long we will probably need to upgrade back to v1

PR #401, opened 2022-04-01
Don't publish deal sealing status update if no one is subscribed

dirkmc:


storagemarket/deal_handler.go
```diff
@@ -31,14 +31,74 @@ type dealHandler struct {
...
+}
+
+// updatesSubscription wraps event.Subscription so that we can add an onClose
+// callback
+type updatesSubscription struct {
+	event.Subscription
+	onClose func(*updatesSubscription)
 }
 
+func (s *updatesSubscription) Close() error {
```
aarshkshah1992, line 64: Where do we call close on the subscription ?

storagemarket/deal_handler.go
```diff
@@ -31,14 +31,74 @@ type dealHandler struct {
...
+	defer d.updateSubsLk.Unlock()
+	d.updateSubs[updatesSub] = struct{}{}
+
+	return updatesSub, nil
+}
+
+// hasUpdateSubscribers indicates if anyone is subscribed to updates.
+// This is useful if we want to check if anyone is listening before doing an
+// expensive operation to publish an event.
+func (d *dealHandler) hasUpdateSubscribers() bool {
```
aarshkshah1992, line 98: nit: can we rename this to `hasActiveSubs` ? The current name is a bit tricky to follow.

storagemarket/deal_handler.go
dirkmc: Yes, it's called when the front end stops subscribing to events.
This can happen because
- the user closes the web page
- a new deal comes in that moves the last deal on the page to page 2 (ie no longer visible)

storagemarket/deal_handler.go
```diff
@@ -31,14 +31,74 @@ type dealHandler struct {
 	transferMu       sync.Mutex
 	transferFinished bool
 	transferErr      error
+
+	updateSubsLk sync.RWMutex
+	updateSubs   map[*updatesSubscription]struct{}
```
aarshkshah1992, line 36: nit: please can we rename this to `activeSubs` ?

aarshkshah1992 (commented):
some nits.

aarshkshah1992 (approved):

Issue #216, opened 2022-02-08
Refactor deal validation code so it's testable and write deal validation unit tests


aarshkshah1992:
We need to refactor the code https://github.com/filecoin-project/boost/blob/2963516e1be481984e38fb10e910d231dfa87ac6/storagemarket/deal_acceptance.go#L22 so it programs to interfaces which can be stubbed. After that, we need to write unit tests for the deal validation  flow.

PR #407, opened 2022-04-05
Deal validation tests

aarshkshah1992:
Closes https://github.com/filecoin-project/boost/issues/216.


itests/dummydeal_test.go
```diff
@@ -84,4 +84,11 @@ func TestDummydeal(t *testing.T) {
 	require.False(t, res.Accepted)
 	require.Contains(t, res.Reason, "storage price per epoch less than asking price")
 	log.Debugw("got response from MarketDummyDeal", "res", spew.Sdump(res))
+
+	// deal fails because signature is invalid
+	res, err = f.makeDummyDeal(dealUuid, carFilepath, rootCid, server.URL+"/"+filepath.Base(carFilepath), false, abi.NewTokenAmount(2000000), withInvalidSignature())
+	require.NoError(t, err)
+	require.False(t, res.Accepted)
+	require.Contains(t, res.Reason, "error verifying signature")
+	log.Debugw("got response from MarketDummyDeal", "res", spew.Sdump(res))
```
dirkmc, line 93: I think we don't really need the invalid signature and min ask tests at the dummy deal level if we're already testing these things in the provider unit test

aarshkshah1992: @dirkmc Dosen't hurt to have it for a sanity check as these are two important validations and these finish quickly as we don't wait for these to be sealed since they get rejected right off the bat.

dirkmc: I think I'd rather keep all the validation testing in one place. Someone looking at this code later might think they should add validation testing code here. In general I favour having the minimal amount of tests to keep the code maintainable.

aarshkshah1992: Okay, done.

dirkmc (approved):

Issue #414, opened 2022-04-07
Add an option to specify wallet for making deal


LexLuthr:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature or enhancement request. If it is, please open a [new idea discussion](https://github.com/filecoin-project/boost/discussions/new?category=ideas) instead. New feature and enhancement requests would be entertained by the boost team after a thorough discussion only.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to boost.

### Boost component

- [ ] boost daemon - storage providers
- [X] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### Boost Version

```text
boost 0.0.1+git.989bfbc
```

### Describe the Bug

Please add an option to specify the wallet to be used for making the deal. A client can have multiple wallets for different use cases and verified data cap.

Example:

NAME:
   boost deal - Make an online deal with Boost

USAGE:
   boost deal [command options] [arguments...]

OPTIONS:
  --wallet                              wallet address to be used for the deal  <-------------------------- New feature
   --http-url value                 http url to CAR file
   --http-headers value             http headers to be passed with the request (e.g key=value)
   --repo value                     repo directory for Boost client (default: "~/.boost-client")
   --provider value                 storage provider on-chain address
   --commp value                    commp of the CAR file
   --piece-size value               size of the CAR file as a padded piece (default: 0)
   --car-size value                 size of the CAR file (default: 0)
   --payload-cid value              root CID of the CAR file
   --start-epoch value              start epoch by when the deal should be proved by provider on-chain (default: 0)
   --duration value                 duration of the deal in epochs (default: 518400)
   --provider-collateral value      deal collateral that storage miner must put in escrow; if empty, the min collateral for the given piece size will be used (default: 0)
   --storage-price-per-epoch value  (default: 1)
   --verified                       whether the deal funds should come from verified client data-cap (default: true)
   --help, -h                       show help (default: false)
   

[...]

PR #422, opened 2022-04-08
initial wallet support for boost

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/414

### DIFF
--- a/cmd/boost/wallet_cmd.go
+++ b/cmd/boost/wallet_cmd.go
@@ -20,17 +20,16 @@ import (
 	"golang.org/x/xerrors"
 )
 
-var walletFlags = []cli.Flag{
-	&cli.StringFlag{
-		Name:  "repo",
-		Usage: "repo directory for Boost client",
-		Value: "~/.boost-client",
-	},
-}
-
 var walletCmd = &cli.Command{
 	Name:  "wallet",
 	Usage: "Manage wallets with Boost",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "repo",
+			Usage: "repo directory for Boost client",
+			Value: "~/.boost-client",
+		},
+	},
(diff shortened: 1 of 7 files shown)


cmd/boost/wallet_cmd.go
```diff
@@ -0,0 +1,476 @@
...
+				if cctx.Bool("id") {
+					id, err := api.StateLookupID(ctx, addr, types.EmptyTSK)
+					if err != nil {
+						row["ID"] = "n/a"
+					} else {
+						row["ID"] = id
+					}
+				}
+
+				if cctx.Bool("market") {
```
nonsense, line 174: This is the same as Lotus, but I think we could get rid of this argument, and always display the Market balances... WDYT?

I'd say a user cares about ALL their balances across a given wallet, or they care only about the addresses in which case they'd use `--addr-only=true`.

LexLuthr: Yeah, I would prefer to keep it minimal intervention. Always display market balance is better.

dirkmc: I think we should always display a market balance 👍 
I also agree it would be nice to be able to show wallets even when not connected to the internet, so let's also add the `--addr-only=true` flag as you suggested.

nonsense: Removed `market` flag.

cmd/boostx/utils_cmd.go
```diff
@@ -81,7 +81,7 @@ var marketCmd = &cli.Command{
 		}
 		defer closer()
 
-		walletAddr, err := n.Wallet.GetDefault()
```
dirkmc, line 84: Add wallet to parameters

dirkmc (approved):

PR #437, opened 2022-04-12
Use deal uuid in filename for download staging area file

dirkmc:
And remove filestore abstraction as we no longer need it

Screen Shot 2022-04-12 at 3 41 47 PM


storagemarket/provider_loop.go
```diff
@@ -124,6 +130,10 @@ func (p *Provider) processDealProposal(deal *types.ProviderDealState) *acceptErr
 		} else if errs == nil {
 			p.dealLogger.Infow(deal.DealUuid, "untagged storage for deal cleanup", deal.Transfer.Size)
 		}
+
+		if deal.InboundFilePath != "" {
```
nonsense, line 134: Why do we need this here? How come we never had this cleanup here so far?

nonsense, line 134: Why do we need this here? How come we never had this cleanup here so far?

dirkmc: This used to live in provider.go

The flow was:
1. Receive a deal proposal over the network
2. Create a file where downloaded data is stored
3. Send a message to the provider loop to check for deal acceptance
4. If deal is rejected / errors, clean up download data file.

I've changed the download file name to be the uuid of the deal (instead of a random file name). So now we need to create the file synchronously with where we add things to the database, in case there are two concurrent requests with the same deal uuid. The place we add things to the database is in the provider loop. So now the flow is:
1. Receive a deal proposal over the network
2. Send a message to the provider loop to check for deal acceptance
3. Create a file where downloaded data is stored
4. If deal is rejected / errors, clean up download data file.

nonsense (approved):

PR #462, opened 2022-04-19
initial migrations support for sqlite3 with goose

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/340

This PR is adding migrations support for our sqlite3 database. I am using `goose` (https://pressly.github.io/goose/) for migrations support, as it seems well-maintained and supports all functionality we care about:
- up/down SQL migrations
- embedding migrations in Go
- CLI tools to do up/down/version checks on the sqlite3 database
- etc.

TODO:
- [x] add dev documentation on how to create migrations - opened a PR for review at https://app.gitbook.com/o/-L_E2woSLfhpBp2IPz9k/s/IurYwdy3AC3rjLVMFZRw/~/changes/lIDj1FBSrzn4N4So7cBD/boost-architecture/getting-started-1 for Boost documentation website
- [x] improve logging around `before` and `after` when we initialise the provider with an SQL db


db/migrations/20220419132443_addcheckpointattodeals.sql
```diff
@@ -0,0 +1,10 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE Deals
+  ADD CheckpointAt DateTime;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+SELECT 'down SQL query';
```
nonsense, line 9: No proper `down` migration here for now.

db/migrate.go
```diff
@@ -0,0 +1,30 @@
...
+var embedMigrations embed.FS
+
+func Migrate(db *sql.DB) error {
+	goose.SetBaseFS(embedMigrations)
+
+	if err := goose.SetDialect("sqlite3"); err != nil {
+		return err
+	}
+
+	err := goose.Version(db, "migrations")
```
nonsense, line 20: Unfortunately this API call just prints out the version via the default `log`, rather than returning it :/

dirkmc (approved):

PR #480, opened 2022-04-22
add non-interactive flag to boostx

nonsense:


cmd/boostx/utils_cmd.go
```diff
@@ -40,12 +40,21 @@ import (
 	"golang.org/x/xerrors"
 )
 
+var (
+	flagAssumeYes = &cli.BoolFlag{
+		Name:    "assume-yes",
```
nonsense, line 45: Borrowed from `apt-get`.

dirkmc (approved):

Issue #449, opened 2022-04-15
CLI: Add commands to query storage and retrieval ask


dirkmc:
The boost client should have equivalents to these lotus commands:
```
lotus client query-ask
lotus client retrieval-ask
```

Put these commands under the `boost provider` category

PR #483, opened 2022-04-25
provider cmds: storage-ask and retrieval-ask

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/449

TODO:
- [x] storage ask over libp2p
- [x] retrieval ask over libp2p

### DIFF
--- a/cmd/boost/provider_cmd.go
+++ b/cmd/boost/provider_cmd.go
@@ -205,13 +205,8 @@ var storageAskCmd = &cli.Command{
 var retrievalAskCmd = &cli.Command{
 	Name:      "retrieval-ask",
 	Usage:     "Query a storage provider's retrieval ask",
-	ArgsUsage: "[data CID]",
+	ArgsUsage: "[provider] [data CID]",
 	Flags: []cli.Flag{
-		&cli.StringFlag{
-			Name:     "provider",
-			Usage:    "storage provider on-chain address",
-			Required: true,
-		},
 		&cli.Int64Flag{
 			Name:  "size",
 			Usage: "data size in bytes",
@@ -221,8 +216,8 @@ var retrievalAskCmd = &cli.Command{
 		ctx := bcli.ReqContext(cctx)
 
 		afmt := NewAppFmt(cctx.App)
-		if cctx.NArg() != 1 {
-			afmt.Println("Usage: retrieval-ask [data CID]")
+		if cctx.NArg() != 2 {
+			afmt.Println("Usage: retrieval-ask [provider] [data CID]")
 			return nil
 		}
 
@@ -237,12 +232,12 @@ var retrievalAskCmd = &cli.Command{
 		}
 		defer closer()
 
-		maddr, err := address.NewFromString(cctx.String("provider"))
+		maddr, err := address.NewFromString(cctx.Args().First())
 		if err != nil {
 			return err
 		}
 
-		dataCid, err := cid.Parse(cctx.Args().First())
+		dataCid, err := cid.Parse(cctx.Args().Get(1))
 		if err != nil {
 			return fmt.Errorf("parsing data cid: %w", err)
 		}
--- a/cmd/boost/provider_cmd.go
+++ b/cmd/boost/provider_cmd.go
@@ -266,16 +266,13 @@ var queryRetrievalAskCmd = &cli.Command{
 			QueryParams: retrievalmarket.QueryParams{},
 		}
 
-		var resp retrievalmarket.QueryResponse
+		var ask retrievalmarket.QueryResponse
 
-		if err := doRpc(ctx, s, &req, &resp); err != nil {
+		if err := doRpc(ctx, s, &req, &ask); err != nil {
 			return fmt.Errorf("send retrieval-ask request rpc: %w", err)
 		}
 
-		ask := resp
-
 		afmt.Printf("Status: %d\n", ask.Status)
-
 		if ask.Status != 0 {
 			return nil
 		}
--- a/cmd/boost/provider_cmd.go
+++ b/cmd/boost/provider_cmd.go
@@ -112,7 +112,7 @@ var libp2pInfoCmd = &cli.Command{
 
 var queryAskCmd = &cli.Command{
 	Name:      "query-ask",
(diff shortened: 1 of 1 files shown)


cmd/boost/provider_cmd.go
```diff
@@ -97,3 +103,150 @@ var libp2pInfoCmd = &cli.Command{
...
+			afmt.Println("Usage: query-ask [minerAddress]")
+			return nil
+		}
+
+		maddr, err := address.NewFromString(cctx.Args().First())
+		if err != nil {
+			return err
+		}
+
+		api, closer, err := lcli.GetFullNodeAPI(cctx)
```
nonsense, line 137: Note that unlike other `boost` cmds, this one expects a Full Node, not just a Gateway.

nonsense: @dirkmc this is no longer the case, we query for asks directly over libp2p and just use the gateway to fetch multiaddrs for the miner.

cmd/boost/provider_cmd.go
```diff
@@ -97,3 +109,192 @@ var libp2pInfoCmd = &cli.Command{
 		return nil
 	},
 }
+
+var queryAskCmd = &cli.Command{
+	Name:      "query-ask",
+	Usage:     "Find a miners ask",
+	ArgsUsage: "[minerAddress]",
```
dirkmc, line 116: In the libp2p-info command we call the argument "provider address". Happy to use either but suggest we change them all to be consistent.

cmd/boost/provider_cmd.go
```diff
@@ -97,3 +109,192 @@ var libp2pInfoCmd = &cli.Command{
 		return nil
 	},
 }
+
+var queryAskCmd = &cli.Command{
+	Name:      "query-ask",
+	Usage:     "Find a miners ask",
```
dirkmc, line 115:
```suggestion
	Usage:     "Query a storage provider's storage ask",
```

cmd/boost/provider_cmd.go
```diff
@@ -97,3 +109,192 @@ var libp2pInfoCmd = &cli.Command{
...
+			Name:  "peerid",
+			Usage: "specify peer ID of node to make query against",
+		},
+		&cli.Int64Flag{
+			Name:  "size",
+			Usage: "data size in bytes",
+		},
+		&cli.Int64Flag{
+			Name:  "duration",
+			Usage: "deal duration",
```
dirkmc, line 128:
```suggestion
			Usage: "deal duration in epochs",
```

cmd/boost/provider_cmd.go
```diff
@@ -97,3 +109,192 @@ var libp2pInfoCmd = &cli.Command{
 		return nil
 	},
 }
+
+var queryAskCmd = &cli.Command{
+	Name:      "query-ask",
```
dirkmc, line 114: Suggest we call the commands "storage-ask" and "retrieval-ask"

cmd/boost/provider_cmd.go
```diff
@@ -97,3 +109,192 @@ var libp2pInfoCmd = &cli.Command{
...
+		}
+		afmt.Printf("Total Price: %s\n", types.FIL(types.BigMul(perEpoch, types.NewInt(uint64(duration)))))
+
+		return nil
+	},
+}
+
+var queryRetrievalAskCmd = &cli.Command{
+	Name:      "retrieval-ask",
+	Usage:     "Get a miner's retrieval ask",
```
dirkmc, line 210:
```suggestion
	Usage:     "Query a storage provider's retrieval ask",
```

cmd/boost/provider_cmd.go
```diff
@@ -97,3 +109,192 @@ var libp2pInfoCmd = &cli.Command{
...
+
+		if err := doRpc(ctx, s, &req, &ask); err != nil {
+			return fmt.Errorf("send retrieval-ask request rpc: %w", err)
+		}
+
+		afmt.Printf("Status: %d\n", ask.Status)
+		if ask.Status != 0 {
+			return nil
+		}
+		afmt.Printf("Ask: %s\n", maddr)
```
dirkmc, line 280: Suggest option for json output here too

cmd/boost/provider_cmd.go
```diff
@@ -97,3 +109,192 @@ var libp2pInfoCmd = &cli.Command{
...
+			Miner: maddr,
+		}
+
+		if err := doRpc(ctx, s, &askRequest, &resp); err != nil {
+			return fmt.Errorf("send ask request rpc: %w", err)
+		}
+
+		ask := resp.Ask.Ask
+
+		afmt.Printf("Ask: %s\n", maddr)
```
dirkmc, line 185: Suggest we also parse the "json" flag and if it's true, output the results as json (see line 91)

cmd/boost/provider_cmd.go
```diff
@@ -97,3 +109,192 @@ var libp2pInfoCmd = &cli.Command{
...
+
+var queryAskCmd = &cli.Command{
+	Name:      "query-ask",
+	Usage:     "Find a miners ask",
+	ArgsUsage: "[minerAddress]",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "peerid",
+			Usage: "specify peer ID of node to make query against",
+		},
```
dirkmc, line 121: It seems like the peerid flag is ignored in the code, and we don't have a peerid flag for retrieval ask, so I suggest we just delete it

dirkmc (approved):
Suggested a few minor changes but overall LGTM 👍

cmd/boost/provider_cmd.go
nonsense: I was going to raise that up! I copied the commands from Lotus, but yeah, I also think we should be consistent.. Changing this now.

dirkmc:
For convenience I think it makes sense for these commands to take the provider address as a command line parameter rather than a flag:
```
storage-ask <provider addr>
rather than
storage-ask --provider=<provider addr>
```

nonsense:
@dirkmc I'll merge as is and tackle the JSON output across all commands in Boost, not just these two.

PR #489, opened 2022-04-28
Use RootDir when migrating Dagstore

LexLuthr:
Reset RootDir path under Dagstore config in boost cfg to use default dagstore path in boost
Include a check to find dagstore under Rootdir if set.

Based on https://filecoinproject.slack.com/archives/C03CKDLEWG1/p1651012378785659


nonsense (approved):

nonsense:
Good catch, I don't understand why we needed this config param in the first place.

LexLuthr:
> Good catch, I don't understand why we needed this config param in the first place.

dagstore can grow to a large size for some SPs. Nicklas from my team is around 200G so they could have placed it in separate storage in lotus config.

cmd/boostd/init.go
```diff
@@ -382,6 +400,7 @@ func migrateMarketsConfig(cctx *cli.Context, mktsRepo lotus_repo.LockedRepo, boo
 		rcfg.LotusDealmaking = mktsCfg.Dealmaking
 		rcfg.LotusFees = mktsCfg.Fees
 		rcfg.DAGStore = mktsCfg.DAGStore
+		rcfg.DAGStore.RootDir = ""
```
nonsense, line 403: Given that some people will copy the dagstore over, whereas other people will move it, we should probably set this var depending on what they pick.

dirkmc:
As discussed sync, let's adopt this approach instead: https://github.com/filecoin-project/boost/issues/496

LexLuthr:
Closing it due to conflicts in favor of https://github.com/filecoin-project/boost/pull/502

PR #500, opened 2022-05-04
Apply new design to react app

dirkmc:


jacobheun:
We need to kick off another round of design to account for all the features we added during the initial phase.

jacobheun:
Closing this as we'll need a design refresh.

PR #505, opened 2022-05-06
update lotus to v1.15.2

dirkmc:


LexLuthr:
@dirkmc v1.15.2 stable release is out. Can we please use stable releases instead of RC?

go.mod
```diff
@@ -72,8 +72,8 @@ require (
 	github.com/jpillora/backoff v1.0.0
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/libp2p/go-eventbus v0.2.1
-	github.com/libp2p/go-libp2p v0.18.1
-	github.com/libp2p/go-libp2p-core v0.14.0
+	github.com/libp2p/go-libp2p v0.19.0
```
magik6k, line 75: You probably want the latest libp2p tag

node/builder.go
```diff
@@ -283,6 +289,14 @@ func ConfigCommon(cfg *config.Common) Option {
 			Override(new(lotus_dtypes.BootstrapPeers), modules.ConfigBootstrap(cfg.Libp2p.BootstrapPeers)),
 		),
 
+		Override(new(network.ResourceManager), lp2p.ResourceManager(cfg.Libp2p.ConnMgrHigh)),
```
magik6k, line 292: You also need the ResourceManagerKey thing from lotus to actually pass the resource manager option to libp2p

magik6k (approved):

Issue #446, opened 2022-04-14
Refactor: allow the user to manually retry / fail deals that have paused with a recoverable error


dirkmc:
Once the data for a deal has been transferred, Boost will publish the deal, add it to a sector, and index and announce the deal.

#### Recoverable vs Non-recoverable errors
At the moment we distinguish between "recoverable" and "non-recoverable" errors:
- Non-recoverable errors immediately fail the deal and the deal data is cleaned up
- Recoverable errors stop deal execution, but don't fail the deal. When Boost restarts it attempts to continue executing any deals that are in a recoverable error state.

#### Proposal

In practice it's not really possible to distinguish between recoverable and non-recoverable errors that come from other subsystems (publishing, sealing, indexing).

Instead, we propose to:
- Treat all deal errors as recoverable, with one exception:
  - if the deal's start epoch is after the current epoch, immediately fail the deal
- When Boost restarts, attempt to restart deals that are in the recoverable error state
- Add a mechanism in the web UI that allows the user to either
  - Immediately fail a deal
  - Immediately retry a deal

PR #521, opened 2022-05-12
Retry deal

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/446

TODO:
- [ ] Refactor deal handler construct in Provider
- [ ] Implement `Provider.RetryDeal`
- [ ] Add graphql endpoint to retry deal
- [ ] Add web UI button to retry deal


storagemarket/deal_execution.go
```diff
@@ -127,58 +149,70 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, pub event.Emitter,
...
-			return &dealMakingError{err: fmt.Errorf("error when matching commP for imported data for offline deal: %w", err)}
+			return &dealMakingError{
+				retry: types.DealRetryFatal,
+				error: fmt.Errorf("error when matching commP for imported data for offline deal: %w", err),
+			}
 		}
 		p.dealLogger.Infow(deal.DealUuid, "commp matched successfully for imported data for offline deal")
 
 		// update checkpoint
 		if err := p.updateCheckpoint(pub, deal, dealcheckpoints.Transferred); err != nil {
```
nonsense, line 207: Again: `err` -> `derr`.

storagemarket/deal_execution.go
```diff
@@ -585,13 +663,18 @@ func (p *Provider) indexAndAnnounce(ctx context.Context, pub event.Emitter, deal
 	// announce to the network indexer but do not fail the deal if the announcement fails
 	annCid, err := p.ip.AnnounceBoostDeal(ctx, deal)
 	if err != nil {
-		p.dealLogger.LogError(deal.DealUuid, "failed to announce deal to network indexer "+
-			"but not failing deal as it's already been handed for sealing", err)
-	} else {
-		p.dealLogger.Infow(deal.DealUuid, "announced deal to network indexer", "announcement-cid", annCid)
+		return &dealMakingError{
```
nonsense, line 666: @dirkmc I am not sure we want to be failing the deal if the announcement fails. We are changing behaviour here right now - is this intended?

storagemarket/provider_loop.go
```diff
@@ -442,6 +446,23 @@ func (p *Provider) run() {
...
+				retryDealReq.done <- err
+				return
+			}
+
+			// start executing the deal
+			if p.startDealThread(deal) {
+				// if the deal wasn't already running, log a message saying
+				// that it was restarted
+				p.dealLogger.Infow(deal.DealUuid, "user initiated deal retry", "checkpoint", deal.Checkpoint)
+			}
```
nonsense, line 462: Maybe log an else statement that `deal is already running`? I think the absence of a log line will make it harder to figure out what's going on.

storagemarket/deal_execution.go
dirkmc: The retry type is DealRetryAuto, so the deal doesn't actually get failed. The deal is paused until either the user retries it manually, or boost automatically retries it the next time boost restarts.
Note that index and announce is the last step in the deal execution (it has already been handed off to the sealer)

nonsense: > Note that index and announce is the last step in the deal execution (it has already been handed off to the sealer)

Oh, OK, I missed that. Then I guess we are good!

nonsense (approved):

PR #525, opened 2022-05-13
refactor: move deal handling code into one place

dirkmc:


nonsense (approved):
nice work! i think this is simpler than what we have right now.

PR #529, opened 2022-05-18
add boost backup & restore commands

LexLuthr:


cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/filecoin-project/lotus/lib/backupds"
+	lotus_repo "github.com/filecoin-project/lotus/node/repo"
+	dstore "github.com/ipfs/go-datastore"
+)
+
+var fm = map[string]fs.FileMode{"api": 0644, "boost.db": 0644, "boost.logs.db": 0644, "config.toml": 0664, "storage.json": 0644, "token": 0600}
+
+var backupCmd = &cli.Command{
+	Name:  "backup",
+	Usage: "Perorms offline backup of the Boost subsystem",
```
dirkmc, line 27:
```suggestion
	Usage: "Performs offline backup of Boost",
```

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+	Flags: []cli.Flag{
+		&cli.PathFlag{
+			Name:     "path",
+			Usage:    "Specify a directory to create backup. Must be inside home directory.",
```
dirkmc, line 31: Why does it have to be inside the home directory?

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+		if err != nil {
+			return err
+		}
+		if !ok {
+			return fmt.Errorf("repo at '%s' is not initialized", cctx.String(FlagBoostRepo))
+		}
+
+		lr, err := r.LockRO(node.Boost)
+		if err != nil {
+			return fmt.Errorf("locking repo: %w", err)
```
dirkmc, line 54: IIRC usually locking the node fails because it's already locked by a running boostd process. So we should let the user know that they need to stop boostd before running the backup.

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		defer lr.Close()
+
+		mds, err := lr.Datastore(context.TODO(), "/metadata")
```
dirkmc, line 58:
```suggestion
		mds, err := lr.Datastore(cctx.Context, "/metadata")
```

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+
+		if err := bds.Backup(cctx.Context, out); err != nil {
+			if cerr := out.Close(); cerr != nil {
+				log.Errorw("error closing backup file while handling backup error", "closeErr", cerr, "backupErr", err)
+			}
+			return fmt.Errorf("backup error: %w", err)
+		}
+
+		if err := out.Close(); err != nil {
```
dirkmc, line 98: This `out.Close()` block should probably be inside a `defer`, so that it happens even if the Backup fails

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+			return fmt.Errorf("closing backup file: %w", err)
+		}
+
+		for name, perm := range fm {
```
dirkmc, line 102: I suggest we read the permissions from the source file, instead of hard-coding them.

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+		if err != nil {
+			return err
+		}
+
+		bpath, err := homedir.Expand(cctx.Path("path"))
+		if err != nil {
+			return fmt.Errorf("expanding backup directory path: %w", err)
+		}
+
+		bkp_dir := path.Join(bpath, "boost_backup_"+time.Now().Format("20060102150405"))
```
dirkmc, line 73: In our code we use camelCase rather than snake_case. Let's change all the variables to be camel case.
```suggestion
		bkpDir := path.Join(bpath, "boost_backup_"+time.Now().Format("20060102150405"))
```

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+var restoreCmd = &cli.Command{
+	Name:  "restore",
+	Usage: "Restores a boost repository from backup",
+	Flags: []cli.Flag{
+		&cli.PathFlag{
+			Name:     "restore-path",
+			Usage:    "Specify the path to backup directory",
+			Required: true,
+		},
+	},
```
dirkmc, line 136: Let's make this a required command line argument rather than a flag

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+			return fmt.Errorf("expanding backup directory path: %w", err)
+		}
+
+		bkp_dir := path.Join(bpath, "boost_backup_"+time.Now().Format("20060102150405"))
+
+		if err := os.Mkdir(bkp_dir, 0755); err != nil {
+			return fmt.Errorf("Error creating backup directory %s: %w", bkp_dir, err)
+		}
+
+		fpath_name := path.Join(bkp_dir, "metadata")
```
dirkmc, line 79: Let's use a constant instead of `"metadata"`

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+
+		fmt.Println("Resetting chainstore metadata")
+
+		chainHead := dstore.NewKey("head")
+		if err := mds.Delete(cctx.Context, chainHead); err != nil {
+			return fmt.Errorf("clearing chain head: %w", err)
+		}
+		if err := store.FlushValidationCache(cctx.Context, mds); err != nil {
+			return fmt.Errorf("clearing chain validation cache: %w", err)
+		}
```
dirkmc, line 220: I don't think we need this - boost shouldn't be storing any chain data

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+
+		ok, err := r.Exists()
+		if err != nil {
+			return err
+		}
+		if ok {
+			return fmt.Errorf("repo at '%s' is already initialized, cannot restore.", repoPath)
+		}
+
+		fmt.Println("Creating boost repo")
```
dirkmc, line 156: I like these notes to the user telling them what the command is doing 👍 
Let's add the same thing to the backup command as well.

dirkmc (commented):
Looks good so far 👍 
I made a few suggestions.

Could you also please post the output of running the backup and restore commands in the description of this PR.

cmd/boostd/backup.go
```diff
@@ -0,0 +1,274 @@
...
+		if err != nil {
+			return err
+		}
+
+		err = ioutil.WriteFile(dest, input, perm)
+		if err != nil {
+			return err
+		}
+	} else {
+		fmt.Printf("Not copying %s as file does not exists\n", src)
```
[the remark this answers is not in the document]
LexLuthr: Here, I am checking if a file exists or not. Let's say .boost/storage.json is not there for any reason. Then, I still want to do the backup but let the user know that said file does not exists at the source. I will as a little clarification.

cmd/boostd/backup.go
LexLuthr: I have put it under an anonymous func with defer. Code looks like a crude but it should work.

cmd/boostd/backup.go
```diff
@@ -0,0 +1,276 @@
...
+	"storage.json",
+	"token"}
+
+var backupCmd = &cli.Command{
+	Name:   "backup",
+	Usage:  "Performs offline backup of Boost",
+	Before: before,
+	Action: func(cctx *cli.Context) error {
+		if cctx.Args().Len() > 1 {
+			return fmt.Errorf("backup only takes one argument (custom location inside home directory)")
```
dirkmc, line 34:
```suggestion
		if cctx.Args().Len() != 1 {
			return fmt.Errorf("usage: boostd backup <backup directory>")
```

cmd/boostd/backup.go
```diff
@@ -0,0 +1,276 @@
...
+var fm = []string{"api",
+	"boost.db",
+	"boost.logs.db",
+	"config.toml",
+	"storage.json",
+	"token"}
+
+var backupCmd = &cli.Command{
+	Name:   "backup",
+	Usage:  "Performs offline backup of Boost",
```
dirkmc, line 30:
```suggestion
	Usage: "boostd backup <backup directory>"
	Description:  "Performs offline backup of Boost",
```

cmd/boostd/backup.go
```diff
@@ -0,0 +1,276 @@
...
+		out, err := os.OpenFile(fpath, os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			return fmt.Errorf("opening backup file %s: %w", fpath, err)
+		}
+
+		defer func(*os.File) {
+			if err := out.Close(); err != nil {
+				log.Errorw("closing backup file: %w", err)
+			}
+		}(out)
```
dirkmc, line 98:
```suggestion
		defer func() {
			if err := out.Close(); err != nil {
				log.Errorw("closing backup file: %w", err)
			}
		}()
```

cmd/boostd/backup.go
```diff
@@ -0,0 +1,276 @@
...
+
+		if err := bds.Backup(cctx.Context, out); err != nil {
+			if cerr := out.Close(); cerr != nil {
+				log.Errorw("error closing backup file while handling backup error", "closeErr", cerr, "backupErr", err)
+			}
```
dirkmc, line 103:
```suggestion
```

cmd/boostd/backup.go
```diff
@@ -0,0 +1,276 @@
...
+
+			if err := copy_files(srcPath, destPath); err != nil {
+				return fmt.Errorf("error copying file %s: %w", srcName, err)
+			}
+
+		}
+
+		fmt.Println("Boost repo successfully backed up at " + bkpDir)
+		fmt.Println("You can now start boost with 'boostd -vv run'")
```
dirkmc, line 131:
```suggestion
```

cmd/boostd/backup.go
```diff
@@ -0,0 +1,276 @@
...
+
+		return nil
+	},
+}
+
+var restoreCmd = &cli.Command{
+	Name:   "restore",
+	Usage:  "Restores a boost repository from backup",
```
dirkmc, line 139:
```suggestion
	Usage:  "boostd restore <backup dir>",
	Description:  "Restores a boost repository from backup",
```

cmd/boostd/backup.go
```diff
@@ -0,0 +1,276 @@
...
+			}
+
+			if err := copy_files(srcPath, destPath); err != nil {
+				return fmt.Errorf("error copying file %s: %w", srcName, err)
+			}
+
+		}
+
+		fmt.Println("Boost repo successfully restored at " + lr.Path())
+		fmt.Println("You can now start boost with 'boostd -vv run'")
```
dirkmc, line 247:
```suggestion
```

cmd/boostd/backup.go
dirkmc: I think people will quite often want to backup files to somewhere other than their homedir (eg to an external drive).
We can use it up to the user to figure out user permissions.

LexLuthr: It will work outside of home directory as well. If a user providers an absolute path, back would be done there.

cmd/boostd/backup.go
```diff
@@ -0,0 +1,273 @@
...
+var fm = []string{"api",
+	"boost.db",
+	"boost.logs.db",
+	"config.toml",
+	"storage.json",
+	"token"}
+
+var backupCmd = &cli.Command{
+	Name:   "backup",
+	Usage: "boostd backup <backup directory>"
```
LexLuthr, line 30:
```suggestion
	Usage: "boostd backup <backup directory>",
```

cmd/boostd/backup.go
```diff
@@ -119,28 +119,47 @@ var backupCmd = &cli.Command{
...
+		}
+
+		fmt.Println("Verifying backup")
+
+		flist := []string{"metadata", "boost.db", "boost.logs.db"}
+		for _, fileName := range flist {
+			_, err = os.Stat(path.Join(bpath, fileName))
+			if os.IsNotExist(err) {
+				return fmt.Errorf("could not locate suitable backup files: %w", err)
+			}
```
dirkmc, line 160:
```suggestion
			} else if (err != nil) {
				return fmt.Errorf("getting status of %s: %w", fileName, err)
			}
```

cmd/boostd/backup.go
```diff
@@ -119,28 +119,47 @@ var backupCmd = &cli.Command{
...
 		if cctx.Args().Len() != 1 {
 			return fmt.Errorf("restore only takes one argument (backup directory path)")
 		}
 
+		bpath, err := homedir.Expand(cctx.Args().First())
+		if err != nil {
+			return fmt.Errorf("expanding backup directory path: %w", err)
+		}
+
+		fmt.Println("Verifying backup")
```
dirkmc, line 153:
```suggestion
		fmt.Printf("Checking backup directory %s\n", bpath)
```

cmd/boostd/backup.go
```diff
@@ -271,3 +289,49 @@ func copy_files(src, dest string) error {
 
 	return nil
 }
+
+func copyKeys(src, dest string) error {
```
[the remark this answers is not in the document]
LexLuthr: changed

dirkmc (approved):

PR #536, opened 2022-05-19
boost stop command

LexLuthr:


LexLuthr:
Closing for now as we don't want to ship a shutdown for now.

PR #541, opened 2022-05-24
Calculate commp on sealing node

dirkmc:
Note this will **only work with lotus v1.16** and above.

Allows the SP to configure boost to assign commp calculation to the sealing node instead of doing commp on the local boost node.

To enable, set the following config parameter:
```
[Dealmaking]
   RemoteCommp = true
```


storagemarket/helper.go
```diff
@@ -151,13 +151,13 @@ func (c *ChainDealManager) dealIDFromPublishDealsMsg(ctx context.Context, tok ct
 		return dealID, ctypes.EmptyTSK, fmt.Errorf("could not find deal in publish deals message %s", publishCid)
 	}
 
-	if dealIdx >= len(dealIDs) {
```
dirkmc, line 154: The changes to this helper are required because the return values of IsDealValid have changed.

nonsense (approved):
lgtm

PR #554, opened 2022-06-02
Piece Store

dirkmc:
New Piece Store implementation

### DIFF
--- a/piecestore/piecestore.go
+++ b/piecestore/piecestore.go
@@ -28,7 +28,7 @@ type Sealer interface {
 	GetReader(id abi.SectorNumber, offset abi.PaddedPieceSize, length abi.PaddedPieceSize) (SectionReader, error)
 }
 
-// CarIndexes keeps indexes that map from multihash => offset
+// CarIndexes manages indexes of CAR files that map from multihash => offset
 type CarIndexes interface {
 	Add(pieceCid cid.Cid, itidx carindex.IterableIndex) error
 	IsIndexed(pieceCid cid.Cid) bool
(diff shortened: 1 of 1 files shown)


piecestore/piecestore.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return nil
+}
+
+func (ps *PieceStore) addIndexForPiece(pieceCid cid.Cid) error {
+	// Check if the indexes have already been added
+	if ps.carIndex.IsIndexed(pieceCid) && ps.mhToPieceIndex.IsIndexed(pieceCid) {
+		return nil
+	}
+
```
nonsense, line 94: @dirkmc maybe we should add debug timers here, so that we have an idea how long it takes to fetch the piece and build an index.

piecestore/piecestore.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return fmt.Errorf("generating index for piece %s: %w", pieceCid, err)
+	}
+
+	itidx, ok := idx.(carindex.IterableIndex)
+	if !ok {
+		return fmt.Errorf("index is not iterable for piece %s", pieceCid)
+	}
+
+	if !ps.carIndex.IsIndexed(pieceCid) {
```
nonsense, line 112: I guess you are checking a second time here in case of another job that races this one, but I'd say this is redundant if the operation above is relatively "fast"... same for the check blow.

piecestore/piecestore.go
dirkmc: Yeah good idea to add some kind of tracing

dirkmc:
Closing in favour of https://github.com/filecoin-project/boost/pull/573

PR #558, opened 2022-06-03
When disabled in config, completely disable index provider

dirkmc:
If the index provider is disabled in config, don't try to start it up.
At the announce step, if the index provider is disabled, skip making an announcement.


nonsense:
@dirkmc isn't the indexer not a "critical" piece of functionality for SPs? I think that if it is disabled, we shouldn't be crashing all deals, or?

dirkmc:
Changed deal execution so that it no longer attempts to announce a deal if the index provider is disabled.

nonsense (approved):

PR #573, opened 2022-06-07
piece meta service (`boostd-data`) initial implementation

nonsense:


LaurenSpiegel:
> This PR is heavy WIP. We cannot merge and release this work, before we have very high confidence on the future data model for the PieceMeta store as well as its performance, alongside migrations from current `boost` to it.
> 
> It must:
> 
> * [ ]  🚧 produce data model for `Store` for both implementations (based loosely on https://docs.google.com/document/d/15ziYCc56q8xsr9P1nfj7dbpDZ5h_ZZWGYxH93RiLgyM/edit)
> * [ ]  ❌ add migrations to get from DAGStore and PIeceStore to `PieceMeta` within `boostd-data`
> * [ ]  ❌ performance benchmarks for the `PieceMeta` and `Store` with `boost` running LevelDB and Couchbase backend for small-, mid- and large- size validators (tbd concretely how many GiB worth of deals these entail) ; [Benchmarks / Scaling requirements analysis for boostd-data state service and PieceMeta #615](https://github.com/filecoin-project/boost/issues/615)
> 
> agenda: 🚧 - in-progress ❌ - not started

1) What is the difference between `Store` and `PieceMeta`?
2) Where did the requirement to support both leveldb and couchbase come from?

docker/devnet/boost/entrypoint.sh
```diff
@@ -45,6 +45,9 @@ if [ ! -f $BOOST_PATH/.init.boost ]; then
 	touch $BOOST_PATH/.init.boost
 fi
 
+# TODO(anteva): fixme: hack as boostd fails to start without this dir
+mkdir -p /var/lib/boost/deal-staging
```
jacobheun, line 49: We need to fix this before we release LID right? Otherwise people will need to create this directory manually?

jacobheun: Yeah, since we moved in some of the LID code to main we needed to add the line. Not sure what the root cause of it is.

nonsense:
merged into `release/v2` (which was branched out of `main`)

PR #583, opened 2022-06-10
Update devnet docs

dirkmc:


Makefile
```diff
@@ -101,7 +101,7 @@ boostci: $(BUILD_DEPS)
 .PHONY: boostci
 
 react: check-node-lts
-	npm install --no-audit --prefix react
+	npm install --no-audit --legacy-peer-deps --prefix react
```
rvagg, line 104: I _think_ this is safe on older versions of npm, but there is the option of just adding a `npm_config_legacy_peer_deps=yes` before `npm` to do it via environment variable which will be safely ignored if it's not supported.

Makefile
```diff
@@ -101,7 +101,7 @@ boostci: $(BUILD_DEPS)
 .PHONY: boostci
 
 react: check-node-lts
-	npm install --no-audit --prefix react
+	npm_config_legacy_peer_deps=yes npm install --no-audit --legacy-peer-deps --prefix react
```
rvagg, line 104:
```suggestion
	npm_config_legacy_peer_deps=yes npm install --no-audit --prefix react
```

documentation/devnet.md
```diff
@@ -115,22 +115,53 @@ boostd -vv init \
...
+  ListenAddresses = ["/ip4/0.0.0.0/tcp/50000", "/ip6/::/tcp/0"]
+```
+
+13. Run boost
 ```
 boostd -vv run
 ```
 
-12. Build the Web UI
+Note down the peer ID of the boost instance
```
rvagg, line 138:
```suggestion
Note down the peer ID of the boost instance, it will be listed in the log output as `"maddr"` on a line that says `Boost libp2p node listening`.
```

documentation/devnet.md
```diff
@@ -115,22 +115,58 @@ boostd -vv init \
...
+2. Send funds to the client wallet
+```
+lotus send --from=$DEFAULT_WALLET <client wallet> 10
+```
+
+3. Follow the guide at https://boost.filecoin.io/tutorials/how-to-store-files-with-boost-on-filecoin
 
 Note that above you already ran a command to export FULLNODE_API (and point it to your local devnet lotus daemon).
+
+Note also that the provider address is `t01000`
```
rvagg, line 172:
```suggestion
Note also that the provider address is `t01000` and you will need to supply an appropriate `--storage-price` when using `boost deal` since the devnet has a minimum price of `3814` per epoch.
```

rvagg:
```suggestion
Note also that the provider address is `t01000` and you will need to supply an appropriate `--storage-price` when using `boost deal` since the devnet has a minimum price. Alternatively, using "Settings" in the Boost web UI to set the deal price to zero. 
```

Issue #587, opened 2022-06-10
Update new config parameters in config file on upgrade


LexLuthr:
If a new parameter is introduced in configuration, it is not copied over to the config.toml and just default value is picked up from within the code. Only way is to init a separate directory, move new file to repo and copy over values for old params from older config file.

We need to make sure these new params are also updated in the config file to ensure they are visible to the users.

Context: This is because we are using lotus code for config and we might have to copy over the code and refactor it for boost to implement this.

PR #589, opened 2022-06-13
migrate config to v1

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/587

Version 1 of config
- Adds a version number to the config file
- Adds comments to config

For example:
```
# The version of the config file (used for migrations)
# 
# type: int
# env var: LOTUS__CONFIGVERSION
ConfigVersion = 1

[Libp2p]
  # Binding address for the libp2p host - 0 means random port.
  # Format: multiaddress; see https://multiformats.io/multiaddr/
  #
  # type: []string
  # env var: LOTUS_LIBP2P_LISTENADDRESSES
  ListenAddresses = ["/ip4/0.0.0.0/tcp/50000", "/ip6/::/tcp/0"]

  # Addresses to explicitally announce to other peers. If not specified,
  # all interface addresses are announced
  # Format: multiaddress
  #
  # type: []string
  # env var: LOTUS_LIBP2P_ANNOUNCEADDRESSES
  #AnnounceAddresses = []
```

TODO:
- [x] Remove unneeded config values from config def (eg we only need `MaxPublishDealsFee` in `MinerFeeConfig`)
- [x] Integration tests


dirkmc:
> Can we open an issue to update backup command with new config directory?

Yes please go ahead and do that

> Is it safe to delete the /config.toml

This PR replaces the config.toml with a symlink.
So the directory structure will be something like:
```
<boost repo>/config/config.toml.0
<boost repo>/config/config.toml.1
<boost repo>/config.toml -> <boost repo>/config/config.toml.1
```

The reason to keep the old config.toml.x files in the config directory is so that the user can migrate down.
When the user changes the version of boost that they are running it will automatically migrate the config file up or down accordingly.

node/config/migrate.go
```diff
@@ -0,0 +1,177 @@
...
+
+	return string(bz), nil
+}
+
+// This struct is used to get the config file version
+type VersionCfg struct {
+	ConfigVersion int
+}
+
+func ConfigMigrate(repoPath string) error {
```
nonsense, line 53: Nitpick: we are in package `config`, we could call this `Migrate`, so that we don't end up with config.ConfigMigrate(...), but just config.Migrate(...)

nonsense: Same for `ConfigUpdate`.

dirkmc: I thought about doing the same thing - but ConfigUpdate etc are copied over from the lotus repo, so I thought it would make it easier to port over any future bug fixes there if we keep the names the same

PR #607, opened 2022-06-20
replace npm install with npm ci

nonsense:
Since not all of our node dependencies are locked in `package.json`, `npm install` upgrades them whenever possible. (more background at: https://stackoverflow.com/questions/45022048/why-does-npm-install-rewrite-package-lock-json)

I think it is preferable that we have reproducible builds and build packages as they are defined in `package-lock.json`, so I suggest we change to `npm ci` and occasionally on-demand upgrade our dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,11 @@ boostci: $(BUILD_DEPS)
 .PHONY: boostci
 
 react: check-node-lts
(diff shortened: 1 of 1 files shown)


dirkmc (approved):
Thanks!

PR #603, opened 2022-06-19
feat: support json output in the `boost wallet` command

nicobao:
Relates to #469 

List of commands (`boost --json wallet <command>`) to implement json support for:
- [x] `new`
- [x] `list`
- [x] `balance`
- [x] `export `
- [x] `import`
- [x] `default`/`get-default`
- [x] `set-default` (no output on the first place)
- [x] `delete` (no output on the first place)

PR #608, opened 2022-06-22
feat(cli): render error outputs as JSON when the option is set

nicobao:
Closes #603 

This PR aims to wrap all errors in a JSON file when the `--json` flag is set:
```json
{
    "error": "error message"
}
```
The feature was discussed here: https://github.com/filecoin-project/boost/pull/603#issuecomment-1160546599

### DIFF
--- a/cmd/boost/provider_cmd.go
+++ b/cmd/boost/provider_cmd.go
@@ -42,53 +42,62 @@ var libp2pInfoCmd = &cli.Command{
 	Before:      before,
 	Action: func(cctx *cli.Context) error {
 		ctx := ctxutil.ReqContext(cctx)
+		outputInJson := cctx.Bool("json")
 
 		if cctx.Args().Len() != 1 {
-			return fmt.Errorf("usage: protocols <provider address>")
+			err := fmt.Errorf("usage: protocols <provider address>")
+			return cmd.PrintError(err, outputInJson)
 		}
 
 		n, err := clinode.Setup(cctx.String(cmd.FlagRepo.Name))
 		if err != nil {
-			return fmt.Errorf("setting up CLI node: %w", err)
+			augmentedError := fmt.Errorf("setting up CLI node: %w", err)
+			return cmd.PrintError(augmentedError, outputInJson)
 		}
 
 		api, closer, err := lcli.GetGatewayAPI(cctx)
 		if err != nil {
-			return fmt.Errorf("setting up gateway connection: %w", err)
+			augmentedError := fmt.Errorf("setting up gateway connection: %w", err)
+			return cmd.PrintError(augmentedError, outputInJson)
 		}
 		defer closer()
 
 		addrStr := cctx.Args().Get(0)
 		maddr, err := address.NewFromString(addrStr)
 		if err != nil {
-			return fmt.Errorf("parsing provider on-chain address %s: %w", addrStr, err)
+			augmentedError := fmt.Errorf("parsing provider on-chain address %s: %w", addrStr, err)
+			return cmd.PrintError(augmentedError, outputInJson)
 		}
 
 		addrInfo, err := cmd.GetAddrInfo(ctx, api, maddr)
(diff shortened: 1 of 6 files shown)


dirkmc:
I think there should be an easier way. Does this code not work?
https://github.com/filecoin-project/boost/blob/5c52ea15d1b5d00b0c96bf4bfc4c1979d2045c1b/cmd/boost/main.go#L47-L56

cmd/boost/provider_cmd.go
```diff
@@ -267,7 +286,8 @@ var retrievalAskCmd = &cli.Command{
 		var ask retrievalmarket.QueryResponse
 
 		if err := doRpc(ctx, s, &req, &ask); err != nil {
-			return fmt.Errorf("send retrieval-ask request rpc: %w", err)
+			augmentedError := fmt.Errorf("send retrieval-ask request rpc: %w", err)
+			return cmd.PrintError(augmentedError, outputInJson)
 		}
 
 		afmt.Printf("Status: %d\n", ask.Status)
```
nicobao, line 293: Why do we use the `afmt` special command here and not everywhere else?
Why don't we support `json` output for this command as well?

cmd/boost/provider_cmd.go
```diff
@@ -174,7 +187,8 @@ var storageAskCmd = &cli.Command{
 		}
 
 		if err := doRpc(ctx, s, &askRequest, &resp); err != nil {
-			return fmt.Errorf("send ask request rpc: %w", err)
+			augmentedError := fmt.Errorf("send ask request rpc: %w", err)
+			return cmd.PrintError(augmentedError, outputInJson)
 		}
 
 		ask := resp.Ask.Ask
```
nicobao, line 194: Why do we use the `afmt` special command here and not everywhere else?
Why don't we support `json` output for this command as well?

nicobao:
> I think there should be an easier way. Does this code not work?
> 
> https://github.com/filecoin-project/boost/blob/5c52ea15d1b5d00b0c96bf4bfc4c1979d2045c1b/cmd/boost/main.go#L47-L56

How should I use this code?
Currrently, on running any command, it doesn't seem to detect the `--json` flag and convert the errors to json.

How would you call it within the commands?

If I put it in `cmd/util.go`, it won't work because I don't have access to the `app` 

I can put it in `cmd/boost/helper.go` but then I'd need to instanciate any printing with `NewAppFmt` which we aren't doing right now. We use `fmt` directly most of the time.
Is that what you want me to use?
```
func (a *AppFmt) PrintError(err error) {
	if isJson, ok := a.app.Metadata["json"].(bool); isJson && ok {
		resJson, err := json.Marshal(map[string]string{"error": err.Error()})
		if err != nil {
			fmt.Fprintln(os.Stderr, fmt.Errorf("marshalling json: %w", err))
		} else {
			fmt.Fprintln(os.Stderr, string(resJson))
		}
	} else {
		fmt.Fprintln(os.Stderr, "Error: "+err.Error())
	}
}
```

nicobao:
It would definitely be better for sure. It's not working right now, the output isn't wrapped to JSON. I take a look!

parkan:
:+1: for JSON everywhere, this will actually help me handle the different error cases more cleanly in our pipelines as well :)

dirkmc:
When I run a command with the `--json` flag and there is an error it outputs the error in json format:
```
boost --json deal --verified=false --provider=t01234 --commp=baga6ea4seaqgbkvk6apsnfbtvnbjabpu4tsv6ns45c36452fdhkqxznt3vlnedq --car-size=1101 --piece-size=2048 --payload-cid=bafk2bzacedsnewk7clxwab2wgwyoi7u5tzdhldx7fkxpqdq7unrxz2zoy4d2g --http-url="http://localhost:5000/1k-file.car"
{"error":"failed to load miner actor: actor not found"}
```

So I'm going to close this PR as it seems this functionality is already working

PR #625, opened 2022-06-28
update backup command to include new config directory

LexLuthr:


cmd/boostd/backup.go
```diff
@@ -291,6 +319,44 @@ var restoreCmd = &cli.Command{
...
+			}
+		}
+
+		//Remove default config.toml created with repo. Don't fail if file cannot be removed
+		if err = os.Remove(path.Join(lr.Path(), "config.toml")); err != nil {
+			fmt.Println(fmt.Errorf("failed to read the restored config files: %w", err))
+			fmt.Println("Please create the config.toml link manually")
+		}
+
+		latestCfgFile := "config.toml." + strconv.Itoa(cfgNum)
```
LexLuthr, line 348: @dirkmc I don't like this code. It is too childish. If you have any better suggestion, please let me know.

dirkmc (commented):
Thanks for putting this PR together 🙏 
Looks good so far - just one suggestion

cmd/boostd/backup.go
```diff
@@ -291,6 +319,44 @@ var restoreCmd = &cli.Command{
 
 		}
 
+		configFiles, err := ioutil.ReadDir(path.Join(lr.Path(), "config"))
```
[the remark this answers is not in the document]
LexLuthr: I tried with both os.Stat() and os.Lstat() but in both scenarios backup directory ends up with a file names config.toml and not a symlink. 
My first approach was to copy and symlink and restore it back but when it kept creating files, I switched this weird code to handle the symlink part.

cmd/boostd/backup.go
```diff
@@ -363,26 +306,47 @@ var restoreCmd = &cli.Command{
...
-		fmt.Printf("Not copying %s as file does not exists\n", src)
-		return nil
-	}
+// This function uses chdir() to deal with relative symlinks be careful when reusing
+func copyFiles(srcDir, destDir string, flist []string) error {
 
-	if err != nil && !os.IsNotExist(err) {
+	if err := os.Chdir(srcDir); err != nil {
 		return err
 	}
```
dirkmc, line 314: I believe it should be possible to write a relative symlink without changing the current working directory.
See https://stackoverflow.com/questions/35541589/creating-a-relative-symbolic-link-through-the-os-package

LexLuthr: It wasn't working for me for some reason. Probably due to some mistake in the code. I have removed the chdir() and tested the binary. It seems to be working properly. I am keeping the new copyFiles() structure as this is more portable for reuse.

cmd/boostd/backup.go
```diff
@@ -297,26 +306,42 @@ var restoreCmd = &cli.Command{
...
+
+			input, err := ioutil.ReadFile(path.Join(srcDir, fName))
+			if err != nil {
+				return err
+			}
+
+			err = ioutil.WriteFile(path.Join(destDir, fName), input, f.Mode())
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
LexLuthr: I will rewrite with buffers after multiple boost node is merged as that will require backup command changes again.

PR #632, opened 2022-07-06
Added env vars for the flags for testing convinience

dd45e640b42e6da7da96faee3996ef7c:
the added flags are all for values that can reasonably be "static" between testing runs of storing files


nonsense (approved):

nonsense:
On a second thought, after having discussed this with the team, we are not sure it makes sense adding this.

We'd rather not have to maintain so many ENV variables for Boost.

If you want to persist these particular values (and not others) for testing you could create an alias, and then specify the rest of the arguments with it:

```
alias bd='boost deal --provider=t01000 ...'
```

bd --commp=xxx ...

PR #640, opened 2022-07-07
http retrieve entire piece (not just CAR file)

dirkmc:


cmd/booster-http/server.go
```diff
@@ -189,6 +198,9 @@ func (s *HttpServer) handleByPieceCid(w http.ResponseWriter, r *http.Request) {
 
 	ctx := r.Context()
 	content, err := s.getPieceContent(ctx, pieceCid)
+	if err == nil && isCar {
+		content, err = s.getCarContent(pieceCid, content)
+	}
```
nonsense, line 203: Aren't `getPieceContent` and `getCarContent` expensive operations? Shouldn't we do one or the other, but not override `content` when `isCar == true` ?

dirkmc: In this case `content` is an `io.ReadSeeker`
`getPieceContent` gets a reader over the unsealed piece.
if `isCar == true` then `getCarContent` figures out the size of the CAR file by asking the DAG store for the max offset of the CAR index. If the DAG store doesn't have the index, `getCarContent` creates the index on the fly. Once it has the max offset, it just wraps the piece reader with a limit reader so that it only serves the CAR file.

mikeal:
🤩

Issue #652, opened 2022-07-20
Deals randomly fail with 'failed to persist deal state: database is locked' error after publish


cerblue:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature or enhancement request. If it is, please open a [new idea discussion](https://github.com/filecoin-project/boost/discussions/new?category=ideas) instead. New feature and enhancement requests would be entertained by the boost team after a thorough discussion only.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to boost.

### Boost component

- [X] boost daemon - storage providers
- [ ] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### Boost Version

```text
Daemon: boostd version 1.2.0+git.c355880
```

### Describe the Bug

Deals randomly fail with error: `failed to persist deal state: database is locked` after deal publish step. The deal will go into error state and cannot be retried. This happens intermittently; there are cases where all deals in a batch succeed, and cases where 3 out 5 deals fail.

My Boost daemon is accepting ~20 deals a day, in few bursts of 5~10 deals. Boost will batch up to 10 deals before publishing the PublishStorageDeals message, but usually flushes the batch with 5~8 deals.

### Logging Information

```text
Jul 20 02:54:08 purumine-miner boostd[1271904]: {"level":"info","ts":"2022-07-20T02:54:08.131Z","logger":"boost-storage-deal","caller":"logs/log.go:40","msg":"sending deal to deal publisher","id":"a163925b-08df-4c8c-a06d-0d7d070bdbfc"}
Jul 20 03:18:06 purumine-miner boostd[1271904]: {"level":"error","ts":"2022-07-20T03:18:06.726Z","logger":"boost-storage-deal","caller":"logs/log.go:53","msg":"deal failed","id":"a163925b-08df-4c8c-a06d-0d7d070bdbfc","err":"failed to persist deal state: database is locked"}
[...]

PR #657, opened 2022-07-20
Fix 'database is locked' error

cerblue:
Fixes https://github.com/filecoin-project/boost/issues/652

Caused by concurrent access from deal goroutines to a single sqlite3 DB connection. By default, this will cause the 'database is locked' error, and it is the reason for failures in deal status updates and deal log appends. The fix comes from suggestions in https://github.com/mattn/go-sqlite3#:~:text=Error%3A%20database%20is%20locked and https://github.com/mattn/go-sqlite3/issues/209.

I wrote a toy program to reproduce the error and confirmed that this change fixes the error, and does the same to my boost daemon instance.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -19,7 +19,12 @@ type Scannable interface {
 }
 
 func SqlDB(dbPath string) (*sql.DB, error) {
-	return sql.Open("sqlite3", "file:"+dbPath)
+	db, err := sql.Open("sqlite3", "file:"+dbPath)
+	if err == nil {
+		db.SetMaxOpenConns(1)
+	}
+
+	return db, err
 }
 
 //go:embed create_main_db.sql


nonsense:
Thank you for this fix @cerblue .

I think we should also add the `?cache=shared` as per docs at https://github.com/mattn/go-sqlite3#faq , what do you think?

---

I think we need to add the `?cache=shared` at two places:

```
diff --git a/node/modules/storageminer.go b/node/modules/storageminer.go
index 0747c80..de7256c 100644
--- a/node/modules/storageminer.go
+++ b/node/modules/storageminer.go
@@ -285,7 +285,7 @@ func StorageNetworkName(ctx helpers.MetricsCtx, a v1api.FullNode) (dtypes.Networ
 }

 func NewBoostDB(r lotus_repo.LockedRepo) (*sql.DB, error) {
-       dbPath := path.Join(r.Path(), "boost.db")
+       dbPath := path.Join(r.Path(), "boost.db?cache=shared")
        return db.SqlDB(dbPath)
 }

@@ -294,7 +294,7 @@ type LogSqlDB struct {
 }

 func NewLogsSqlDB(r repo.LockedRepo) (*LogSqlDB, error) {
-       dbPath := path.Join(r.Path(), "boost.logs.db")
+       dbPath := path.Join(r.Path(), "boost.logs.db?cache=shared")
        d, err := db.SqlDB(dbPath)
        if err != nil {
                return nil, err
```

jacobheun:
Closing this in favor of https://github.com/filecoin-project/boost/pull/679, which includes this.

Issue #667, opened 2022-07-22
Config Option To Record Support For HTTP


hannahhoward:
# Goals

Providers need to have a way to indicate they are running booster-http for the QueryAsk v2 protocol

# How

1. Add a "HTTPRetrievalURL" string to https://github.com/filecoin-project/boost/blob/fe1b0139be4c769a1348abd9bb0a6e4e55973435/node/config/types.go#L123

2. When the URL is not a blank string, QueryAskV2 will return HTTP options. When it's blank, from QueryAsk v2's point of view, HTTP retrieval is disabled.

This config is the PUBLIC url for HTTP retrieval. Currently, booster HTTP only allows configuring the port, and doesn't support HTTPS. The assumption is if you want your server to run over HTTPS, on a different port externally, or with a filter, it's configured through an Nginx/Apache reverse proxy. 

# For Discussion

The current proposal is this URL is set manually in the config file, independently of booster-http being started/stopped. That means that there's no guarantee it's correct, or that while set, booster-http is actually running. Should we try to do something to disable returning the URL when booster-http is stopped?

jacobheun:
> Should we try to do something to disable returning the URL when booster-http is stopped?

I think we could do this later. It should be fairly trivial to have booster-http tell boost to enable/disable on startup/shutdown via the boost api. This might get annoying to track if we go beyond 1:1 scaling. If we ended up doing public endpoint configuration through a boost-http flag on startup, then we'll probably want to enable this.

PR #671, opened 2022-07-29
Retrieval query ask v2

### LINKED ISSUE #666: QueryAsk v2 Protocol
# Goals

Enable discoverability of HTTP retrieval support

The primary changes for v2 protocol are:

Query:
- You can ask for just a PieceCID (only piece retrieval protocols will be returned)

Response:
- Information will be given about available protocols, and parameters specific to those protocols

For now the newly design protocol makes all additional information protocol specific

# How

- probably make a new retrievalmarket folder to collect newly built retrieval pieces
- make an implementation of the [query protocol for libp2p](../tree/main/storagemarket/lp2pimpl), also drawing from [retrieval query protocol v1](https://github.com/filecoin-project/go-fil-markets/tree/master/retrievalmarket/network)

The schema for the QueryAsk v1 request / response (in IPLD schema -- it's encoded as DAG CBOR) is as follows:

```ipldsch
# QueryParams - V1 - indicate what specific information about a piece that a retrieval
# client is interested in, as well as specific parameters the client is seeking
# for the retrieval deal
type QueryParams struct {
	PieceCID nullable Link  # optional, query if miner has this cid in this piece. some miners may not be able to respond.
}

# Query is a query to a given provider to determine information about a piece
# they may have available for retrieval
type Query struct {
	PayloadCID  Link 
	QueryParams QueryParams
}

# QueryResponseStatus indicates whether a queried piece is available
type QueryResponseStatus enum {
  # QueryResponseAvailable indicates a provider has a piece and is prepared to
	# return it
	| QueryResponseAvailable ("0")

	# QueryResponseUnavailable indicates a provider either does not have or cannot
	# serve the queried piece to the client
	| QueryResponseUnavailable ("1")

	# QueryResponseError indicates something went wrong generating a query response
	| QueryResponseError ("2")
} representation int

# QueryItemStatus (V1) indicates whether the requested part of a piece (payload or selector)
# is available for retrieval
type QueryItemStatus enum {
	# QueryItemAvailable indicates requested part of the piece is available to be
	# served
	| QueryItemAvailable ("0")

	# QueryItemUnavailable indicates the piece either does not contain the requested
	# item or it cannot be served
	| QueryItemUnavailable ("1")

	# QueryItemUnknown indicates the provider cannot determine if the given item
	# is part of the requested piece (for example, if the piece is sealed and the
	# miner does not maintain a payload CID index)
	| QueryItemUnknown ("2")
} representation int

type Address Bytes
type TokenAmount Bytes

# QueryResponse is a miners response to a given retrieval query
type QueryResponse struct {
	Status        QueryResponseStatus
	PieceCIDFound QueryItemStatus # V1 - if a PieceCID was requested, the result

	Size Int

	PaymentAddress             Address # address.Address to send funds to -- may be different than miner addr
	MinPricePerByte            TokenAmount
	MaxPaymentInterval         Int
	MaxPaymentIntervalIncrease Int
	Message                    String
	UnsealPrice                TokenAmount
}
```

The proposed v2 schema is:

```ipldsch

# QueryKind specifies whether you are interested in a payload or a piece
type QueryKind enum {
    | Piece ("pc")
    | Payload ("pl")
} representation string

# Query is a query to a given provider to determine information about a piece
# they may have available for retrieval
type Query struct {
        Kind  QueryKind
        # If kind is Payload, response will error if not included
	PayloadCID optional Link
	# If kind is Piece, response will error if not included
	PieceCID optional Link 
}

# QueryResponseStatus indicates whether a queried piece is available
type QueryResponseStatus enum {
  # QueryResponseAvailable indicates a provider has a piece and is prepared to
	# return it
	| QueryResponseAvailable ("0")

	# QueryResponseUnavailable indicates a provider either does not have or cannot
	# serve the queried piece to the client
	| QueryResponseUnavailable ("1")

	# QueryResponseError indicates something went wrong generating a query response
	| QueryResponseError ("2")
} representation int

# QueryItemStatus (V1) indicates whether the requested part of a piece (payload or selector)
# is available for retrieval
type QueryItemStatus enum {
	# QueryItemAvailable indicates requested part of the piece is available to be
	# served
	| QueryItemAvailable ("0")

	# QueryItemUnavailable indicates the piece either does not contain the requested
	# item or it cannot be served
	| QueryItemUnavailable ("1")

	# QueryItemUnknown indicates the provider cannot determine if the given item
	# is part of the requested piece (for example, if the piece is sealed and the
	# miner does not maintain a payload CID index)
	| QueryItemUnknown ("2")
} representation int

type Address Bytes
type TokenAmount Bytes

type ProtocolName enum {
   | GraphsyncFIlecoinV1 ("graphsync/v1")
   | HTTPFilecoinV1 ("http/v1")
} representation string

# QueryResponse is a miners response to a given retrieval query
type QueryResponse struct {
	Status        QueryResponseStatus
	Protocols {ProtocolName:Any}
}

# Structure of Graphsync Protocol Data
type GraphsyncFilecoinV1Response struct {
	Size Int
	PaymentAddress             Address # address.Address to send funds to -- may be different than miner addr
	MinPricePerByte            TokenAmount
	MaxPaymentInterval         Int
	MaxPaymentIntervalIncrease Int
	Message                    String
	UnsealPrice                TokenAmount
}

# Structure of HTTP V1 Response
type HTTPFilecoinV1Response {
      URL String
      Size Int
}
```

hannahhoward:
# Goals

fix #666 #667  #668 

# Implementation

1. Generate query ask v2 types & ipld schema based on the one documented in #666 
2. Create a libp2p implementation to send and request v2 Query's & QueryResponses -- retrievalmarket/lp2pimpl, based on storagemarket/lp2pimpl
3. Build a retrieval provider which implements query handling. Most of the code here is copied from the go-fil-markets retrieval provider query handling, with additional logic to add http info when neccesary, and to handle piece only queries
4. I also removed the RetrievalProviderNode abstraction since the storagemarket side in boost removes this abstraction
5. Setup a test of the retrieval query system, based on tests in go-fil-markets retrievalmarket impl, plus additional tests for http logic
6. Add config for HTTPRetrievalURL to DealMaking config
7. Add builder logic to construct the retrieval provider for boost
8. Add a cli command to send a query and output a response, specifically for piece only retrievals for now (tailored for Evergreen usage)

# For Discussion

- to do: complete an integration test, manual test the retrieval command

### DIFF
--- a/cmd/boost/retrieve_piece_cmd.go
+++ b/cmd/boost/retrieve_piece_cmd.go
@@ -1,0 +1,82 @@
+package main
+
+import (
+	"fmt"
+
+	bcli "github.com/filecoin-project/boost/cli"
+	clinode "github.com/filecoin-project/boost/cli/node"
+	"github.com/filecoin-project/boost/cmd"
+	"github.com/filecoin-project/boost/retrievalmarket/lp2pimpl"
+	"github.com/filecoin-project/boost/retrievalmarket/types"
+	"github.com/filecoin-project/go-address"
+	lcli "github.com/filecoin-project/lotus/cli"
+	"github.com/ipfs/go-cid"
+	"github.com/urfave/cli/v2"
+)
+
+var retrievePieceCmd = &cli.Command{
+	Name:  "retrieve-piece",
+	Usage: "",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:     "provider",
+			Usage:    "storage provider on-chain address",
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:     "piece-cid",
+			Usage:    "",
+			Required: true,
+		},
+	},
+	Before: before,
+	Action: func(cctx *cli.Context) error {
+		ctx := bcli.ReqContext(cctx)
+
+		pieceCID, err := cid.Parse(cctx.String("piece-cid"))
+		if err != nil {
+			return err
+		}
+
+		n, err := clinode.Setup(cctx.String(cmd.FlagRepo.Name))
+		if err != nil {
+			return err
+		}
+
+		api, closer, err := lcli.GetGatewayAPI(cctx)
+		if err != nil {
+			return fmt.Errorf("cant setup gateway connection: %w", err)
+		}
+		defer closer()
+
+		maddr, err := address.NewFromString(cctx.String("provider"))
+		if err != nil {
+			return err
+		}
+
+		addrInfo, err := cmd.GetAddrInfo(ctx, api, maddr)
+		if err != nil {
+			return err
+		}
+
+		log.Debugw("found storage provider", "id", addrInfo.ID, "multiaddrs", addrInfo.Addrs, "addr", maddr)
+
+		if err := n.Host.Connect(ctx, *addrInfo); err != nil {
+			return fmt.Errorf("failed to connect to peer %s: %w", addrInfo.ID, err)
+		}
+
+		dc := lp2pimpl.NewQueryClient(n.Host)
+		resp, err := dc.SendQuery(ctx, addrInfo.ID, types.Query{
+			PieceCID: &pieceCID,
+		})
+		if err != nil {
+			return fmt.Errorf("send deal status request failed: %w", err)
+		}
+
+		if resp.Error != "" {
+			return fmt.Errorf("query response error: %s", resp.Error)
+		}
+		fmt.Println(resp.Protocols.HTTPFilecoinV1.URL)
+		return nil
+	},
+}
--- a/retrievalmarket/lp2pimpl/net.go
+++ b/retrievalmarket/lp2pimpl/net.go
@@ -1,0 +1,138 @@
+package lp2pimpl
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/filecoin-project/boost/retrievalmarket"
+	"github.com/filecoin-project/boost/retrievalmarket/types"
+	"github.com/filecoin-project/go-fil-markets/shared"
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/ipld/go-ipld-prime/codec/dagcbor"
+	host "github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/network"
+	peer "github.com/libp2p/go-libp2p-core/peer"
+	protocol "github.com/libp2p/go-libp2p-core/protocol"
+)
+
+var log = logging.Logger("boost-net")
+
+// QueryV2ProtocolID is the protocol for querying information about retrieval
+// deal parameters
+const QueryV2ProtocolID = protocol.ID("/fil/retrieval/qry/2.0.0")
+
+// QueryProvider listens for incoming deal proposals over libp2p
+type QueryProvider struct {
+	ctx  context.Context
+	host host.Host
+	prov *retrievalmarket.Provider
+}
+
+const providerReadDeadline = 10 * time.Second
+const providerWriteDeadline = 10 * time.Second
+const clientReadDeadline = 10 * time.Second
+const clientWriteDeadline = 10 * time.Second
+
+// QueryClientOption is an option for configuring the libp2p storage deal client
+type QueryClientOption func(*QueryClient)
+
+// RetryParameters changes the default parameters around connection reopening
+func RetryParameters(minDuration time.Duration, maxDuration time.Duration, attempts float64, backoffFactor float64) QueryClientOption {
+	return func(c *QueryClient) {
+		c.retryStream.SetOptions(shared.RetryParameters(minDuration, maxDuration, attempts, backoffFactor))
+	}
+}
+
+// QueryClient sends retrieval queries over libp2p
+type QueryClient struct {
+	retryStream *shared.RetryStream
+}
+
+// SendQuery sends a retrieval query over a libp2p stream to the peer
+func (c *QueryClient) SendQuery(ctx context.Context, id peer.ID, query types.Query) (*types.QueryResponse, error) {
+	log.Debugw("send query", "pieceCID", query.PieceCID, "payloadCID", query.PayloadCID, "provider-peer", id)
+
+	// Create a libp2p stream to the provider
+	s, err := c.retryStream.OpenStream(ctx, id, []protocol.ID{QueryV2ProtocolID})
+	if err != nil {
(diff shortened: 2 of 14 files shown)


retrievalmarket/types/queryaskv2.go
```diff
@@ -0,0 +1,147 @@
...
+
+func init() {
+	for _, r := range []struct {
+		typ     interface{}
+		typName string
+	}{
+		{(*QueryResponse)(nil), "QueryResponse"},
+		{(*QueryProtocols)(nil), "QueryProtocols"},
+		{(*GraphsyncFilecoinV1Response)(nil), "GraphsyncFilecoinV1Response"},
+		{(*HTTPFilecoinV1Response)(nil), "HTTPFilecoinV1Response"},
```
rvagg, line 140: I think these aren't used directly via the registry but they are wired up through `QueryResponse` so they shouldn't need to be registered here.

retrievalmarket/types/queryaskv2.ipldsch
```diff
@@ -0,0 +1,54 @@
+# Query is a query to a given provider to determine information about a piece
+# they may have available for retrieval
+type Query struct {
+  # If kind is Payload, response will error if not included
```
rvagg, line 4: there's some space vs tabs inconsistencies in this file

cmd/boost/retrieve_piece_cmd.go
```diff
@@ -0,0 +1,82 @@
...
+			PieceCID: &pieceCID,
+		})
+		if err != nil {
+			return fmt.Errorf("send deal status request failed: %w", err)
+		}
+
+		if resp.Error != "" {
+			return fmt.Errorf("query response error: %s", resp.Error)
+		}
+		fmt.Println(resp.Protocols.HTTPFilecoinV1.URL)
```
rvagg, line 79: `resp.Protocols.HTTPFilecoinV1` could be `nil` couldn't it? Maybe something like this:

```suggestion
		if resp.Protocols.HTTPFilecoinV1 != nil {
			fmt.Println(resp.Protocols.HTTPFilecoinV1.URL)
		} else {
			return fmt.Errorf("peer does not support HTTP retrievals")
		}
```

retrievalmarket/lp2pimpl/net.go
```diff
@@ -0,0 +1,148 @@
...
+
+	defer s.Close() // nolint
+
+	// Set a deadline on writing to the stream so it doesn't hang
+	_ = s.SetWriteDeadline(time.Now().Add(clientWriteDeadline))
+	defer s.SetWriteDeadline(time.Time{}) // nolint
+
+	// Write the retrieval query to the stream
+	// Write the re to the client
+	err = types.BindnodeRegistry.TypeToWriter(query, s, dagcbor.Encode)
```
rvagg, line 80: maybe at this point it should interrogate `query` to see if it's valid, i.e. has a `PieceCID` or a `PayloadCID`, before proceeding to send it since it'll be rejected anyway

retrievalmarket/types/queryaskv2.ipldsch
```diff
@@ -0,0 +1,54 @@
...
+	# return it
+	| QueryResponseAvailable ("0")
+
+	# QueryResponseUnavailable indicates a provider either does not have or cannot
+	# serve the queried piece to the client
+	| QueryResponseUnavailable ("1")
+
+	# QueryResponseError indicates something went wrong generating a query response
+	| QueryResponseError ("2")
+} representation int
```
rvagg, line 22: `string` enums are a lot nicer to see when looking at message contents and they shouldn't cost much more if you choose short names; I think we have a couple of places of regret in the current filecoin stack where int enums exist and we could have gone with strings (IIRC there was a discussion in graphsync when we did the initial bindnode convert). Up to you, not a big deal but the message format here is already fairly verbose and would be pretty nice to look at if you're just decoding the raw messages because of all the maps, but then you get to these numbers and they're not self explanatory.

dirkmc: Agree, I'd favour strings 👍

cmd/boost/retrieve_piece_cmd.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/filecoin-project/boost/retrievalmarket/lp2pimpl"
+	"github.com/filecoin-project/boost/retrievalmarket/types"
+	"github.com/filecoin-project/go-address"
+	lcli "github.com/filecoin-project/lotus/cli"
+	"github.com/ipfs/go-cid"
+	"github.com/urfave/cli/v2"
+)
+
+var retrievePieceCmd = &cli.Command{
+	Name:  "retrieve-piece",
```
dirkmc, line 18: I'd suggest adding a `retrieve` category, and renaming this action to `query-ask`, so the CLI command would be
`boost retrieve query-ask --provider=t01234 --piece-cid=1234`

I also don't think we really need to provide the piece cid. The path to get things can be "well-known", eg `/piece/<cid>.car`

cmd/boost/retrieve_piece_cmd.go
dirkmc, line 79: I'd suggest outputting all protocols and let the user parse for the one they want to use.
Let's also implement output json if the --json flag is set, so that it's easy for people to write parsers (see `boost deal` for an example)

dirkmc:
### Current Design

1. Client sends query over libp2p with piece cid / payload cid
     Server queries sealing subsystem to check if it's possible to retrieve the piece / block
     Server responds with either
     - url to get data http://1.2.3.4:1234/piece/abc123.car
     - error message (eg no unsealed piece found)
2. Client makes an HTTP request to get data
    Server queries sealing subsystem to check if it's possible to retrieve the piece / block
    Server responds with either
    - data
    - error message (eg no unsealed piece found)

### Alternative Design
1. Client sends query over libp2p to ask what protocols server supports
     Server responds with supported protocols (graphsync, http at http://1.2.3.4:1234/)
2. Client makes an HTTP request to get data
    Server queries sealing subsystem to check if it's possible to retrieve the piece / block
    Server responds with either
    - data
    - error message (eg no unsealed piece found)
   
The reason I'm proposing the alternative design is because
- It simplifies the libp2p part (no need to query sealing subsystem), just returns an IP & port (not the entire URL with the path to the specific car file)
- Once the client knows the address of the miner's HTTP interface the client doesn't need to send 2 requests every time it asks for data (one request to libp2p and a second request to http). It can just directly query the http interface.

PR #672, opened 2022-08-01
Upgrade lotus dependency to v1.17.0-rc3 [Do not merge till Anton/Dirk's review]

LexLuthr:
Lotus v1.17.0-rc3 has upgraded their API version and thus older version of boost is no longer working with new lotus RCs.


LexLuthr:
MacOS build has been fixed in the main so it should not be failing test going forward.

nonsense (approved):

Issue #673, opened 2022-08-02
Estuary <> Boost CommP mismatch investigation


rvagg:
Some notes on the investigation so far, I know this is a lot of blather and you can skim down through to the last half to see the juiciest bits. Not completely solved but I think I'm close and I believe the actions I've listed at the bottom will get us to resolution.

Previous threads:

* Filecoin Slack `#ecosystem-dev` channel has threads ([here](https://filecoinproject.slack.com/archives/C016APFREQK/p1659106722165669) & [here](https://filecoinproject.slack.com/archives/C016APFREQK/p1659136013530079))
* https://github.com/application-research/filclient/pull/91
* https://github.com/ipld/go-car/pull/290#issuecomment-1200652674

Thankfully Alvin (via stuberman and lodge) was able to provide a DAG that is failing so we can dig deeper into the nature of the failure. A ~32G DAG hanging off `bafybeietzjzc4vudlevv6k6sxdixhp5nnmfblyjhqheyjyd4d3uluvqdgm`. Looking at the version that `ipfs dag export` (proper exhaustive selector export, what we would expect for a well-formed CAR) gives compared to the one that Boost apparently has on its end where it's reporting the mismatch we can see:

* They are the same size
* They contain the same blocks
* The blocks are out of order

The ordering problem can be seen just by looking at the first few blocks. Here's what we expect (`ipfs dag export`):

```
bafybeietzjzc4vudlevv6k6sxdixhp5nnmfblyjhqheyjyd4d3uluvqdgm
bafkreigbnoobwwzjdl4yoccgyyeyybltkqqq5v7uix45zwi5rjvs64xwfy
bafkreih5rbh2rwzbg2v66rxpyjr755ycub246asuxa55k7ey3xlk457xiq
bafkreig6vmc5bg6lyn65k6jbteqqpugkhzxglbryoiyfiwwn5qef4hyiuu
bafkreibd3ult3qdw4xty3bj4n2kmryn3m6x7kshkjnb7dacbuxi5snj6ki
bafkreid73ihqui4bjpsse7ytsakpxg5r5yfhafsr35dyq67u2mojtvn6nq
bafkreifz4ya5gbziqtyxi4pua3tplk6d2awklicb74pdkmre65vioigvkq
bafkreidq4n2g6bqgct2s35naypliplkhfm3ljt2lxckfi5qxisr4wc6sta
[...]

PR #675, opened 2022-08-03
fix: don't allow go-merkledag to reorder loaded links

rvagg:
Ref: https://github.com/filecoin-project/boost/issues/673#issuecomment-1202312003

Ideally we can remove go-merkledag entirely from here because most (all?) of the deal-making infrastructure uses go-codec-dagpb + go-ipld-prime traversals and post-decode link-reordering is a subtle difference that impacts CAR ordering for non-spec-compliant DAG-PB blocks.

But for now, this is a quick-fix that should preserve the links in the order as they are appear in the block, which is what our other traversals are working with.

### DIFF
--- a/car/car_offset_writer.go
+++ b/car/car_offset_writer.go
@@ -106,8 +106,13 @@ func (s *CarOffsetWriter) writeBlocks(ctx context.Context, w io.Writer, headerSi
 			return nil, fmt.Errorf("getting block %s: %w", c, err)
 		}
 
+		// take a copy of the links array before nd.RawData() triggers a sort
+		links := make([]*format.Link, len(nd.Links()))
+		copy(links, nd.Links())
+		byts := nd.RawData()
+
 		// Get the size of the block and metadata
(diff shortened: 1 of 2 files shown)


rvagg:
Added test case for this which fails on `main` but passes here, it should demonstrate what we're trying to fix.

masih (approved):

car/car_offset_writer.go
```diff
@@ -106,8 +106,13 @@ func (s *CarOffsetWriter) writeBlocks(ctx context.Context, w io.Writer, headerSi
 			return nil, fmt.Errorf("getting block %s: %w", c, err)
 		}
 
+		// take a copy of the links array before nd.RawData() triggers a sort
```
willscott, line 109: the real solution is to get https://github.com/ipld/go-car/pull/291 to replace this class, right?

willscott (approved):

car/car_offset_writer.go
rvagg: yes, that's the proper thing that needs to be done

PR #680, opened 2022-08-03
(Do not merge) v1.0.0/filclient-commp-fix branch + patch

rvagg:
This has the two commits from https://github.com/filecoin-project/boost/pull/675 but is branched off `v1.0.0`, the intention with this is to get https://github.com/application-research/filclient updated with v1.0.0 as it's currently relying on cd38741e464d but I think pulling it to v1.0.0 won't be too traumatic. We'll see, we might have to pull the branch back a bit to not make it too big a jump.


jacobheun:
This has been tagged in https://github.com/filecoin-project/boost/releases/tag/v1.0.1.

PR #681, opened 2022-08-03
feat: DagstoreDestroyShard API and CLI

LexLuthr:
A way to safely remove incorrect registered shards with the dagstore.


nonsense (approved):

LexLuthr:
Depends on https://github.com/filecoin-project/dagstore/pull/135 to work as dagstore does not return a nil on waiting channel on success.

PR #686, opened 2022-08-04
fix gostream compatibility

gmelodie:
I was getting this while trying to import boost:
```
# github.com/filecoin-project/boost/transport/httptransport
../../go/pkg/mod/github.com/filecoin-project/boost@v1.2.0/transport/httptransport/libp2p_server.go:251:33: undefined: gostream.Stream
```
It seems that `gostream` now uses `network.Stream` instead of `gostream.Stream`. This makes boost compatible with it again

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -72,7 +72,7 @@ require (
 	github.com/libp2p/go-eventbus v0.2.1
 	github.com/libp2p/go-libp2p v0.20.3
 	github.com/libp2p/go-libp2p-core v0.16.1
-	github.com/libp2p/go-libp2p-gostream v0.4.1-0.20220720161416-e1952aede109
+	github.com/libp2p/go-libp2p-gostream v0.4.0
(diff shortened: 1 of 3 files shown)


jacobheun:
This unfortunately doesn't work as the gostream and network stream interfaces aren't the same (tests will fail). ref https://github.com/libp2p/go-libp2p-gostream/pull/67. I'm not sure exactly what Boost is leveraging the different interface for but we need to fix the dependency branch usage here. @dirkmc has the full context here to help resolve and is back on Monday.

dirkmc:
I've added a commit to https://github.com/libp2p/go-libp2p-gostream/pull/67 so that it uses network.Stream instead of gostream.Stream
Once this change is merged and released I'll bubble up to this PR

rvagg:
hopefully solved @ https://github.com/filecoin-project/boost/pull/1009

gmelodie:
Closing in favor of #1009

Issue #683, opened 2022-08-04
Boost uses incorrect config for MaxPublishDealsFee


nonsense:
When publishing deals with Boost, at the moment the `spec` propagated to Lotus when signing/submitting a message is based on:

```
[LotusFees]
  MaxPublishDealsFee = "... FIL"
```

whereas it should probably be based on:

```
[Dealmaking]
  PublishMsgMaxFee = "... FIL"
```

---

Additionally we are using an incorrect value to `tag` funds on Boost side... i.e. the max price is different to the value we tag on Boost side.

PR #687, opened 2022-08-05
deal publishing / fees: remove redundant configs

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/683

---

Ultimately I think we don't need these values -- we added them before adding backward-compat support for Lotus Markets. After adding Lotus Markets, it looks like we moved to using the values from its config, making these three redundant.


dirkmc:
I'd suggest adding a config migration to remove the old keys from the config file itself:
https://github.com/filecoin-project/boost/blob/main/node/config/migrate.go#L20

PR #691, opened 2022-08-10
Add optional signature on query-ask v2

hannahhoward:
# Goals

@s0nik42 suggest a number of very smart improvements to the retrieval and storage protocols in #666. We probably need to think these through, but since we are in the middle of redoing the QueryAsk protocol for retrievals anyway, I made this follow on PR to #671 that adds support for optional signature verification.

# Implementation

- Make a signed query ask wrapper
- Verify the signature only if an optional client address is present
- Update tests
- Update libp2p protocol to send signed query asks
- Update CLI command to sign ask with wallet sig

### DIFF
--- a/client/retrieval.go
+++ b/client/retrieval.go
@@ -45,7 +45,7 @@ func NewRetrievalClient() (*RetrievalClient, error) {
 }
 
 // Query sends a retrieval query v2 to another peer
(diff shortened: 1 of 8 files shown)


s0nik42:
👍🏻

PR #693, opened 2022-08-10
feat: replace CarOffsetWriter with car.NewCarV1StreamReader

rvagg:
WIP:

* builds on https://github.com/ipld/go-car/pull/291 (plus 326, I'm assuming that'll be merged into that branch)
* car/ directory still exists, needs to be deleted
* better test coverage to cover some of the things that the tests in car/ were doing that we're not trusting go-car to do
* doesn't have the benefit of the `BlockInfoCache`, so this version will re-load all the blocks during a resumption - it should do it properly, but the existing version keeps a cache of the block's position in the CAR and all of the links in the block so we can take shortcuts with `merkledag.Walk`; we need to consider if/how we support that. Perhaps it's just a matter of holding onto the reader and reusing it, or holding onto some part of the reader? Since the new `TraverseResumer` should take care of these concerns for us.

### DIFF
--- a/transport/httptransport/libp2p_server.go
+++ b/transport/httptransport/libp2p_server.go
@@ -10,12 +10,15 @@ import (
 	"sync"
 	"time"
 
-	"github.com/filecoin-project/boost/car"
 	"github.com/filecoin-project/boost/transport/types"
 	"github.com/google/uuid"
 	"github.com/ipfs/go-cid"
 	blockstore "github.com/ipfs/go-ipfs-blockstore"
 	logging "github.com/ipfs/go-log/v2"
+	car "github.com/ipld/go-car/v2"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/ipld/go-ipld-prime/storage/bsadapter"
+	selectorparse "github.com/ipld/go-ipld-prime/traversal/selector/parse"
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/peer"
@@ -36,9 +39,7 @@ type Libp2pCarServer struct {
 	auth   *AuthTokenDB
 	bstore blockstore.Blockstore
 	cfg    ServerConfig
-	bicm   car.BlockInfoCacheManager
 
-	ctx           context.Context
 	cancel        context.CancelFunc
 	server        *http.Server
 	netListener   net.Listener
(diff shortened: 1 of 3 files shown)


rvagg:
I _think_ it should be possible to replicate the `BlockInfoCacheManager` pattern but associate a `PayloadCid` with the reader returned from `car.NewCarV1StreamReader` (and its associated closer that I make in here). Then when we come back to `serveContent()` for the same payload, we reuse it and the `readEmitter` that wraps around it will reset it with a `Seek()` then start passing on new `Read()`s.

I just don't know about parallel requests for the same `PayloadCid`, we can only use one reader at a time.

willscott:
Instead of making the stream readers somehow safe for multiple concurrent users, which will be tough given the semantics of read vs seek, it should be pretty cheap to be able to do a clone() to make a deep copy of metadata for a new user, and then maybe also not too bad to have a `merge` to be able to push additional learned metadata back into the primary-cached reader at the end of a read.

transport/httptransport/libp2p_server.go
```diff
@@ -183,12 +174,43 @@ func (s *Libp2pCarServer) checkAuth(r *http.Request) (string, *AuthValue, *httpE
...
+			}
+			cancelComplete <- struct{}{}
+		}()
+		select {
+		case <-parentCtx.Done():
+			return parentCtx.Err()
+		case <-cancelComplete:
+		}
+		return err
+	}
```
dirkmc, line 213: nit: I'd suggest moving the cancelContent function inside of `sendCar` as it's not used anywhere else

dirkmc (commented):
It would be great to get rid of CarOffsetWriter! 👍

transport/httptransport/libp2p_server.go
```diff
@@ -149,9 +142,7 @@ func (s *Libp2pCarServer) handler(w http.ResponseWriter, r *http.Request) {
 	defer s.h.ConnManager().Unprotect(pid, tag)
 
 	// Get a block info cache for the CarOffsetWriter
-	bic := s.bicm.Get(authVal.PayloadCid)
-	err = s.serveContent(w, r, authToken, authVal, bic)
-	s.bicm.Unref(authVal.PayloadCid, err)
+	_ = s.serveContent(w, r, authToken, authVal)
```
dirkmc, line 145: Maybe we should log this error somewhere?

jacobheun:
@rvagg what's the current state of this? It sounded like there were some challenges with the go-car updates but following the threads in the related issues wasn't quite clear. I'd like to understand what the blocker on this is so we can sort out next steps.

dirkmc:
Closing as out of date

PR #694, opened 2022-08-10
chore: update xcode version to supported

LexLuthr:
xcode 12.5.0 is no longer supported by circleCI
https://circleci.com/docs/testing-ios


jacobheun (approved):
This is fine to get fixed for the moment. I don't know what circle's deprecation timelines are but we should probably upgrade to 13.x soon.

PR #697, opened 2022-08-12
update deal status on client

LexLuthr:
Changed the deal status on client side.

TODO:
Figure out when deal status is "Complete" on server side
Update client side based on above finding

### DIFF
--- a/cmd/boost/deal_status_cmd.go
+++ b/cmd/boost/deal_status_cmd.go
@@ -173,9 +173,9 @@ func statusMessage(resp *types.DealStatusResponse) string {
 	case dealcheckpoints.PublishConfirmed.String():
 		return "Adding to Sector"
 	case dealcheckpoints.AddedPiece.String():
-		return "Announcing"
-	case dealcheckpoints.IndexedAndAnnounced.String():
 		return "Sealing"
+	case dealcheckpoints.IndexedAndAnnounced.String():
+		return "Complete"
 	case dealcheckpoints.Complete.String():
 		if resp.DealStatus.Error != "" {
 			return "Error: " + resp.DealStatus.Error


parkan:
thank you @LexLuthr!

just found this piece of documentation that suggests that Complete actually means "deal is over": https://github.com/filecoin-project/boost/blob/93f9b26abc0ee610fccba1cef870a9d7b98962b8/react/src/DealDetail.js#L563

which makes more sense with my errored deals being in "Complete" state; however, there is no state listed in that docstring that would correspond to "deal has finished sealing" (ActiveOnChain equivalent in Estuary)

the code here seems to suggest that once in IndexedAndAnnounced state the *actual* state depends on sealingState here: https://github.com/filecoin-project/boost/blob/93f9b26abc0ee610fccba1cef870a9d7b98962b8/gql/resolver.go#L536

(I believe this is used on the server side)

so IndexedAndAnnounced could mean either Sealing or "ActiveOnChain" depending on that result; I don't quite know enough about client vs server representation of the checkpoints to say how this should work but as a user I *really* need a useful value here that tells me deal is active/proving

(separately, I'd love a more specific Error state rather than Complete, looks like it has a special case for Cancelled but not for timeouts etc, might open a separate issue for that)

parkan:
my suggestion for clarifying the state names would be something like:

Complete (deal expired) -> Expired
Complete (timeout or other error) -> Error
IndexedAndAnnounced (sealing) -> Sealing
IndexedAndAnnounced (proving) -> Active

to me these feel like the most intuitive descriptions of what's happening and more lined up with the terminology lotus and estuary use

dirkmc:
@parkan that sounds correct to me 👍 

The storage deal status protocol is described here:
https://boost.filecoin.io/boost-architecture/libp2p-protocols#storage-deal-status-protocol

I suggest that:
- on the server side we
  - make a [call to lotus miner](https://github.com/filecoin-project/boost/blob/93f9b26abc0ee610fccba1cef870a9d7b98962b8/gql/resolver.go#L537) to request the sealing status
  - add a `DealStatus.SealingState` field with the sealing status as a string
- on the client side
  - Include the SealingState field in the JSON output

PR #377, opened 2022-03-29
[WIP] Bitswap integration in Boost

aarshkshah1992:

PR #707, opened 2022-08-17
booster bitswap MVP executable

hannahhoward:
# Goals

This is basically a "mashup" of booster-http, https://github.com/filecoin-project/boost/pull/377, and https://github.com/filecoin-project/dagstore/pull/116 to kick the tires and see how to wire everything up for an independent boost executable.

# Implementation

- Copy booster-http files
- Replace HTTPServer with bitswap server
- Migrate over the indexed blockstore from DAGStore PR and modify it to only use the available APIs
- Migrate over the piece selection function from  https://github.com/filecoin-project/boost/pull/377, modify to not use go-fil-markets

# TODO

- [ ] Tests, comments
- [ ] Read the libp2p key from the boostd repo
- [ ] Return [ipld `ErrNotFound`](https://github.com/ipfs/go-ipld-format/blob/master/merkledag.go#L15) from `GetSize()` if the block is not found, because [Bitswap checks for this error](https://github.com/ipfs/go-bitswap/blob/4e3736a773545bf09d5178a2be24043b76560f87/server/internal/decision/blockstoremanager.go#L97) to decide if a block is present
    Note that the RemoteBlockstore calls across an RPC boundary, so the code will need to do string matching to check if the error message contains "not found"
- [ ] Optimize DAG store PR
  - [ ] Make sure that `blockstore.Has` and `blockstore.GetSize` are efficient
  - [ ] Make sure that checking whether a block is in a piece that is unsealed a zero-cost is efficient

## Merge Note
This PR is using the base branch of release/lotus1.17.2, so that we can use that for landing all upcoming changes that require the a libp2p update that is landing in Lotus 1.17.2 (early October). This is a workaround until we get better separation of the dependency chain with Lotus.

# For discussion

Key issues identified:
- dagstore.AcquireShard is not available on the API (and I assume it won't be) so that means unless a CarV2 was stored, we're going to be indexing every time we pull a shard. Perhaps this moves up the priority of #573 or associated efforts to make indexes more easily accessible
- I couldn't figure out pricing without moving over a lot of go-fil-markets machinery around asks. We probably need to consider the API for figuring out of if a deal is free.


hannahhoward:
Note: the build is probably broken until the libp2p 0.21 updates filter through the Lotus stack
-- https://github.com/filecoin-project/lotus/pull/8970

dirkmc:
Thanks for putting this together 🙏 

1. With respect to the blockstore interface, I'd suggest we expose the blockstore methods that bitswap needs on the boostd API. For now we can implement it in boostd using the DAG store, and then we can switch over to the piece meta service when it's ready.
2. With regards to figuring out if a deal is free, we could just treat blocks in non-free deals as not found. eg Has(cid) would return false if the block is in a non-free deal.

willscott:
What's the plan for index update to indicate bitswap is available?

is this a separate libp2p identity (or at least multiaddr host / port endpoint) the provider will expose?

ribasushi:
> is this a separate libp2p identity (or at least multiaddr host / port endpoint) the provider will expose?

I'd say it should be configurable on the SP side: "announce using X and Y"

jacobheun:
I added an issue for tracking the index announcement work, https://github.com/filecoin-project/boost/issues/711, we should clarify what needs to happen in there. I added the address registration as a consideration we need to account for.

dirkmc:
I did a bit of refactoring to clean up some unused variables etc, and rebased on top of `main`

dirkmc:
I was able to retrieve a block using IPFS:
```
ipfs swarm connect <multiaddr>
ipfs get <cid>
```

dirkmc:
Note: As part of productionizing this PR we need to look at optimizations in the DAG store PR: at the moment it's not very efficient in the way that it checks if a block is in a piece that is unsealed and free (cost is zero) to download.

hannahhoward:
👍 on refactoring changes @dirkmc

hannahhoward:
@dirkmc I've made the requested changes in go-fil-markets and move the rest of the is-free-and-unsealed code within this repo.

jacobheun:
I've updated the base branch on this to release/lotus1.17.2, we can use that for landing all upcoming changes that require the libp2p update that will land with 1.17.2

jacobheun:
For the rest of the outstanding TODO's we want to complete after merging this, can we get issues for those created and tracked in #709 so we can move forward with the merge here?

Is there anything else outstanding to land this PR other than creating those issues? I'd like to land this by eod Tuesday to unblock the tracing integration.

Issue #705, opened 2022-08-16
Limit parallel HTTP downloads for storage deals


dirkmc:
Currently boost has a config option to limit parallel Graphsync downloads.

We should add a config option to
- limit parallel HTTP downloads per peer (eg max 20 concurrent downloads to peer A)
- limit parallel HTTP downloads overall (eg max 100 concurrent downloads to all peers)

Note that the limit is on boost downloading data for storage deals.

willscott:
i wonder if this is the right policy knob to give providers:
* at the highest level, it could be reasonable to allow for limitation of overall bandwidth "use 100mbit". that seems preferable over a number of streams, where if they get connected to by slow peers they'd have 100 downloads but wouldn't use any bandwidth.
* Why do we want to limit 1 peer? in the case next 6 months this seems like the primary effect is to encourage degraded service in our common use case of the HTTP gateways pulling a bulk of content from SPs?

nonsense:
> Why do we want to limit 1 peer?

If we are downloading 100 deal each 512MiB, ideally we want to have some concurrency factor, and not download them all at the same time, because then the sealing pipeline would be idle. We want to continuously stage data for the sealing pipeline and not get it downloaded at the same time and passed-through in a huge chunk.

willscott:
is this for downloading by the SP from data providing client, or retrieval downloads _from_ the SP

dirkmc:
I added a note to the description to clarify that the limit is on boost downloading data for storage deals (not retrieval)

nonsense:
the former - downloading by the SP (i.e. storage deal)

dirkmc:
For http retrieval there are existing tools that SPs can use to rate limit (eg nginx has rate limiting features)

hannahhoward:
@willscott for existing SPs, this is similar to the knobs they have for graphsync. I don't know if "keep it familiar" is the design goal, but my recollection is we tried to collect feedback before on what was ideal and this is what they wanted.

hannahhoward:
Is it a cross protocol knob? Like are we applying it to Legacy SPs as well?

Issue #676, opened 2022-08-03
Show number of active transfers on the deal transfers page


magik6k:
Would be nice to see more things there.

Maybe also a number/graph of unique providers we're retrieving data from / or even bandwidth graph per provider (though that may have slightly too high cardinality in some cases)

PR #710, opened 2022-08-18
Transfer limiter

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/705

#### Problem summary

Currently boost starts a new download for every storage deal that it receives. If a client sends 1000 deals, boost will open 1000 concurrent connections. Because bandwidth is limited, these downloads will proceed very slowly. Meanwhile the sealing pipeline will sit idle until a transfer finishes.

#### Solution

The Storage Provider should be able to adjust a parameter to limit the number of parallel transfers, so that they can keep their sealing pipeline busy.

For simplicity this PR only uses a single config value: an overall "soft" limit on the number of transfers.
See the explanation at https://github.com/filecoin-project/boost/pull/710/files#diff-dafd8b081e95b0e754470a4a23253b2e4a08ba06390015ac77d0e9a59cbe3a5aR30-R54

### DIFF
--- a/node/config/def.go
+++ b/node/config/def.go
@@ -97,7 +97,12 @@ func DefaultBoost() *Boost {
 			},
 
 			MaxTransferDuration: Duration(24 * 3600 * time.Second),
-			RemoteCommp:         false,
+
+			RemoteCommp: false,
+
+			HttpTransferMaxConcurrentDownloads: 20,
+			HttpTransferStallTimeout:           Duration(5 * time.Minute),
+			HttpTransferStallCheckPeriod:       Duration(30 * time.Second),
 		},
 
 		LotusDealmaking: lotus_config.DealmakingConfig{
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -167,12 +167,23 @@ type DealmakingConfig struct {
 
 	// The maximum amount of time a transfer can take before it fails
 	MaxTransferDuration Duration
+
 	// Whether to do commp on the Boost node (local) or on the Sealer (remote)
 	RemoteCommp bool
 
 	// The public multi-address for retrieving deals with booster-http.
 	// Note: Must be in multiaddr format, eg /dns/foo.com/tcp/443/https
 	HTTPRetrievalMultiaddr string
+
+	// The maximum number of concurrent storage deal HTTP downloads.
+	// Note that this is a soft maximum; if some downloads stall,
+	// more downloads are allowed to start.
+	HttpTransferMaxConcurrentDownloads uint64
+	// The period between checking if downloads have stalled.
+	HttpTransferStallCheckPeriod Duration
+	// The time that can elapse before a download is considered stalled (and
+	// another concurrent download is allowed to start).
+	HttpTransferStallTimeout Duration
 }
 
 type FeeConfig struct {
--- a/go.mod
+++ b/go.mod
@@ -99,6 +99,7 @@ require (
 	go.uber.org/atomic v1.9.0
 	go.uber.org/fx v1.15.0
 	go.uber.org/multierr v1.8.0
+	golang.org/x/exp v0.0.0-20210715201039-d37aa40e8013
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/tools v0.1.11
 	golang.org/x/xerrors v0.0.0-20220411194840-2f41105eb62f
--- a/node/config/doc_gen.go
+++ b/node/config/doc_gen.go
@@ -246,6 +246,27 @@ see https://docs.filecoin.io/mine/lotus/miner-configuration/#using-filters-for-f
 			Comment: `The public multi-address for retrieving deals with booster-http.
 Note: Must be in multiaddr format, eg /dns/foo.com/tcp/443/https`,
 		},
+		{
+			Name: "HttpTransferMaxConcurrentDownloads",
+			Type: "uint64",
+
+			Comment: `The maximum number of concurrent storage deal HTTP downloads.
+Note that this is a soft maximum; if some downloads stall,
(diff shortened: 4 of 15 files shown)


node/config/def.go
```diff
@@ -98,6 +98,10 @@ func DefaultBoost() *Boost {
 			},
 
 			MaxTransferDuration: Duration(24 * 3600 * time.Second),
+
+			HttpTransferMaxConcurrentDownloads: 20,
+			HttpTransferStallTimeout:           Duration(30 * time.Second),
```
magik6k, line 103: I'd set this much higher, maybe 5min, then tune based on what we observe. 30s is likely to happen quite often for many random reasons (temporary link drops, bgp reshufflings, temporary lock/io contention on either end)

node/config/types.go
```diff
@@ -167,6 +167,16 @@ type DealmakingConfig struct {
 
 	// The maximum amount of time a transfer can take before it fails
 	MaxTransferDuration Duration
+
+	// The maximum number of concurrent storage deal HTTP downloads.
+	// Note that this is a soft maximum; if some downloads stall,
+	// more downloads are allowed to start.
+	HttpTransferMaxConcurrentDownloads uint64
```
magik6k, line 174: This 'soft' limit probably only makes sense if we're not starting more streams with a peer which has stalled connections - e.g. when we're retrieving 20 streams at 1mb/s (where it's likely that some streams will stall for >30s) from a very overloaded estuary shuttle, starting more streams will only make things worse.

magik6k: Having hard per-peer limit could potentially help here

parkan: FWIW on our systems we want very high per host concurrency as they are general purpose HTTP servers and so routing is tuned for fairness and each connection is quite slow but overall throughput is very high and scales linearly with stream count (this is fairly common with many HTTP hosts, thus the popularity of "download accelerators"). To some extent this also happens naturally if TCP congestion control get stuck in avoidance mode.

Ideally not only would we have multiple parallel file downloads per host but also multipart within each file, like `aria2c --split=X --max-connection-per-server=X*n...` with X > 16. I believe some tooling like Filswan already does this with aria2.

parkan: To expand on the above a bit: obviously the two scenarios we are imagining are nearly polar opposites, and difficult to tell apart by heuristics on the boostd side. One immediate solution would be to make this configurable by the client (`--max-connection-per-peer`), with a default representing whichever case you think is most likely for a naive user.

For reference, we're currently doing 150 concurrent transfers per SP comfortably and testing how much higher it can go, this could easily go into thousands with multipart (scaling up our transfer capacity accordingly)

dirkmc: @parkan thanks for the comments, that makes sense 👍 
I've added some context to the PR description to make clear why SPs want this change.
For your use case I think the SP will still want to have some limit on the number of concurrent transfers, to make sure their sealing pipeline is always busy. It will just be a higher number. But certainly we will have an option whereby SPs can disable transfer limiting entirely.

dirkmc: @magik6k thanks for pointing that out. I agree that we should not open more connections to a peer that already has stalled transfers. I'll make that change 👍 

I would prefer to try to keep the number of configuration knobs to a minimum because in the past transfer limits have caused a lot of confusion for SPs. For example the graphsync download limit would often result in the download pipeline getting completely blocked up and SPs wouldn't know why they weren't getting any data.

To improve visibility we're going to show the transfer rate for each deal on the deals page and also in summary format on the transfers page (as well as https://github.com/filecoin-project/boost/issues/676).

I'd suggest we try out this simple change first and see how it works, then add more knobs if it's not sophisticated enough.

parkan: @dirkmc totally understand the SP side of this (one of mine actually ran into a stampede situation recently as well), just want to make sure we aren't getting shut out in the process. If the limits are high enough it may have limited impact on us as long as we get enough SPs.

LexLuthr:
@dirkmc Can we clean up https://github.com/filecoin-project/boost/blob/b5d35e12004868a93d7496f64ac3d759a94e2908/node/config/def.go#L82-L83 if they are not in use? As we are developing new config for boost deals then it would be ideal to get rid of unused options if possible.

dirkmc:
The transfer limiter is only for storage transfers made over HTTP.
So maybe we should rename `SimultaneousTransfersForRetrieval` and `SimultaneousTransfersForStorage` to `GraphsyncSimultaneousTransfersForRetrieval` and `GraphsyncSimultaneousTransfersForStorage`

LexLuthr:
We can rename them but unfortunately, they are still part of boost deal making config along with lotus deal making config. I would recommend removing them from boost dealing making config altogether as they server no function there and keep them within lotus deal making config that uses graphsync.

dirkmc:
Oh I see, I didn't realize they weren't being used at all.
I will make a follow-up PR to remove then entirely.

dirkmc:
[Remove unused vars from boost config](https://github.com/filecoin-project/boost/issues/733#top)

dirkmc:
@LexLuthr is [Remove unused vars from boost config](https://github.com/filecoin-project/boost/issues/733#top) something you would have time to take on?

LexLuthr:
@dirkmc Sure, let me give it a try. If I need help, I will reach out to you.

benjaminh83:
Hi guys. I'm sad to report, but in terms of fixing the client that sends 1000 deals, then this fix only moved the problem one step. It still clogs up Boost, as it will put all deals Queued, but still take from the reserve of available capacity and funds. 

image

So I will not have a boost node with 100TiB free space and 10000FIL on there, just so it doesn't clog up from Estuary sending me 1000s of deals that its never going to transfer :(

PR #735, opened 2022-08-26
Update config version to 3

LexLuthr:
1. Remove redundant config from boost deal making
2. Update config version to 3
3. Add "HTTPRetrievalMultiaddr" to the file by default

### DIFF
--- a/node/config/migrate.go
+++ b/node/config/migrate.go
@@ -15,13 +15,14 @@ var log = logging.Logger("cfg")
 
 // CurrentVersion is the config version expected by Boost.
 // We need to migrate the config file to this version.
-const CurrentVersion = 2
+const CurrentVersion = 3
 
 type migrateUpFn = func(cfgPath string) (string, error)
 
 var migrations = []migrateUpFn{
 	v0Tov1, // index 0 => version 0
-	v1Tov2, // index 1 => version 1
+	v1Tov2, // index 1 => version 3
+	v2Tov3, // index 2 => version 3
 }
 
 // This struct is used to get the config file version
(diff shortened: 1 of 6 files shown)


node/config/migrate.go
```diff
@@ -15,13 +15,14 @@ var log = logging.Logger("cfg")
...
 // We need to migrate the config file to this version.
-const CurrentVersion = 2
+const CurrentVersion = 3
 
 type migrateUpFn = func(cfgPath string) (string, error)
 
 var migrations = []migrateUpFn{
 	v0Tov1, // index 0 => version 0
-	v1Tov2, // index 1 => version 1
+	v1Tov2, // index 1 => version 3
```
dirkmc, line 24:
```suggestion
	v1Tov2, // index 1 => version 2
```

node/config/v2_to_v3.go
```diff
@@ -0,0 +1,31 @@
...
+		return "", fmt.Errorf("parsing config file %s: %w", cfgPath, err)
+	}
+
+	boostCfg, ok := cfg.(*Boost)
+	if !ok {
+		return "", fmt.Errorf("unexpected config type %T: expected *config.Boost", cfg)
+	}
+
+	// For the migration from v1 to v2 just remove fields and update config version and add
+	// comments to the file
```
dirkmc, line 20: I don't think we need this comment

dirkmc (commented):
A couple of nits but looks good 👍 

Are there any other boost deal-making params that are unused?

LexLuthr:
I don't see any other unused config.

node/config/v2_to_v3.go
dirkmc: I think you may have removed the wrong comment 😆

LexLuthr: Fixed.

dirkmc (approved):

PR #738, opened 2022-08-29
tracing for boost and booster-http

nonsense:
This PR is adding OpenTelemetry tracing to `boostd` and `booster-http` and generating a few dummy traces. It also includes an exporter which works with the devnet/monitoring docker stack we have.

Next step is to properly instrument `boostd`, as well as `booster-http` and propagate a tracing context from `booster-http` to `boostd` and marshal/unmarshal it correctly if this is not support out-of-the-box.

---

This PR is also adding `booster-http` to the devnet docker stack. Fixes: https://github.com/filecoin-project/boost/issues/766

cc @kylehuntsman 

-----

TODO:

- [x] fix tests


dirkmc (approved):
LGTM, nice work! 🙌

PR #743, opened 2022-08-30
consolidate markdown documentation

nonsense:
This PR is consolidating all README.md files in Boost into one main README.md

I've also removed some instructions that are duplicates across the various README.md

I've also moved `/build/devnet` into `/docker/devnet` as I don't think these Dockerfiles belong to `/build`.


README.md
```diff
@@ -4,79 +4,266 @@ Boost is a tool for Filecoin storage providers to manage data storage and retrie
...
-boostd --vv run
+3. Follow the guide at https://boost.filecoin.io/tutorials/how-to-store-files-with-boost-on-filecoin
+
+Note that above you already ran a command to export FULLNODE_API (and point it to your local devnet lotus daemon).
+
+Note also that the provider address is `t01000` and you will need to supply an appropriate `--storage-price` when using `boost deal` since the devnet has a minimum price. Alternatively, using "Settings" in the Boost web UI to set the deal price to zero.
+
+## Running Boost devnet in Docker
+
+### Building Docker images
```
airenas, line 207: In the future, I would suggest pushing images to dockerhub and skipping this step for new users to start quickly without the building step.

README.md
```diff
@@ -4,79 +4,266 @@ Boost is a tool for Filecoin storage providers to manage data storage and retrie
...
+
+1. Select Lotus version, for example: `lotus_version=1.17.1-rc2`. It must be the tag name of [the Lotus git repo](https://github.com/filecoin-project/lotus/tags) without `v` prefix.
+
+2. Select Boost version, for example: `boost_version=1.3.0-rc1`. Docker images for Boost will be built on the current codebase. The `boost_version` is just used to tag images. If you want to build images for a specific boost version then you have to checkout that version first.
```
airenas, line 211:
```suggestion
2. Select Boost version, for example: `boost_version=1.3.0-rc1`. 
```
The original phrase is already obsolete

README.md
```diff
@@ -4,79 +4,266 @@ Boost is a tool for Filecoin storage providers to manage data storage and retrie
...
 ```
 
-5. Interact with Boost
+### Start devnet docker stack
 
-Pass the client address (wallet) and the provider address to the `dummydeal` command.
-Note that
-- the client address is the address of a wallet with funds in `lotus wallet list`
-- you can find the provider address in `~/.boost/config.toml` under the config key `Wallets.Miner`
+1. Run
```
airenas, line 222: We need instruction, that a user must first `cd examples/devnet`

README.md
```diff
@@ -4,79 +4,266 @@ Boost is a tool for Filecoin storage providers to manage data storage and retrie
...
-## Running a devnet:
+### Cleaning up
 
-Follow the instructions in the [devnet guide](./documentation/devnet.md)
+To stop containers and drop everything:
+```
+docker compose down --rmi local
+
+rm -rf ./data
+```
```
airenas, line 267: Not sure, but maybe we need also the instruction to delete `/var/tmp/filecoin-proof-parameters`?

README.md
nonsense: Fixed. Also moved `examples/devnet` under `docker/devnet` -- I think the `docker-compose.yml` should be close to where we build the images.

README.md
nonsense: Done

README.md
nonsense: > In the future, I would suggest pushing images to dockerhub and skipping this step for new users to start quickly without the building step.

Agree, but I am not sure we should document the pushing here really... Pushing of images is going to happen within the team, by whoever has access to the Docker Registry.

---

We should make sure we have support to build Boost from source, because most developers on the team would be running a dev version of the Boost image, and sometimes even of Lotus images.

PR #744, opened 2022-08-30
fix: dagstore destroy-shard

LexLuthr:
the check keep failing each time for errored out shards. We should directly issue a deleted instead of checking. incorrect CID will be covered by error thrown from destroy call itself.


dirkmc:
I'm not sure if this is going to fix the issue you mentioned. If there's any kind of error, BoostDagstoreDestroyShard will return the error on line 459, so the code you removed was redundant anyway.

LexLuthr:
This is just to remove the redundant code. The slack issue is different. I will open a new issue for it once I have everything ready for it.

rvagg (approved):
:thumbsup: redundant

dirkmc (approved):

PR #747, opened 2022-08-31
change network name for devnet

nonsense:
This PR is changing the default network name of the devnet from `devnet_default` to `devnet`. The idea is that we'll have to connect other containers to it, i.e. `prometheus` and/or `tempo` and maybe others from the `monitoring` stack.

cc @airenas


dirkmc (commented):
LGTM 👍

PR #751, opened 2022-08-31
feat: dagstore lookup-cid command

LexLuthr:


cmd/boostd/dagstore.go
```diff
@@ -309,3 +311,46 @@ var dagstoreDestroyShardCmd = &cli.Command{
 		return nil
 	},
 }
+
+var dagstoreLookupCmd = &cli.Command{
+	Name:      "lookup-cid",
```
LaurenSpiegel, line 316: wouldn't you want to be more explicit that you are looking up a piece cid?  `lookup-piece-cid`?

LexLuthr: The description provides the functionality details. I would prefer to keep command name to be short.

nonsense: I agree with @LaurenSpiegel that it'd be better to be explicit here as `cid` it very generic. There is the `alias` parameter that you could use to set a short command alias (such as `lpc` or something else).

nonsense (approved):

PR #760, opened 2022-09-02
monitoring docker stack

nonsense:
This PR is adding a docker compose stack for our monitoring stack (tracing, etc.)

It is also modifying the `devnet` stack to build `boost` from the local working directory.


nonsense:
cc @airenas

dirkmc (approved):

Issue #742, opened 2022-08-30
Improve devnet docker setup


nonsense:
This issue is a follow-up from https://github.com/filecoin-project/boost/issues/695

Ultimately we need a devnet setup (lotus + lotus-miner + boost) that is:
- easy to use by new-comers to Boost, so that they can try Boost quickly and help them get started with Boost
- very easy way to submit a test deal, and retrieve it, and experience Filecoin + Boost
- easy to develop against - in general this means that we should be able to quickly iterate on Boost, but sometimes this also involves changes to Lotus, so it should be easy to quickly rebuild and redeploy individual images

Improvements / Tasks:
- [x] https://github.com/filecoin-project/boost/pull/743 Documentation needs improvement - this is mostly done, but documentation needs a bit of consolidation -- right now we have some docs:
    -- at the main README.md
    -- at examples/devnet/README.md
    -- at documentation/devnet.md
    -- at build/devnet/README.md
    -- at https://boost.filecoin.io/

We should consolidate these, keep a minimal README.md regarding the devnet, and everything else should go to https://boost.filecoin.io/

- [x] #759
- [x] Add support for building Lotus / Lotus-miner images from local source
- [ ] Support other architectures, not just x86_64. For example the setup right now doesn't work on M1 or M2.
- [x] Modify the lotus-miner configuration (~/.lotusminer/config.toml) to:
-- disable BatchPreCommits
-- disable AggregateCommits
-- reduce WaitDealsDelay
- [x] Support 8MiB sectors, rather than 2KiB sectors.
- [ ] Change `npm install` to `npm ci` -- I think `npm install` also updates dependencies, whereas we want to install the dependencies as per the committed `lock` file
- [ ] Use temp directories, when checking out Lotus to build images, rather than local paths. This way the OS can cleanup the temp directories on restart (or we do cleanup after building), and we keep the local git repo and working directory clean.
- [ ] https://github.com/filecoin-project/boost/issues/767 Utilize Docker caches (image layers) to make building of Boost image faster - at the moment building a Boost docker image is rather slow due to the fact that we have to download a lot of dependencies (go.mod etc.) -- I suggest adding an intermediate Docker builder (in the multistage Dockerfile) that builds and caches Boost/`main` branch, which we can periodically clean and rebuild.
- [x] #779

---

This setup should be easy to integrate with the monitoring/tracing setup (grafana + prometheus + tempo)

nonsense:
[...]

PR #761, opened 2022-09-02
Add support for building Boost images from local source

airenas:
2 tasks for #742

**Implemented**:

- Add support for building Boost images from local source
- Change npm install to npm ci 

Now the build process uses local dir when creating `boost` and `boost-gui` docker images.


airenas:
it is covered by #760

PR #771, opened 2022-09-07
Allow Boost RPC API to capture parent trace contexts

### LINKED ISSUE #769: Marshal/Unmarshal tracing context in Boost/Lotus JSON-RPC
We need to do that in order to propagate and link traces from `booster-http` and `boostd`.

kylehuntsman:
This PR enables the ability for Opencensus or OpenTelemtry span contexts passed into Boost RPC client calls to be captured by Boost RPC server, nesting Boost RPC traces under the parent span passed to a given RPC method. Fixes #769.

`filecoin-project/go-jsonrpc` already handles span contexts, but does so with Opencensus (OC) while boost is using Opentelemetry (OT). OC and OT trace spans are not compatible. However, OT has an OC API wrapper they call a bridge that allows us to use our OT tracer as the default OC tracer. This PR creates that bridge and sets up the default OC tracer to our OT tracer. 

### Client Side
[`filecoin-project/go-jsonrpc/client.go`](https://github.com/filecoin-project/go-jsonrpc/blob/v0.1.5/client.go#L487-L491) handles the given context and starts a new span from it called `api.call`. It puts this span context into a property on the request called [`Meta.SpanContext`](https://github.com/filecoin-project/go-jsonrpc/blob/v0.1.5/client.go#L509-L517).

### Server Side
[`filecoin-project/go-jsonrpc/handler.go`](https://github.com/filecoin-project/go-jsonrpc/blob/9918975d9439a30446a5ebd96b77f73c317da7b4/handler.go#L192-L213) then looks for `Meta.SpanContext` in the request object and decodes it into a `SpanContext`, starting a new span with it.

### Additionally...
- booster-http was not being built with `make build/all`, so I added it to `make build/all` and `make push/all`.


node/impl/boost.go
```diff
@@ -195,6 +195,10 @@ func (sm *BoostAPI) BoostDagstoreListShards(ctx context.Context) ([]api.Dagstore
 }
 
 func (sm *BoostAPI) BoostDagstorePiecesContainingMultihash(ctx context.Context, mh multihash.Multihash) ([]cid.Cid, error) {
+	ctx, span := tracing.Tracer.Start(ctx, "Boost.BoostDagstorePiecesContainingMultihash")
```
nonsense, line 198: our span names are a bit too long, we should figure a scheme to shorten them.

nonsense (approved):

Issue #774, opened 2022-09-07
Move tracing behind feature flag and add configuration for it


nonsense:

nonsense:
Fixed at https://github.com/filecoin-project/boost/pull/778 and https://github.com/filecoin-project/boost/pull/776

PR #776, opened 2022-09-08
Add --enable-tracing flag to booster-http

kylehuntsman:
Adds a `--enable-tracing` flag to booster-http. The flag is false by default, preventing the instantiation and startup of the tracing exporter. The flag is set to true for the devnet docker setup. Part of #774.


dirkmc (approved):
LGTM 👍

Issue #737, opened 2022-08-29
Limit or reject slow clients


LexLuthr:
With the connection limiting feature in boost, users can control how many simultaneous download can happen at once.
If a malicious or bad actor in the network makes multiple deals at once and all these deals are downloading at a very slow speed then Boost node will not be able to onboard more deals.
This slow down will results in good deals getting rejected or not being able to seal on time due to slow downloads of bad deals.

We need to make boost more robust against such bad deals and  allow some kind of limiter/rejected either based on current download speed of a deal or based on historical data for that client.

I think an outright rejection based on preset speed (set by user) for a limited time would be more robust but that can result in uptick of network chatter.

scaseye:
estuary for example has at least 4 shuttles, the SPs connection speed to each shuttle will vary based on location.

is there a way to distinguish or limit based on the peer id in the transfer params?  estuary has one main peer id, but uses a sub peerid (the shuttle peerid) on the transfers.  

i would love to reject deals from the slow shuttles, and only except from the faster ones.

dirkmc:
We're about to release an [HTTP Transfer Limiter](https://boost.filecoin.io/configuration/http-transfer-limit) that should help out somewhat. There's a detailed explanation there, but essentially it will favour faster connections over slower connections.

jcace:
I'm having issues with this too. The HTTP Transfer Limiter seems like a good approach, do you know when/what version of Boost it will be released in?

LexLuthr:
@jcace We will cutting a release candidate tomorrow. Boost v1.4.0-rc2

jcace:
Just tested this out in Boost v1.4.0-rc2, works wonderfully. Nice work!

My only concern with it is I'm still running in to an issue where the queue is clogged up with a bunch of slow transfers. I have `LOTUS_DEALMAKING_HTTPTRANSFERMAXCONCURRENTDOWNLOADS` set to `10`, but now all 10 of those transfers are going slowly, will take several hours to complete, blocking faster transfers from getting picked up. 

I read that the algorithm will prefer deals with peers that don't currently have ongoing transfers, but I wonder if it would also be useful to implement another configuration parameter, i.e, `MaxConcurrentDownloadsPerPeer` which could be set to limit transfers within a single peer only, instead of a limit applied to all transfers. This way the download queue could not be saturated with slow transfers and could leave room for some other peers.

dirkmc:
[...]

PR #777, opened 2022-09-08
Fail the deal if the transfer fails

dirkmc:
This should fix the problem described at https://github.com/filecoin-project/boost/issues/737#issuecomment-1240300160

### DIFF
--- a/storagemarket/deal_execution.go
+++ b/storagemarket/deal_execution.go
@@ -143,35 +143,11 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, deal *types.Provide
 				return derr
 			}
 
-			if err := p.transferAndVerify(dh.transferCtx, pub, deal); err != nil {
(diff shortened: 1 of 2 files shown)


storagemarket/deal_execution.go
```diff
@@ -143,35 +143,11 @@ func (p *Provider) execDealUptoAddPiece(ctx context.Context, deal *types.Provide
...
-				// If the transfer failed because the user cancelled the
-				// transfer, it's non-recoverable
-				if dh.TransferCancelledByUser() {
-					return &dealMakingError{
-						retry: types.DealRetryFatal,
-						error: fmt.Errorf("data transfer cancelled after %d bytes: %w", deal.NBytesReceived, err),
-					}
-				}
-
-				// If the transfer failed because boost was shut down, it's
```
nonsense, line 165: @dirkmc shouldn't we leave this block around, i.e. if the context is canceled, do not mark the deal as fatally failed?

PR #778, opened 2022-09-08
tracing behind feature-flag for boostd; configuration in config.toml

nonsense:
This PR is adding the `[Tracing]` configuration section in the Boost node config.toml

It also moves the instantiation of the tracing system within the Boost node dependency injection.

It is also renaming the flag on `booster-http` from `--enable-tracing` to just `--tracing`, and adds `--tracing-endpoint`.

### DIFF
--- a/docker/devnet/docker-compose.yaml
+++ b/docker/devnet/docker-compose.yaml
@@ -55,6 +55,9 @@ services:
       - LOTUS_API_LISTENADDRESS=/dns/boost/tcp/1288/http
       - LOTUS_PATH=/var/lib/lotus
       - LOTUS_MINER_PATH=/var/lib/lotus-miner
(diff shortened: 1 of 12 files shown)


docker/devnet/docker-compose.yaml
```diff
@@ -55,6 +55,9 @@ services:
       - LOTUS_API_LISTENADDRESS=/dns/boost/tcp/1288/http
       - LOTUS_PATH=/var/lib/lotus
       - LOTUS_MINER_PATH=/var/lib/lotus-miner
+      - LOTUS_TRACING_ENABLED=true
```
nonsense, line 58: We should rename this from `LOTUS` to `BOOST` at some point.

PR #780, opened 2022-09-08
Configurable download limit per host

dirkmc:
Limit the amount of space that can be reserved for downloaded data per-host. This helps mitigate against a single host with a slow connection reserving all the download space, when there may be other hosts with faster connections that have not had a chance to propose a deal yet.

For example
- the total download space is 100GB
- the percentage per host is limited to 50%

This means that the total amount of space for downloaded data per host is 50GB. The reserved space is the sum of
- the amount of data queued for download
- the amount of data currently transferring
- the amount of data that is in the download staging area (waiting to be handed off to the sealing subsystem)

TODO:
- [x] Sanity check config params (eg MaxStagingDealsPercentPerHost must be between 0 - 100)

```
[Dealmaking]
  MaxStagingDealsBytes = 1000
  MaxStagingDealsPercentPerHost = 50
```

### DIFF
--- a/storagemanager/storagemanager.go
+++ b/storagemanager/storagemanager.go
@@ -26,13 +26,6 @@ type Config struct {
 	MaxStagingDealsPercentPerHost uint64
 }
 
-func (c Config) check() error {
-	if c.MaxStagingDealsPercentPerHost > 100 {
-		return fmt.Errorf("MaxStagingDealsPercentPerHost is %d but it must be a percentage between 0 - 100", c.MaxStagingDealsPercentPerHost)
-	}
-	return nil
-}
-
 type StorageManager struct {
 	lr                 lotus_repo.LockedRepo
 	db                 *db.StorageDB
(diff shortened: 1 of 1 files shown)


storagemanager/storagemanager.go
```diff
@@ -22,7 +22,15 @@ var (
...
 type Config struct {
-	MaxStagingDealsBytes uint64
+	MaxStagingDealsBytes          uint64
+	MaxStagingDealsPercentPerHost uint64
+}
+
+func (c Config) check() error {
+	if c.MaxStagingDealsPercentPerHost > 100 {
+		return fmt.Errorf("MaxStagingDealsPercentPerHost is %d but it must be a percentage between 0 - 100", c.MaxStagingDealsPercentPerHost)
+	}
```
nonsense, line 32: Nit: I suggest embedding inline this, and getting rid of `check()`. I don't see much value in having it extracted in a separate function.

nonsense (approved):

parkan:
THANK YOU!!!

PR #781, opened 2022-09-09
get bitswap branch back to tagged Lotus

hannahhoward:
# Goals

Get Bitswap branch back to tagged Lotus so we can merge it.

# Implementation

- update go.mod to require golang version 1.17 -- 1.17 can reduce module tree depedencies significantly, plus Lotus and other repos are up to date. I'm not entirely sure if this is needed but I think it is.
- the build issue we had arised from our dependence on Lotus's Libp2p modules -- lp2p and the NetAPI implementation. This essentially copies them over in libp2p 0.21 form.


hannahhoward:
So close, and yet, so far.

Things fall down for integration tests.

I think the next move would be to extract itests/kit from Lotus to it's own repo and build a CLI around it. Then, once installed, instead of using all the code in repo, just run a cli to startup the other processes as part of running itests

jacobheun:
Closing as this has been resolved and merged in a separate pr.

PR #786, opened 2022-09-12
LoadBalancer for bitswap (and later, more of libp2p)

hannahhoward:
# Goals

Implement a simple libp2p load balancer for booster bitswap based on https://pl-strflt.notion.site/Simple-Libp2p-Load-Balancer-a1f1d8c182f74ea5bba4b48ef5a4992d

The design was simplified significantly based on #801 

I've also made the job of configuring the load balancer much simpler -- when you start booster-bitswap for the first time, it automatically sets the load balancers configured peer id for bitswap

# Implementation

- add message types as IPLD serializable go types
- add utility functions for writing various message types to streams
- implement the load balancer based on the psuedo code in the load balancer design document. Load balancer:
   - takes a set peer config of expected service nodes and the protocols they will listen on
   - when service nodes connect, activates all protocols registered for service node
   - when service nodes disconnect, deactivates all protocols registered for service node
   - handles inbound and outbound forwarding requests
   - peer routing configuration can be updated as the cost of all service nodes disconnecting and reconnected
 - implement a service node based on the psuedo code in the load balancer design document. Service node handles:
   - handling inbound and outbound forwarding requests
- config for bitswap peer id in boost
   - if this is set, go ahead and list in the available retrieval protocols 
- add a module to start the load balancer within boost
   - only accept routing requests currently from set bitswap peer id
- add APIs to programmatically get/set the bitswap peer id
   - when set, updates the peer config in the load balancer 
- in booster bitswap -
   - on start up, load or generate a peer key (previous, generated a new one every time)
   - query the boost for the currently set bitswap peer id
   - if peer id is not set, set it to the newly generated peer id
   - if peer id is set, verify it matches booster bitswaps peer id
      - if no match, error out, unless a special override-peer-id CLI option is passed, which tells booster-bitswap to replace the bitswap peer ID in Boost config with the current booster-bitswap peer

I have verified the load balancer works round trip -- I was able to get a bitswap block from boost's main peer ID


dirkmc:
Thanks for putting this together Hannah.

I have some general questions about the load balancer implementation:
1. Can we assume that the booster-bitswap node has the same peer ID as the load balancer? If so, then I think it simplifies things.
2. Can the load balancer register routes as soon as it gets a connection from the booster-bitswap node, and remove the routes when the connection goes down? Then we wouldn't need a lot of the messages. For example we wouldn't need the renew / termination messages and logic.

The way I'd imagine this working:

In boostd config (maybe in a LoadBalancer section) we could have
  - the protocols that the routed host serves
  - the multiaddr of the routed host
When the load balancer gets a connection from the routed host, it starts listening on the protocols in config.
When the load balancer gets an inbound connection over these protocols
- if there is no connection to the routed host, it does a stream reset
- if there is a connection to the routed host it just passes data through to the routed host as-is

On the booster-bitswap side:
When bitswap creates a new outbound connection, we would need to add some intercept code that would send a message to the load balancer telling it to open the connection to the peer. From that point forward it would just pass through any inbound and outbound data over the connection.

loadbalancer/loadbalancer.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+func (lb *LoadBalancer) terminateRoutes(p peer.ID, protocols []protocol.ID) error {
+
+	// check routes to verify ownership
+	for _, id := range protocols {
+		registeredPeer, ok := lb.routes[id]
+		if !ok || p != registeredPeer {
+			// error if this protocol not registered to this peer
+			return ErrNotRegistered{p, id}
```
rvagg, line 201: is this a bit too harsh for a "remove them all please!"? should we instead terminate any that exist but also return an `ErrNotRegistered` if any of those didn't exist rather than bail if just one doesn't exist?

loadbalancer/loadbalancer.go
```diff
@@ -0,0 +1,367 @@
...
+	p := s.Conn().RemotePeer()
+	request, err := messages.ReadForwardingRequest(s)
+	if err != nil {
+		log.Warnf("reading forwarding request: %s", err)
+		_ = s.Reset()
+		return
+	}
+
+	defer s.Close()
+
```
rvagg, line 243:
```suggestion
```
dupe

loadbalancer/loadbalancer_test.go
```diff
@@ -0,0 +1,633 @@
...
+			willErrorOpening: false,
+			willErrorReading: true,
+			registerHandler:  true,
+			rejectResponse:   true,
+		},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			//ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+			//defer cancel()
```
rvagg, line 453: were they taking too long or did you forget to uncomment?

loadbalancer/messages/messages.ipldsch
```diff
@@ -0,0 +1,46 @@
+type RequestKind enum {
+	| New ("n")
```
rvagg, line 2:
```suggestion
  | New ("n")
```

loadbalancer/messages/messages.ipldsch
```diff
@@ -0,0 +1,46 @@
...
+type ResponseCode enum {
+  | Ok ("o")
+  | Rejected ("r")
+} representation string
+
+type DateTime int
+
+type RoutingResponse struct {
+  Code ResponseCode (rename "c")
+  Message String (rename "m") # more info if rejected
```
rvagg, line 22: maybe `optional` too?

rvagg: there's probably a case for an inline union here with a success and failure subtype, keyed by code, something like:

```ipldsch
type RoutingResponse union {
  | RoutingResponseOk "o"
  | RoutingResponseRejected "r"
} representation inline {
  discriminantKey "c"
}

type RoutingResponseOk struct {
  Message String (rename "m")
}

type RoutingResponseRejected struct {
  Expiry DateTime
}
```

... having written all of that; I don't think this is supported in bindnode, maybe it's not even in the current Go schema parser! I guess there's a TODO here because the union would give you good validation on the existence of the two different properties.

loadbalancer/messages/messages.ipldsch
```diff
@@ -0,0 +1,46 @@
...
+  Expiry optional DateTime
```
rvagg, line 23:
```suggestion
  Expiry optional DateTime (rename "e")
```

loadbalancer/messages/messages.ipldsch
```diff
@@ -0,0 +1,46 @@
...
+type ForwardingKind enum {
+  | Inbound ("i")
+  | Outbound ("o")
+} representation string
+
+type PeerID string
+type PubKey bytes
+
+type ForwardingRequest struct {
+   Kind ForwardingKind (rename "k")
```
rvagg, line 35: another case for a union here; we should look at doing this so we have protocol level enforcement of types

loadbalancer/servicenode.go
```diff
@@ -0,0 +1,445 @@
...
+			case <-registeredHandler.closing:
+				registeredHandler.closeLk.Unlock()
+				continue
+			default:
+			}
+			close(registeredHandler.closing)
+			registeredHandler.closeLk.Unlock()
+
+			// stop any renew timers
+			registeredHandler.renewTimer.Stop()
```
rvagg, line 93:
```suggestion
			if registeredHandler.renewTimer != nil {
				registeredHandler.renewTimer.Stop()
			}
```

loadbalancer/servicenode.go
```diff
@@ -0,0 +1,445 @@
...
+		return
+	case <-sn.ctx.Done():
+		return
+	default:
+	}
+	close(handler.closing)
+	handler.closeLk.Unlock()
+
+	// stop any timers on the channel
+	handler.renewTimer.Stop()
```
rvagg, line 424:
```suggestion
	if handler.renewTimer != nil {
		handler.renewTimer.Stop()
	}
```

loadbalancer/messages/messages.ipldsch
hannahhoward: Indeed. At the moment, the bindnode go union type implementation sucks enough that I don't want to deal with it. (I believe currently it's a struct with a bunch of nullable members?)

loadbalancer/messages/messages.ipldsch
hannahhoward: I believe that Union types are supported by the parser (https://github.com/ipld/go-ipld-prime/blob/7548eb883bda4712355797547a0628a0ad1c00cb/node/bindnode/infer.go#L221) but the implementation in go is pretty rough -- a wrapper struct with a set of members that are all nullable for each potential type in the union. Eek! Not sure what type safety you actually get out of that at least on the go-side, but seems like there are lots of ways to make said struct fail when actually serialized. It is enough that it makes me not want to use it.

loadbalancer/loadbalancer_test.go
hannahhoward: it would appear so

node/modules/retrieval.go
```diff
@@ -58,3 +70,74 @@ func HandleRetrievalTransports(lc fx.Lifecycle, l *lp2pimpl.TransportsListener)
...
+			bsPeerID, err := peer.Decode(cfg.Dealmaking.BitswapPeerID)
+			if err != nil {
+				return nil, err
+			}
+			peerConfig[bsPeerID] = []protocol.ID{
+				network.ProtocolBitswap,
+				network.ProtocolBitswapNoVers,
+				network.ProtocolBitswapOneOne,
+				network.ProtocolBitswapOneZero,
+			}
```
dirkmc, line 87: The list of bitswap protocols is repeated in a couple of places, can we move to a constant?

node/modules/retrieval.go
```diff
@@ -58,3 +70,74 @@ func HandleRetrievalTransports(lc fx.Lifecycle, l *lp2pimpl.TransportsListener)
...
+		}
+		return loadbalancer.NewLoadBalancer(h, peerConfig)
+	}
+}
+
+func HandleLoadBalancer(lc fx.Lifecycle, lb *loadbalancer.LoadBalancer) {
+	lc.Append(fx.Hook{
+		OnStart: func(ctx context.Context) error {
+			log.Debug("starting load balancer")
```
dirkmc, line 96: I'd suggest we output these logs at the info level. This is an important component.

node/modules/retrieval.go
```diff
@@ -58,3 +70,74 @@ func HandleRetrievalTransports(lc fx.Lifecycle, l *lp2pimpl.TransportsListener)
...
+		}
+		err = lb.UpdatePeerConfig(peerConfig)
+		return
+	}, nil
+}
+
+func NewGetBitswapPeerIDFunc(r lotus_repo.LockedRepo) (dtypes.GetBitswapPeerIDFunc, error) {
+	return func() (p peer.ID, err error) {
+		var pString string
+		err = mutateCfg(r, func(cfg *config.Boost) {
```
dirkmc, line 130: Should this be readCfg instead of mutateCfg?

node/modules/retrieval.go
```diff
@@ -58,3 +70,74 @@ func HandleRetrievalTransports(lc fx.Lifecycle, l *lp2pimpl.TransportsListener)
...
+			pString = cfg.Dealmaking.BitswapPeerID
+		})
+		if err != nil {
+			return
+		}
+		if pString == "" {
+			err = fmt.Errorf("no bitswap peer id set")
```
dirkmc, line 137: For simplicity I'd suggest we just return "" here instead of an error. That way we don't need to do string matching across the API boundary to detect the string "no bitswap peer id set"

loadbalancer/errors.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+)
+
+// ErrAlreadyRegistered indicates a protocol has already been registered as a route
+type ErrAlreadyRegistered struct {
```
dirkmc, line 12: It looks like several of the errors in this file are no longer used. Can we remove them?

loadbalancer/loadbalancer.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/filecoin-project/boost/loadbalancer/messages"
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+	"github.com/multiformats/go-multiaddr"
+)
+
+const DefaultDuration = time.Hour
```
dirkmc, line 20: Looks like this is no longer used

loadbalancer/loadbalancer.go
```diff
@@ -0,0 +1,289 @@
...
+
+var log = logging.Logger("loadbalancer")
+
+type LoadBalancer struct {
```
dirkmc, line 24: I agree with Rod's comment that LoadBalancer is somewhat of a misnomer. For example it doesn't make sense to load balance bitswap requests as bitswap is stateful.

loadbalancer/loadbalancer.go
```diff
@@ -0,0 +1,289 @@
...
+	return nil
+}
+
+func (lb *LoadBalancer) Start(ctx context.Context) {
+	lb.ctx = ctx
+	lb.h.SetStreamHandler(ForwardingProtocolID, lb.handleForwarding)
+	lb.h.Network().Notify(lb)
+}
+
+func (lb *LoadBalancer) Close() error {
```
dirkmc, line 65: I'd suggest removing the error return value, as it's always nil

loadbalancer/protocols.go
```diff
@@ -0,0 +1,11 @@
+package loadbalancer
+
+import (
+	"github.com/libp2p/go-libp2p-core/protocol"
+)
+
+// RegisterRoutingProtocolID identifies the protocol for registering routes with a libp2p load balancer
+const RegisterRoutingProtocolID protocol.ID = "/libp2p/balancer/register-routing/0.0.1"
```
dirkmc, line 8: I believe this is no longer used

loadbalancer/servicenode.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/filecoin-project/boost/loadbalancer/messages"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+)
+
+// ServiceNode is a host that behaves as a service node connected to a load balancer
+// -- all traffic is routed through the load balancer for each registered protocol
+type ServiceNode struct {
```
dirkmc, line 17: I'd suggest renaming ServiceNode to something more specific, like "ForwardingHost"

loadbalancer/servicenode.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// read the forwarding request
+	request, err := messages.ReadForwardingRequest(s)
+	if err != nil {
+		log.Warnf("reading forwarding request: %s", err)
+		_ = s.Reset()
+		return
+	}
+
+	log.Debugw("received forwarding request for protoocl", "protocols", request.Protocols, "remote", request.Remote)
```
dirkmc, line 100:
```suggestion
	log.Debugw("received forwarding request for protocol", "protocols", request.Protocols, "remote", request.Remote)
```

cmd/booster-bitswap/run.go
```diff
@@ -125,3 +138,18 @@ func getBoostAPI(ctx context.Context, ai string) (api.Boost, jsonrpc.ClientClose
 
 	return api, closer, nil
 }
+
+func dataDirPath(ctx *cli.Context) string {
+	dataDir := ctx.String("data-dir")
```
dirkmc, line 143: The "data-dir" flag doesn't seem to be defined in the run command.
I'd suggest using cmd.FlagRepo for consistency with the `boost` client.

dirkmc:
If I understand correctly, the load balancer
- starts listening on bitswap protocols when the bitswap node connects to it
- stops listening on bitswap protocols when the bitswap node disconnects from it

From a client's perspective, whenever they open a stream to the load balancer over bitswap they will get a "protocol not supported" error. To a client this looks like a more long-term state: "This server doesn't support bitswap".

Would it make more sense for the client to get a stream reset instead? To a client that would look like a short term error state.

dirkmc:
With respect to the mechanism of booster-bitswap automatically registering its peer ID with the load balancer, it may cause issues to close and restart the load balancer after it's already been started.

I'd suggest instead that we add a `booster-bitswap init` command (similar to `boost init`) that creates a peer ID. Then someone who wants to add a booster-bitswap node on their SP would:
1. Run `booster-bitswap init` to create a peer ID. If it has already been inited, it just outputs the already created peer ID.
2. Add the bitswap peer ID to boostd's config.toml
3.  Restart boostd
4. Start booster-bitswap

hannahhoward:
re: booster-bitswap init

It seems like it would be nice for it to use the API call to copy the value directly. What's stored on disk is the private key, not exactly obvious how to direct copy it. But I agree that it's reasonable to expect a restart rather than hot reload.

hannahhoward:
I've updated the PR so that:
- configured routes where the service node is not connected return stream reset, instead of protocol not supported
- the peer id can be configured through the booster-bitswap init command
- no longer attempting to support hot reloading.

cmd/booster-bitswap/init.go
```diff
@@ -0,0 +1,154 @@
...
+
+type PeerIDAPI interface {
+	DealsGetBitswapPeerID(ctx context.Context) (peer.ID, error)
+	DealsSetBitswapPeerID(ctx context.Context, p peer.ID) error
+}
+
+func configureRepo(ctx context.Context, cfgDir string, papi PeerIDAPI, overrideExistingPeerID bool) (crypto.PrivKey, error) {
+	if cfgDir == "" {
+		return nil, fmt.Errorf("dataDir must be set")
+	}
```
dirkmc, line 32: Now that the repo flag has a default value, I don't think we need this check

cmd/booster-bitswap/init.go
```diff
@@ -0,0 +1,154 @@
...
+	existingPid, err := papi.DealsGetBitswapPeerID(ctx)
+	if err != nil {
+		return nil, err
+	}
+	peerIDNotSet := existingPid == peer.ID("")
+	matchesPid := existingPid == selfPid
+	log.Infow("get/set peer id of bitswap from boost", "local", selfPid.String(), "boost", existingPid.String(), "boost not set", peerIDNotSet, "override", overrideExistingPeerID)
+	// error if a peer id is set that is different and we aren't overriding
+	if !peerIDNotSet && !matchesPid && !overrideExistingPeerID {
+		return nil, errors.New("bitswap peer id does not match boost node configuration. use --override-peer-id to force a change")
```
dirkmc, line 56: Looks like the `--override-peer-id` flag is no longer defined

dirkmc:
This is looking really good 👍 

> What's stored on disk is the private key, not exactly obvious how to direct copy it.

I don't think you need to directly copy the private key. I'd suggest that the `booster-bitswap init` command outputs the peer ID (which is derived from the public key). Then the user can manually set the peer ID on the `boostd` node.

My concern is that if there is a booster-bitswap process on a diffferent node that can silently update a config value on the boostd node that might lead to confusing situations for the user. I think it's clearer for them to explicitly update the config values manually. At the moment it's a little tricky to follow the logic in the `configureRepo` method. The suggested change should simplify all that.

I'd suggest that
- `booster-bitswap init` creates a new peer ID, or prints out the existing peer ID if one was already created (take a look at `boost init` that does something similar)
- `booster-bitswap run` fails if `booster-bitswap init` has not yet been run

hannahhoward:
@dirkmc changes made and hopefully ready to go!

dirkmc (approved):

cmd/booster-bitswap/init.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/libp2p/go-libp2p/p2p/muxer/mplex"
+	"github.com/libp2p/go-libp2p/p2p/muxer/yamux"
+	quic "github.com/libp2p/go-libp2p/p2p/transport/quic"
+	"github.com/libp2p/go-libp2p/p2p/transport/tcp"
+	"github.com/urfave/cli/v2"
+)
+
+func configureRepo(ctx context.Context, cfgDir string, createIfNotExist bool) (peer.ID, crypto.PrivKey, error) {
+	if cfgDir == "" {
+		return "", nil, fmt.Errorf("dataDir must be set")
```
dirkmc, line 25: Suggest changing this error message to `FlagRepo.name + " is a required flag"

cmd/booster-bitswap/run.go
```diff
@@ -83,14 +83,25 @@ var runCmd = &cli.Command{
 		remoteStore := remoteblockstore.NewRemoteBlockstore(bapi)
 		// Create the server API
 		port := cctx.Int("port")
-		server := NewBitswapServer(port, remoteStore)
-
+		repoDir := cctx.String(FlagRepo.Name)
+		host, err := setupHost(ctx, repoDir, port)
+		if err != nil {
+			return fmt.Errorf("setting up libp2p host: %w", err)
```
dirkmc, line 89: If the peer key is not found, I think this will output an error like "setting up libp2p host: libp2p.key not found"
I'd suggest we show something more informative, eg
```
if os.IsNotExist(err) {
	return nil, fmt.Errorf("booster-bitswap has not been initialized. Run the booster-bitswap init command.")
}
```

dirkmc:
Nice work! 🙌 

I merged the PR, and left a couple of small UI things to address in a follow-up PR.

node/config/types.go
```diff
@@ -206,6 +206,8 @@ type DealmakingConfig struct {
 	// The time that can elapse before a download is considered stalled (and
 	// another concurrent download is allowed to start).
 	HttpTransferStallTimeout Duration
+
+	BitswapPeerID          string
```
dirkmc, line 210: Let's add a detailed comment about what this config value is. The comment ends up being copied into the config file itself, so it should be easy for users to understand what the value is and know what to set it to.

PR #791, opened 2022-09-12
Start deal execution in order

dirkmc:


dirkmc:
This ensures that deals start in order, but they will still run in parallel, so there's no guarantee of order of operations. Therefore I'm going to close the PR.

PR #796, opened 2022-09-13
booster-bitswap devnet and tracing

nonsense:
Initial tracing setup for `booster-bitswap`.

### DIFF
--- a/cmd/booster-bitswap/run.go
+++ b/cmd/booster-bitswap/run.go
@@ -38,7 +38,7 @@ var runCmd = &cli.Command{
 		},
 		&cli.BoolFlag{
 			Name:  "tracing",
-			Usage: "enables tracing of booster-http calls",
+			Usage: "enables tracing of booster-bitswap calls",
 			Value: false,
 		},
 		&cli.StringFlag{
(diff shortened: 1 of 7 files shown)


cmd/booster-bitswap/run.go
```diff
@@ -35,6 +36,16 @@ var runCmd = &cli.Command{
...
+		&cli.BoolFlag{
+			Name:  "tracing",
+			Usage: "enables tracing of booster-http calls",
+			Value: false,
+		},
+		&cli.StringFlag{
+			Name:  "tracing-endpoint",
+			Usage: "the endpoint for the tracing exporter",
+			Value: "http://tempo:14268/api/traces",
+		},
```
dirkmc, line 48: Would it make sense to only have the "tracing-endpoint" flag? It's implied that tracing is turned on if this flag is there so maybe we don't need the "tracing" flag?

dirkmc (approved):

cmd/booster-bitswap/run.go
nonsense: I find it nice not having to specify the tracing-endpoint when running the devnet, although what you suggest also makes sense.

Issue #772, opened 2022-09-07
Metrics infrastructure (OpenCensus) for Boost


nonsense:
- [x] 1. Setup initial metrics/stats exporter in Boost, which should export metrics/stats to our devnet / monitoring stack.
- [x] 2. Initial success/failure counters and timers for happy- path of HTTP booster-http retrieval
- [x] 3. Automatically provision dashboards for 2. within our monitoring stack
- [x] 4. Export Go Runtime metrics for booster-http
- [ ] 5. Export Go Runtime metrics for boostd

PR #797, opened 2022-09-13
Set up initial infrastructure for Grafana metrics capture with docker monitoring stack

kylehuntsman:
This sets up the initial infrastructure needed to capture boost and booster-http metrics via the docker monitoring stack.
- Modified the metrics exporter to work with any boost process outside of boostd, instead of just for boostd
  - Instantiates an exporter for booster-http to capture request counts and durations
- Dashboard for metrics are saved in `/docker/monitoring/dashboards/exported_dashboard.json`
  - The dashboard panels can still be modified in browser, but when saving you will be prompted to copy the JSON model of the dashboard to your clipboard. PRs will need to be made to modify the dashboard panels.
- Prometheus was told to watch the booster-http process via the `/docker/monitoring/prometheus.yaml`
- Additionally, adds a trace to booster-http's `/piece/<piece-cid>` endpoint

When running both the devnet and the monitoring docker stacks, the command `docker network connect devnet prometheus` must be ran after starting both stacks to ensure prometheus can actually talk to the processes running in the devnet stack.

Updates #772


docker/monitoring/docker-compose.yaml
```diff
@@ -53,12 +53,15 @@ services:
     container_name: grafana
     image: grafana/grafana:main
     volumes:
-      - ./grafana.ini:/etc/grafana/grafana.ini
-      - ./grafana-datasources.yaml:/etc/grafana/provisioning/datasources/datasources.yaml
-    #environment:
+      - ./grafana/config.ini:/etc/grafana/config.ini
```
nonsense, line 56: Does this work? I thought we need to have `/etc/grafana/grafana.ini`.

metrics/exporter.go
```diff
@@ -2,16 +2,20 @@ package metrics
 
 import (
 	"net/http"
-	_ "net/http/pprof"
```
nonsense, line 5: Wouldn't removing this stop the exporting of pprof metrics?

nonsense (approved):

docker/monitoring/docker-compose.yaml
kylehuntsman: Yeah, this is fine. I discovered that the only thing that is important is where the file is located, and what file extension it has. As long as those two things are right, Grafana will read the file and determine what it is.

PR #798, opened 2022-09-13
return ipld ErrNotFound from remote blockstore interface

dirkmc:
Return [ipld ErrNotFound](https://github.com/ipfs/go-ipld-format/blob/master/merkledag.go#L15) from GetSize() if the block is not found, because [Bitswap checks for this error](https://github.com/ipfs/go-bitswap/blob/4e3736a773545bf09d5178a2be24043b76560f87/server/internal/decision/blockstoremanager.go#L97) to decide if a block is present.
Note that the RemoteBlockstore calls across an RPC boundary, so the code needs to do string matching to check if the error message matches the ipld ErrNotFound error.

### DIFF
--- a/cmd/booster-bitswap/remoteblockstore/remoteblockstore_test.go
+++ b/cmd/booster-bitswap/remoteblockstore/remoteblockstore_test.go
@@ -48,6 +48,16 @@ func TestNormalizeError(t *testing.T) {
 		expected:      fmt.Errorf("some err: %w", format.ErrNotFound{Cid: dummyCid}),
 		isNotFoundErr: true,
 	}, {
+		name:          "ipld ErrorNotFound no cid with prefix",
+		err:           fmt.Errorf("some err: %w", format.ErrNotFound{}),
+		expected:      fmt.Errorf("some err: %w", format.ErrNotFound{}),
+		isNotFoundErr: true,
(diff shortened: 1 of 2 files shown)


nonsense (approved):

cmd/booster-bitswap/remoteblockstore/remoteblockstore_test.go
```diff
@@ -0,0 +1,73 @@
...
+		err:           fmt.Errorf(format.ErrNotFound{}.Error()),
+		expected:      format.ErrNotFound{},
+		isNotFoundErr: true,
+	}, {
+		name:          "ipld ErrorNotFound with prefix",
+		err:           fmt.Errorf("some err: %w", format.ErrNotFound{Cid: dummyCid}),
+		expected:      fmt.Errorf("some err: %w", format.ErrNotFound{Cid: dummyCid}),
+		isNotFoundErr: true,
+	}, {
+		name:          "stringified ipld ErrorNotFound with prefix",
```
hannahhoward, line 46: should we test ErrNotFound with prefix and no cid?

hannahhoward (approved):
LGTM.

Do we have a plan if the error string ever changes? I guess cross that bridge...

PR #801, opened 2022-09-13
experiment: simplified load balancer

hannahhoward:
# Goals

See how much simpler we can make the load balancer by removing the routing protocol as well as forwarding responses for inbound forwarding requests

# Implementation

- Remove the routing protocol, instead relying on a set peer config.  (map of peer ID -> list of protocol IDs)
   - On peer connect, if peer is service node in the peer config, start listening on peers list of protocols, making the route "active"
   - On peer disconnect, if peer is service node in the peer config, stop listening on peers list of protocols
   - Makes code for both libp2p load balancer and service node much simpler, particularly service node (overall this PR has net 500 less LoC, which includes most complicated concurrency stuff - and that's without updatings tests, which will get shorter)
- Also remove forwarding response for inbound requests
   - not sure how big a win this really is -- just a few deletions as a result


dirkmc (approved):
Nice! Looks like a good simplification 👍

PR #804, opened 2022-09-14
dagstore tracing

nonsense:


nonsense:
We've added tracing to piece directory, and we are moving away from the dagstore, so I am closing this.

PR #810, opened 2022-09-15
scrape_config for lotus-miner on host

nonsense:
I'd like to be able to use the same monitoring setup that we have for the devnet with a production miner, where `lotus-miner`, `lotus` and `boost` are generally running outside docker.

I am on the fence on this PR, regarding whether it belongs to this repo at all, as production miners can have a variety of deployments and `lotus`, `lotus-miner`, `boost` might be on different hosts, and not necessarily on the same host.

At the same time I'd like to take advantage of all the goodies that @kylehuntsman is adding, such as automatically provisioned dashboards, etc. and have prometheus scrape and tempo collect traces.

---

This PR is enabling `prometheus` to scrape an endpoint that is on the local host, i.e. it is applicable for our production mainnet miner.

### DIFF
--- a/docker/monitoring/prometheus.yaml
+++ b/docker/monitoring/prometheus.yaml
@@ -12,3 +12,6 @@ scrape_configs:
   - job_name: 'booster-http'
     static_configs:
       - targets: [ 'booster-http:7777' ]
+  - job_name: 'lotus-miner'
+    static_configs:
+      - targets: [ 'host.docker.internal:8787/debug/metrics' ]
--- a/docker/monitoring/docker-compose.yaml
+++ b/docker/monitoring/docker-compose.yaml
@@ -42,6 +42,8 @@ services:
     command: [ "--config.file=/etc/prometheus.yaml" ]
     volumes:
       - ./prometheus.yaml:/etc/prometheus.yaml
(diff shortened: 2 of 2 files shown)


docker/monitoring/prometheus.yaml
```diff
@@ -11,4 +11,7 @@ scrape_configs:
       - targets: [ 'tempo:3200' ]
   - job_name: 'booster-http'
     static_configs:
-      - targets: [ 'booster-http:7777' ]
\ No newline at end of file
+      - targets: [ 'booster-http:7777' ]
+  - job_name: 'lotus-miner'
+    static_configs:
+      - targets: [ 'host.docker.internal:8787/debug/metrics' ]
```
nonsense, line 17: `lotus-miner` is not necessarily on the local host, and not necessarily located on `:8787`

nonsense:
`lotus-miner` is already instrumented with prometheus metrics -- such as https://github.com/filecoin-project/lotus/blob/4e830a8c313f1eea0a62295f9890bcecf2b372a7/storage/sealer/piece_reader.go#L139 -- and I'd like for us to include dashboards for them as well -- specially those that we can about when working on `boost`.

PR #819, opened 2022-09-19
piecemetastore delete options

LexLuthr:
This will be useful for clean after deal expires. This PR only deals with low level changes to the interface and does not handle piecemeta store clean after deal/sector expires.


LexLuthr:
Closing in favor of https://github.com/filecoin-project/boost/pull/822

PR #822, opened 2022-09-20
Add deal-removal, index-removal and piece-removal

LexLuthr:
This PR adds metadata removal in the piecemeta store. These can be used to clean up the piecemeta store after deal/sector expiration.

This PR only adds the functions to the leveldb side. Some implementations are still left in couchbase side(will need a separate PR).

### DIFF
--- a/extern/boostd-data/client/client.go
+++ b/extern/boostd-data/client/client.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/filecoin-project/boost/cmd/boostd-data/model"
+	"github.com/google/uuid"
 	"github.com/ipfs/go-cid"
 	logger "github.com/ipfs/go-log/v2"
 	"github.com/ipld/go-car/v2/index"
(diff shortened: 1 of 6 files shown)


extern/boostd-data/client/client.go
```diff
@@ -123,3 +124,33 @@ func (s *Store) GetOffset(pieceCid cid.Cid, hash mh.Multihash) (uint64, error) {
 
 	return resp, nil
 }
+
+func (s *Store) RemoveDeal(pieceCid cid.Cid, dealId uuid.UUID) error {
```
nonsense, line 128: @LexLuthr please keep names on the Client and names on the RPC server consistent, otherwise it get confusing quickly.

removeDealForPiece => removeDeal
removeAllMultihashes => removeMultihashes

etc.

extern/boostd-data/couchbase/db.go
```diff
@@ -300,3 +300,76 @@ func has(list []cid.Cid, v cid.Cid) bool {
 	}
 	return false
 }
+
+// RemoveAllRecords
+func (db *DB) RemoveAllRecords(ctx context.Context, cursor uint64) error {
```
nonsense, line 305: This method should not be exported I think, as it is using an internal cursor.

extern/boostd-data/ldb/db.go
```diff
@@ -280,3 +280,61 @@ func has(list []cid.Cid, v cid.Cid) bool {
 	}
 	return false
 }
+
+// RemoveAllRecords
+func (db *DB) RemoveAllRecords(ctx context.Context, cursor uint64) error {
```
nonsense, line 285: This method should not be exported as it is using an internal cursor.

extern/boostd-data/couchbase/db.go
LexLuthr: Fixed

extern/boostd-data/client/client.go
LexLuthr: Fixed. I have made a couple of more changes to tweak record removal logic.

LexLuthr:
Closing in favor of https://github.com/filecoin-project/boost/pull/953

PR #825, opened 2022-09-21
Add block filter via BadBits

hannahhoward:
# Goals

Provide a mechanism for filtering out illegal CID requests in bitswap.

# Implementation

- This implements a lightweight block filter module that fetches the bad bits list and checks bitswap requests against what is in the list.
   - The list downloads and updates every five minutes in the background -- currently, it's only 1meg large
   - The in place block check is O(1) against an in memory copy of the list (currently, again, not a serious memory overhead)
- integrates the block filter into the remote blockstore in booster bitswap

# For discusion (that's why it's a draft)

- Currently, this lives inside of booster bitswap -- should it instead live inside of boost and possibly the block API calls?

- I didn't put optimization around passing If-Not-Modified as request header and checking for a 304 Not Modified response -- this may be prudent as the list updates a lot but not every five minutes

- There's no on disk cache -- this was done for simplicity but does mean you need to fetch at startup. Also, this is a blocking call currently -- maybe that's not ideal?

- What is the right way to handle fetching/parsing errors? Right now it's just log.Errorf

- blocked on #786 (currently that branch is the base for this one)

### DIFF
--- a/cmd/booster-bitswap/blockfilter/blockfilter.go
+++ b/cmd/booster-bitswap/blockfilter/blockfilter.go
@@ -5,7 +5,6 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -54,7 +53,8 @@ func FetchBadBitsList(ifModifiedSince time.Time) (bool, io.ReadCloser, error) {
 		return false, nil, nil
 	}
 	if response.StatusCode < 200 && response.StatusCode > 299 {
-		return false, nil, errors.New("http status code was not success")
+		bodyText, _ := io.ReadAll(response.Body)
+		return false, nil, fmt.Errorf("expected HTTP success code, got: %s, response body: %s", http.StatusText(response.StatusCode), string(bodyText))
 	}
 	return true, response.Body, nil
 }
@@ -220,7 +220,7 @@ func (bf *BlockFilter) run(cachedCopy bool) {
 	if cachedCopy {
 		bf.updateDenyList()
 	}
-	updater := bf.clock.Timer(UpdateInterval)
+	updater := bf.clock.Ticker(UpdateInterval)
 	// call the callback if set
 	if bf.onTimerSet != nil {
 		bf.onTimerSet()
@@ -232,9 +232,6 @@ func (bf *BlockFilter) run(cachedCopy bool) {
 		case <-updater.C:
 			// when timer expires, update deny list
 			bf.updateDenyList()
-
-			// then reset the timer
-			updater.Reset(UpdateInterval)
 			// call the callback if set
 			if bf.onTimerSet != nil {
 				bf.onTimerSet()
--- a/cmd/booster-bitswap/blockfilter/blockfilter.go
+++ b/cmd/booster-bitswap/blockfilter/blockfilter.go
@@ -1,0 +1,190 @@
+package blockfilter
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/benbjohnson/clock"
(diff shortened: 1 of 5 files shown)


dirkmc:
> Currently, this lives inside of booster bitswap -- should it instead live inside of boost and possibly the block API calls?

Bitswap has a [built-in mechanism for deny-lists](https://github.com/ipfs/go-bitswap/blob/13c89f1ba6c65253a3dfc3b2bafcb1730b320b66/server/internal/decision/engine.go#L223) that we could use.

> I didn't put optimization around passing If-Not-Modified as request header and checking for a 304 Not Modified response -- this may be prudent as the list updates a lot but not every five minutes

I think that's a good idea - we expect the size of the list to grow in future so it would be good to avoid unnecessary downloads

> There's no on disk cache -- this was done for simplicity but does mean you need to fetch at startup. Also, this is a blocking call currently -- maybe that's not ideal?

I'd suggest we do use an on-disk cache. The size of the list is likely to increase in future. I'd suggest that on startup
- if the list hasn't ever been fetched, block on fetching it
- if there is already an on-disk cache, just do a new fetch in the background

> What is the right way to handle fetching/parsing errors? Right now it's just log.Errorf

That sounds right to me - I think it makes sense for the denylist to be best-effort

> blocked on https://github.com/filecoin-project/boost/pull/786 (currently that branch is the base for this one)

Merged ✅

hannahhoward:
@dirkmc 

PR updated:
- uses bitswap peer request block filter feature in bitswap for filtering
- supports if-not-modified when fetching deny list over http
- holds on disk cache of last fetched list

cmd/booster-bitswap/blockfilter/blockfilter.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	response, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return false, nil, err
+	}
+	if response.StatusCode == http.StatusNotModified {
+		return false, nil, nil
+	}
+	if response.StatusCode < 200 && response.StatusCode > 299 {
+		return false, nil, errors.New("http status code was not success")
```
dirkmc, line 57: It could be helpful to include the error code and the error text in the returned error.
I think the error text should be in response.Body.

cmd/booster-bitswap/blockfilter/blockfilter.go
```diff
@@ -0,0 +1,244 @@
...
+// It blocks to confirm at least one synchronous update of the denylist
+func (bf *BlockFilter) Start(ctx context.Context) error {
+	bf.ctx, bf.cancel = context.WithCancel(ctx)
+	// open the cache file if it eixsts
+	cache, err := os.Open(bf.cacheFile)
+	var cachedCopy bool
+	// if the file does not exist, synchronously fetch the list
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return err
```
dirkmc, line 100: It could be helpful to include the context of the error here:
```suggestion
			return fmt.Errorf("fetching badbits list: %w", err)
```

cmd/booster-bitswap/blockfilter/blockfilter.go
```diff
@@ -0,0 +1,244 @@
...
+	bf.filteredHashesLk.Unlock()
+}
+
+// run periodically updates the deny list asynchronously
+func (bf *BlockFilter) run(cachedCopy bool) {
+	// if there was a cached copy, immediately asynchornously fetch an update
+	if cachedCopy {
+		bf.updateDenyList()
+	}
+	updater := bf.clock.Timer(UpdateInterval)
```
dirkmc, line 223: Would it make sense to use [a ticker](https://gobyexample.com/tickers) here instead of a timer? Then you don't need the Reset logic.

dirkmc (approved):
I've approved this PR, with a few suggestions for minor improvements. Nice work! 🙌

hannahhoward:
@dirkmc yes, we can definitely do follow up experiments if the list grows large. One interesting thing to note: currently, the list is only about 1MB large, way smaller than I would have guessed given how long it's been in operation.

Issue #824, opened 2022-09-21
Track retrieval successes and failures


jacobheun:
Retrieval types to track
* [ ] Bitswap
* [ ] HTTP
* [ ] Graphsync

Metrics to Track:
* Successes
* Failures where not found
* Other failures

PR #827, opened 2022-09-22
Add http response code metrics to dashboard

kylehuntsman:
Adds metrics for booster-http response codes counts for both Payload by Cid and Piece by Cid to the monitoring Grafana dashboard. Adds counts for 200, 400, 404, and 500 status codes over 30 second intervals.
  - Payload by CID Status Response Counts
  - Piece by CID Status Response Counts

Updates #824


kylehuntsman:
Updated panels to show counts over a 30 second interval rather than a rate.

nonsense (approved):

PR #831, opened 2022-09-22
remove lotus-test in docker/build target requirements

nonsense:


airenas:
I suggest moving it to `docker/all` for first time users to work:
 
`docker/all: docker/lotus-test .... `

PR #836, opened 2022-09-26
Protocol Proxy cleanup

hannahhoward:
Follow on to #786 , per @dirkmc suggestions

### DIFF
--- a/cmd/booster-bitswap/init.go
+++ b/cmd/booster-bitswap/init.go
@@ -22,7 +22,7 @@ import (
 
 func configureRepo(ctx context.Context, cfgDir string, createIfNotExist bool) (peer.ID, crypto.PrivKey, error) {
 	if cfgDir == "" {
-		return "", nil, fmt.Errorf("dataDir must be set")
+		return "", nil, fmt.Errorf("%s is a required flag", FlagRepo.Name)
 	}
 
 	if err := os.MkdirAll(cfgDir, 0744); err != nil {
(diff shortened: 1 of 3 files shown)


node/config/types.go
```diff
@@ -207,7 +207,14 @@ type DealmakingConfig struct {
...
 	HttpTransferStallTimeout Duration
 
-	BitswapPeerID          string
+	// The peed id used by booster-bitswap. To set, copy the value
+	// printed by running 'booster-bitswap init'. If this value is set,
+	// Boost will:
+	// - listen on bitswap protocols on its own peer id and forward them
+	// to booster bitswap
+	// - advertise bitswap records to the content indexer
+	// - list bitswap in available transports on the retrieval transport protocol
```
rvagg, line 216:
```suggestion
	//   - listen on bitswap protocols on its own peer id and forward them
	//     to booster bitswap
	//   - advertise bitswap records to the content indexer
	//   - list bitswap in available transports on the retrieval transport protocol
```

I know the primary consumer of this is for config file docs and that there's already precedent in there of lists starting at the front, but I wonder if we shouldn't be adapting in here now to the 1.19 godoc rules for properly rendering lists? It'll show up nicely on godocs and also looks a bit clearer in text IMO.

take it or leave it

rvagg (approved):

cmd/booster-bitswap/init.go
```diff
@@ -70,7 +70,7 @@ func loadPeerKey(cfgDir string, createIfNotExists bool) (crypto.PrivKey, error)
 			return nil, err
 		}
 		if os.IsNotExist(err) && !createIfNotExists {
```
dirkmc, line 72: nit:
```suggestion
		if !createIfNotExists {
```

dirkmc (approved):

PR #839, opened 2022-09-27
Announce Bitswap records to indexer

hannahhoward:
# Goals

When booster-bitswap is configured, announce Bitswap as a supported protocol to the indexer

# Implementation

- First, see https://github.com/filecoin-project/go-fil-markets/pull/752 - this updates go-fil-markets to support a custom metadata generation function for a given deal. It also removes treating provider.ErrAlreadyAdvertised as an error when announcing all deals
- In this PR, the indexprovider.Wrapper struct is augmented in the following ways:
   - 1. It's configured to run IndexerAnnounceAllDeals once on startup (to catch config updates)
   - 2. We no longer treat provider.ErrAlreadyAdvertised as an error when announcing all deals.
   - 3. The announcement is augmented on Bitswap if config.DealMaking.BitswapPeerID is set.
- A new constructor function for the legacy storage market provider also configures a custom metadata generator that announces legacy deals on bitswap if config.Dealmaking.BitswapPeerID is set

### DIFF
--- a/indexprovider/wrapper.go
+++ b/indexprovider/wrapper.go
@@ -108,7 +108,12 @@ func (w *Wrapper) IndexerAnnounceAllDeals(ctx context.Context) error {
 	var merr error
 
 	for _, d := range deals {
-		if d.Checkpoint >= dealcheckpoints.IndexedAndAnnounced {
+		// filter out deals that will announce automatically at a later
+		// point in their execution, as well as deals that are not processing at all
+		// (i.e. in an error state or expired)
+		// (note technically this is only one check point state IndexedAndAnnounced but is written so
+		// it will work if we ever introduce additional states between IndexedAndAnnounced & Complete)
+		if d.Checkpoint < dealcheckpoints.IndexedAndAnnounced || d.Checkpoint >= dealcheckpoints.Complete {
 			continue
 		}
 
--- a/indexprovider/wrapper.go
+++ b/indexprovider/wrapper.go
@@ -17,6 +17,7 @@ import (
 	"github.com/filecoin-project/lotus/markets/dagstore"
 	"github.com/filecoin-project/lotus/markets/idxprov"
 
+	"github.com/filecoin-project/boost/node/config"
 	"github.com/filecoin-project/boost/storagemarket/types/dealcheckpoints"
 	"github.com/hashicorp/go-multierror"
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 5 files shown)


indexprovider/wrapper.go
```diff
@@ -35,36 +36,48 @@ var shardRegMarker = ".boost-shard-registration-complete"
...
+			meshCreator:    meshCreator,
+			cfg:            cfg.DAGStore,
+			bitswapEnabled: cfg.Dealmaking.BitswapPeerID != "",
+			enabled:        !isDisabled,
 		}
+		// announce all deals on startup in case of a config change
+		lc.Append(fx.Hook{
+			OnStart: func(ctx context.Context) error {
+				go func() {
+					_ = w.IndexerAnnounceAllDeals(ctx)
```
dirkmc, line 75: Looking at how this function is implemented it looks like it will announce
- all legacy deals that are in a sealing state (but not in an error or completed state)
- all new deals that have not already been announced (and are not in an error or completed state)

Is that what we want to happen here?

Also keep in mind that some providers have up to 100k deals.

hannahhoward: If there is no change to the protocols supported, this is essentially a no-op -- the call to announce will error with ErrAlreadyPublished.

per @willscott at the moment there is no more efficient way to do this.

willscott (approved):
this is okay from the indexer side

indexprovider/wrapper.go
dirkmc: Ok but I don't think this is going to re-announce deals that are not in one of the two categories I mentioned. For example non-legacy deals that have already been announced.

hannahhoward: ok I now have more questions having read the code. I'm writing it up in a new issue. Basically, based on the behavior of the code as it exists, I really can't figure out what IndexerAnnounceAllDeals was intended to do.

The intent of the go-fil-markets code looks like a re-announce, while the code in boost looks like a first time announce, which will include several deals that really shouldn't be announced yet (i.e. haven't even had a data transfer)

Will document in more detail in a seperate issue.

hannahhoward: ok @dirkmc I think I've put this in a reasonable state. There are two new commits --
https://github.com/filecoin-project/go-fil-markets/pull/752/commits/5dd7bc10ed7d3e2e9b67f5dfa807a3e58ef876b2
- this removes StorageDealStaged from the list of states that announce, because StorageDealStaged is the stage where they get announces (vaguely equivalent to dealcheckpoints.AddedPiece in Boost)
- otherwise, I think the go-fil-markets logic is correct --
   - StorageDealActive describes the entire time the deal is on chain, which seems like a reasonable bound keep the announcement alive, and it is already among the states that get reannounced. The actual "completion states" are either errors or after the deal has actually expired (i.e. we're past SectorEndEpoch).
   
 https://github.com/filecoin-project/boost/pull/839/commits/63ba7018a5369e68d26bd15a3a0361aae5cbe614
 - this changes the condition for announce to essentially be only those deals that are at the IndexedAndAnnounced. Technically, it filters out all that are < IndexedAndAnnouced and >= Complete, which is currently the same as != IndexedAndAnnounced but I thought it'd be useful to treat this as a range in case we add something between IndexedAndAnnounced and Complete in the future.

hannahhoward: this addresses most of the issues in #882 though not all of them. It certainly shouldn't make things worse. :)

hannahhoward:
I'm not sure why test-all is failing -- I don't seem to get this failure locally and it seems unrelated.

dirkmc:
LGTM 👍 

@hannahhoward thanks for digging through all this code and figuring out how to fix it the right way 🙏

PR #851, opened 2022-09-30
Show legacy deals link at top of deals page

dirkmc:
Screen Shot 2022-09-30 at 2 41 13 PM

### DIFF
--- a/react/src/Deals.js
+++ b/react/src/Deals.js
@@ -2,7 +2,7 @@
 import {useQuery} from "@apollo/react-hooks";
 import {
     DealsCountQuery,
-    DealsListQuery,
+    DealsListQuery, LegacyDealsCountQuery,
 } from "./gql";
 import moment from "moment";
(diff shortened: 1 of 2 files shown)


react/src/Deals.js
```diff
@@ -132,6 +133,23 @@ function StorageDealsContent(props) {
...
 }
 
+function LegacyDealsLink(props) {
+    const {data} = useQuery(LegacyDealsCountQuery, {
+        pollInterval: 5000,
+        fetchPolicy: 'network-only',
+    })
+
+    if (!data || !data.legacyDealsCount) {
+        return null
```
RobQuistNL, line 143: Since this call can take quite long on hosts with a lot of legacy deals, can we maybe still show it?
If there's no legacy deals, the page will be empty - but otherwise it can take a while to show up.

dirkmc: Fair enough
https://github.com/filecoin-project/boost/pull/854#top

PR #856, opened 2022-10-03
bitswap client

dirkmc:
Fixes [Allow direct connection to booster-bitswap](https://github.com/filecoin-project/boost/issues/863#top)

TODO:
- [x] create temporary on-disk blockstore
- [x] write output to file


cmd/booster-bitswap/client.go
```diff
@@ -0,0 +1,201 @@
...
+
+		rootCidStr := cctx.Args().Get(1)
+		rootCid, err := cid.Parse(rootCidStr)
+		if err != nil {
+			return fmt.Errorf("parsing cid %s: %w", rootCidStr, err)
+		}
+
+		outputCarPath := cctx.Args().Get(2)
+
+		defer profile.Start(profile.TraceProfile, profile.ProfilePath(".")).Stop()
```
nonsense, line 78: Note to remove this before merging.

nonsense (approved):
lgtm

cmd/booster-bitswap/client.go
```diff
@@ -0,0 +1,220 @@
...
+			return fmt.Errorf("parsing cid %s: %w", rootCidStr, err)
+		}
+
+		outputCarPath := cctx.Args().Get(2)
+
+		ctx := lcli.ReqContext(cctx)
+
+		// setup libp2p host
+		log.Infow("generating libp2p key")
+		privKey, _, err := crypto.GenerateECDSAKeyPair(rand.Reader)
```
aschmahmann, line 82: Any reason you went with ECDSA over ED25519, I tend to see the latter as the default in more places.

cmd/booster-bitswap/client.go
```diff
@@ -0,0 +1,220 @@
...
+		if err != nil {
+			return err
+		}
+
+		host, err := libp2p.New(
+			libp2p.Transport(tcp.NewTCPTransport),
+			libp2p.Transport(quic.NewTransport),
```
aschmahmann, line 89: Are these the only transports you're supporting because they're the only ones Boost supports?

cmd/booster-bitswap/client.go
```diff
@@ -0,0 +1,220 @@
...
+		bsClient := client.New(ctx, net, bs, client.WithBlockReceivedNotifier(brn))
+		defer bsClient.Close()
+		net.Start(bsClient)
+
+		// Connect to host
+		connectStart := time.Now()
+		log.Infow("connecting to server", "server", serverAddrInfo.String())
+		err = host.Connect(ctx, *serverAddrInfo)
+		if err != nil {
+			return fmt.Errorf("connecting to %s: %w", serverAddrInfo, err)
```
aschmahmann, line 122: This will likely cause you problems with large files since if the connection gets pruned on the server side it'll never get re-established and your download will stall (e.g. like https://github.com/ipfs/ipget/issues/103).

It's resolvable by trying to keep the connection alive.

aschmahmann (commented):
@dirkmc just a few pass-by comments on the `fetch` command in case you're interested.

Also worth noting that this code has been duplicated in a few places. Might be worth taking the pieces and putting them into a library so not everyone runs into the same issues/concerns although maybe this is different enough that the few hundred lines is no big deal.

Ones I know of:
- https://github.com/ipfs/ipget/
- https://github.com/application-research/estuary/blob/master/cmd/bsget/main.go
   - https://github.com/ipfs-shipyard/vole/ (really just copied from above because I didn't want to deal with Filecoin FFI things and Go being uncooperative)
   
For example, all three implementations (and this one) have the same problem with host disconnects and requiring some code like the one in the linked issue that kubo has, to make larger data transfers more reliable even when the server is busy (maybe some peering/retry code should end up as a utility in go-libp2p 🤷).

dirkmc:
@aschmahmann thanks a lot for the comments 🙏 

This client is just intended as a convenience for testing, not to be used in production. So we kept it simple on purpose, but I agree that using kubo or of the other clients would make more sense in production.

PR #878, opened 2022-10-06
Add prometheus remote write configuration to PL Grafana

kylehuntsman:
Adds a remote write Prometheus configuration to allow for remote writing metrics to a central Prometheus instance.
- Adds `prometheus-remote-write.yaml` configuration file
- Adds host label value of `local` to local `prometheus.yaml` configuration file
- Feature flag for saving Grafana panels to memory
- Updates the Grafana dashboard (a lot of changes have been made)

Before starting prometheus, the `url`, `region`, `access-key` and `secret-key`, and `host` labels in the `docker/monitoring/prometheus-remote-write.yaml` file will need to be updated to use PL Grafana instance and valid credentials for the provider.

The remote write configuration can be used by setting the `PROMETHEUS_CONFIG_FILE` environment variable to `prometheus-remote-write.yaml` upon running `docker compose up -d` on the monitoring stack.

Ex: `PROMETHEUS_CONFIG_FILE=prometheus-remote-write.yaml docker compose up -d`.

### DIFF
--- a/docker/monitoring/prometheus-remote-write.yaml
+++ b/docker/monitoring/prometheus-remote-write.yaml
@@ -7,28 +7,35 @@ scrape_configs:
     static_configs:
       - targets: [ 'localhost:9090' ]
         labels:
-          host: 'local'
+          host: 'unknown'
   - job_name: 'tempo'
     static_configs:
       - targets: [ 'tempo:3200' ]
         labels:
-          host: 'local'
+          host: 'unknown'
   - job_name: 'boost'
     static_configs:
       - targets: [ 'boost:1288' ]
         labels:
-          host: 'local'
+          host: 'unknown'
   - job_name: 'booster-http'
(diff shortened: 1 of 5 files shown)


nonsense:
Rebased on `main` to give this a try.

docker/monitoring/prometheus-remote-write.yaml
```diff
@@ -0,0 +1,41 @@
+global:
+  scrape_interval:     15s
```
nonsense, line 2: @kylehuntsman having read a bit more on prometheus over the last week, it is important that we set out ranges to minimum of `1m`, given that our scrape_interval is `15s`. The recommendation for `range` is 4 times the scrape interval, as prometheus needs at least 2 points to do a calculation, and we have to account for missed scrapes.

nonsense (approved):

Issue #888, opened 2022-10-13
Rename HTTP URL from "payload" to something else


LexLuthr:
Current setup:

http://URL/pieceCID   returns a unpadded car file of the data

http://URL/pieceCID.car   returns a padded car file of the data (whole piece)

http://URL/payloadCID   returns a unpadded car file of the data

http://URL/payloadCID.car   returns a padded car file of the data (whole piece)

The distinction between the URLs is to allow users to download either by specifying the pieceCID or payloadCID whatever is available to them.

But this gets confusing because payload makes you think that we are talking about the unixfs data here.

Perhaps, we can do it like:

http://URL/pieceCID   returns a unpadded car file of the data

http://URL/pieceCID.car   returns a padded car file of the data (whole piece)

http://URL/payloadCID   returns a unixfs data

http://URL/payloadCID.car   returns a unpadded car file of the data

or rename "payload" to something else to confusion between data and car file. "Payload" also gives you incorrect idea that you can retrieve partial DAGs from within the uploaded car files

LexLuthr:
CC: @hannahhoward @jacobheun @LaurenSpiegel @dirkmc @nonsense 
Please comment your thoughts here and possible alternatives.

LaurenSpiegel:
I liked @jacobheun 's suggestion of using query params.  So, for now there would be one path: `http://foo.com/piece`

With query params--
To get car by piece:
`http://foo.com/piece?piececid=bagaSomePieceCID&format=car`
To get piece by piece:
`http://foo.com/piece?piececid=bagaSomePieceCID&format=piece`
To get car by payload cid:
`http://foo.com/piece?payloadcid=bafySomePayloadCid&format=car`
To get piece by payload cid:
`http://foo.com/piece?payloadcid=bafySomePayloadCid&format=piece`

I might have messed up some of the details there so appreciate 👁️ s and other viewpoints.

nonsense:
I like the suggestion with query params, but I suggest we don't use `piece` as part of the URL, as it has a specific meaning in Filecoin... I'd rather we use `http://foo.com/data` as path.

LexLuthr:
As Anton said, piece has a very specific meaning within Fil. This would also be confusing to Fil people. Perhaps description needs to be updated.

My suggestion:

To get padded car (piece) by pieceCID:  
http://foo.com/piece?piececid=bagaSomePieceCID&format=car
To get unpadded car by pieceCID: 
[...]

PR #898, opened 2022-10-17
feat: redesign http query url

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/888

### DIFF
--- a/cmd/booster-http/server.go
+++ b/cmd/booster-http/server.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
@@ -36,6 +37,8 @@ var ErrNotFound = errors.New("not found")
 var lastModified = time.UnixMilli(1)
 
 const carSuffix = ".car"
+const piecePrefix = "piececid"
+const payloadPrefix = "payloadcid"
 
 type HttpServer struct {
 	path          string
@@ -60,10 +63,6 @@ func NewHttpServer(path string, port int, allowIndexing bool, api HttpServerApi)
 	return &HttpServer{path: path, port: port, allowIndexing: allowIndexing, api: api}
 }
 
-func (s *HttpServer) payloadBasePath() string {
-	return s.path + "/payload/"
-}
-
 func (s *HttpServer) pieceBasePath() string {
 	return s.path + "/piece/"
 }
@@ -73,8 +72,7 @@ func (s *HttpServer) Start(ctx context.Context) {
 
 	listenAddr := fmt.Sprintf(":%d", s.port)
 	handler := http.NewServeMux()
-	handler.HandleFunc(s.payloadBasePath(), s.handleByPayloadCid)
-	handler.HandleFunc(s.pieceBasePath(), s.handleByPieceCid)
+	handler.HandleFunc(s.pieceBasePath(), s.handlePieceRequest)
 	handler.HandleFunc("/", s.handleIndex)
 	handler.HandleFunc("/index.html", s.handleIndex)
 	handler.Handle("/metrics", metrics.Exporter("booster_http")) // metrics
@@ -112,7 +110,7 @@ const idxPage = `
           Download a raw piece by payload CID
         </td>
         <td>
-          <a href="/payload/payloadcid">/payload/<payload cid></a>
+          <a href="/piece?payloadcid=bafySomePayloadCid&format=piece" > /payload/<payload cid></a>
         </td>
       </tr>
       <tr>
@@ -120,7 +118,7 @@ const idxPage = `
           Download a CAR file by payload CID
         </td>
         <td>
-          <a href="/payload/payloadcid.car">/payload/<payload cid>.car</a>
+          <a href="/piece?payloadcid=bafySomePayloadCid&format=car" > /payload/<payload cid>.car</a>
         </td>
       </tr>
       <tr>
@@ -128,7 +126,7 @@ const idxPage = `
           Download a raw piece by piece CID
         </td>
         <td>
-          <a href="/piece/piececid">/piece/<piece cid></a>
+          <a href="/piece?piececid=bagaSomePieceCID&format=piece" > /piece/<payload cid></a>
         </td>
       </tr>
       <tr>
@@ -136,7 +134,7 @@ const idxPage = `
           Download a CAR file by piece CID
         </td>
         <td>
-          <a href="/piece/piececid.car">/piece/<piece cid>.car</a>
+          <a href="/piece?piececid=bagaSomePieceCID&format=car" > /piece/<payload cid>.car</a>
         </td>
       </tr>
       </tbody>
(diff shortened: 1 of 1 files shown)


ribasushi:
@dirkmc @LexLuthr please make sure/test that this keeps working as expected with caching in the mix. I.e. someone calls with a range and a `&format=car` and then drop the `format` specifier: will they get different bytes as they should?

cmd/booster-http/server.go
```diff
@@ -150,33 +148,49 @@ func (s *HttpServer) handleIndex(w http.ResponseWriter, r *http.Request) {
...
+func (s *HttpServer) handlePieceRequest(w http.ResponseWriter, r *http.Request) {
+	q, err := url.ParseQuery(r.URL.RawQuery)
+	if err != nil {
+		msg := fmt.Sprintf("parsing query: %s", err.Error())
+		writeError(w, r, http.StatusBadRequest, msg)
+	}
+
+	// Check provided cid and format and redirect the request appropriately
+	if len(q[payloadPrefix]) == 1 && len(q["format"]) == 1 {
+		s.handleByPayloadCid(w, r)
```
dirkmc, line 160: I think it will simplify the code to have all the query parsing in one place.
So I'd suggest that inside handlePieceRequest we get the query parameters, parse cids etc, and then pass those parsed variables to handleByPayloadCid and handleByPieceCid as parameters.

cmd/booster-http/server.go
```diff
@@ -36,6 +37,8 @@ var ErrNotFound = errors.New("not found")
 var lastModified = time.UnixMilli(1)
 
 const carSuffix = ".car"
+const piecePrefix = "piececid"
+const payloadPrefix = "payloadcid"
```
dirkmc, line 41: I'd suggest using camel-case for query parameters, and renaming these variables to make clear they're parameters (not prefixes)
```suggestion
const pieceCidParam = "pieceCid"
const payloadCidParam = "payloadCid"
```

dirkmc (commented):
Looks good, just some minor refactoring suggestions

cmd/booster-http/server.go
```diff
@@ -150,33 +148,49 @@ func (s *HttpServer) handleIndex(w http.ResponseWriter, r *http.Request) {
...
+		msg := fmt.Sprintf("parsing query: %s", err.Error())
+		writeError(w, r, http.StatusBadRequest, msg)
+	}
+
+	// Check provided cid and format and redirect the request appropriately
+	if len(q[payloadPrefix]) == 1 && len(q["format"]) == 1 {
+		s.handleByPayloadCid(w, r)
+	} else if len(q[piecePrefix]) == 1 && len(q["format"]) == 1 {
+		s.handleByPieceCid(w, r)
+	} else {
```
LaurenSpiegel, line 163: if don't specify format, should default to "piece" format

cmd/booster-http/server.go
```diff
@@ -112,31 +110,31 @@ const idxPage = `
           Download a raw piece by payload CID
         </td>
         <td>
-          <a href="/payload/payloadcid">/payload/<payload cid></a>
+          <a href="/piece?payloadcid=bafySomePayloadCid&format=piece" > /payload/<payload cid></a>
```
dirkmc, line 113: Let's update the parameter name to use camel case here too

cmd/booster-http/server.go
```diff
@@ -150,32 +148,55 @@ func (s *HttpServer) handleIndex(w http.ResponseWriter, r *http.Request) {
...
-func (s *HttpServer) handleByPayloadCid(w http.ResponseWriter, r *http.Request) {
-	startTime := time.Now()
-	ctx, span := tracing.Tracer.Start(r.Context(), "http.payload_cid")
-	defer span.End()
+func (s *HttpServer) handlePieceRequest(w http.ResponseWriter, r *http.Request) {
+	q, err := url.ParseQuery(r.URL.RawQuery)
+	if err != nil {
+		msg := fmt.Sprintf("parsing query: %s", err.Error())
+		writeError(w, r, http.StatusBadRequest, msg)
+	}
```
dirkmc, line 156:
```suggestion
		return
	}
```

cmd/booster-http/server.go
```diff
@@ -150,32 +148,55 @@ func (s *HttpServer) handleIndex(w http.ResponseWriter, r *http.Request) {
...
-		msg := fmt.Sprintf("path '%s' is missing payload CID", r.URL.Path)
-		writeError(w, r, http.StatusBadRequest, msg)
-		stats.Record(ctx, metrics.HttpPayloadByCid400ResponseCount.M(1))
-		return
+	if len(q["format"]) == 1 {
+		if q["format"][0] == "car" { // Check if format value is car
+			isCar = true
+		} else if q["format"][0] != "piece" { // Check if format value is not piece
+			writeError(w, r, http.StatusBadRequest, "unsupported format")
+		}
```
dirkmc, line 165:
```suggestion
			return
		}
```

cmd/booster-http/server.go
```diff
@@ -150,32 +148,55 @@ func (s *HttpServer) handleIndex(w http.ResponseWriter, r *http.Request) {
...
+	} else { // Error if more than 1 format value
+		writeError(w, r, http.StatusBadRequest, "unsupported query")
 	}
```
dirkmc, line 168:
```suggestion
		return
	}
```

cmd/booster-http/server.go
```diff
@@ -150,32 +148,55 @@ func (s *HttpServer) handleIndex(w http.ResponseWriter, r *http.Request) {
...
-		stats.Record(ctx, metrics.HttpPayloadByCid400ResponseCount.M(1))
-		return
+	if len(q["format"]) == 1 {
+		if q["format"][0] == "car" { // Check if format value is car
+			isCar = true
+		} else if q["format"][0] != "piece" { // Check if format value is not piece
+			writeError(w, r, http.StatusBadRequest, "unsupported format")
+		}
+	} else { // Error if more than 1 format value
+		writeError(w, r, http.StatusBadRequest, "unsupported query")
```
dirkmc, line 167: Let's try to give the user as much information as possible if they get the parameters wrong:
```suggestion
		writeError(w, r, http.StatusBadRequest, "missing 'format' query parameter")
```

LexLuthr: We support no format query and default to a piece in that case. I have changed the error msg to be more detailed though.

cmd/booster-http/server.go
```diff
@@ -112,31 +110,31 @@ const idxPage = `
           Download a raw piece by payload CID
         </td>
         <td>
-          <a href="/payload/payloadcid">/payload/<payload cid></a>
+          <a href="/piece?payloadCid=bafySomePayloadCid&format=piece" > /payload/<payload cid></a>
```
dirkmc, line 113: Let's also change the link target text for these:
```suggestion
          <a href="/piece?payloadCid=bafySomePayloadCid&format=piece" >/piece?payloadCid=<payload cid></a>
```

cmd/booster-http/server.go
dirkmc: I'm not sure if I'm reading the code correctly, but it looks like it will return an error if the user doesn't specify the `format` query parameter.

If the user doesn't specify the `format` query parameter then `len(q["format])` will be zero here, right?
```
if len(q["format"]) == 1
```

LexLuthr: My bad. Fixing it.

dirkmc (approved):

Issue #895, opened 2022-10-17
Allow retry for offline deals in case of commp mismatch


LexLuthr:
A few places, we should retry.

1. User could make a spelling mistake in file name. ManualRetry seems to be correct choice here after calling API again with correct file name.

https://github.com/filecoin-project/boost/blob/7e555d6cf36bba5d03ccd6ae36d43ddaee10246b/storagemarket/deal_execution.go#L92-L98

2. Allow retry from commp as user could accidentally provide incorrect file. ManualRetry seems to be correct choice here after calling API again with correct file name.

https://github.com/filecoin-project/boost/blob/7e555d6cf36bba5d03ccd6ae36d43ddaee10246b/storagemarket/deal_commp.go#L28-L34

3. We should allow a manual retry for Local commp errors. Possible scenario would be disk getting up filled and IO copy does not work. This would be similar to allow manual retry if Lotus API is down. 

https://github.com/filecoin-project/boost/blob/7e555d6cf36bba5d03ccd6ae36d43ddaee10246b/storagemarket/deal_commp.go#L57-L63

PR #899, opened 2022-10-17
feat: retry offline deal after commp errors

LexLuthr:
fixes https://github.com/filecoin-project/boost/issues/895

### DIFF
--- a/storagemarket/deal_commp.go
+++ b/storagemarket/deal_commp.go
@@ -1,7 +1,6 @@
 package storagemarket
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -16,7 +15,7 @@ import (
 	carv2 "github.com/ipld/go-car/v2"
 )
 
-var errLocalCommpIOFail = errors.New("failed to write to CommP writer")
+var ErrCommpMismatch = fmt.Errorf("commp mismatch")
 
 // Verify that the commp provided in the deal proposal matches commp calculated
 // over the downloaded file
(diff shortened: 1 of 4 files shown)


storagemarket/deal_execution.go
```diff
@@ -91,9 +91,18 @@ func (p *Provider) execDeal(deal *smtypes.ProviderDealState, dh *dealHandler) *d
 		// Read the bytes received from the downloaded / imported file
 		fi, err := os.Stat(deal.InboundFilePath)
 		if err != nil {
-			return &dealMakingError{
-				error: fmt.Errorf("failed to get size of %s '%s': %w", transferType, deal.InboundFilePath, err),
-				retry: smtypes.DealRetryFatal,
+			// Allow manual retry in case user accidentally gave wrong input
+			// like a non-existent file
```
dirkmc, line 95: nit: can we put this comment inside the `if` statement so it's clear that it refers to the case where IsOffline is true

storagemarket/deal_commp.go
```diff
@@ -27,9 +28,17 @@ func (p *Provider) verifyCommP(deal *types.ProviderDealState) *dealMakingError {
...
-		return &dealMakingError{
-			retry: types.DealRetryFatal,
-			error: fmt.Errorf("commP mismatch, expected=%s, actual=%s", clientPieceCid, pieceCid),
+		if deal.IsOffline {
+			// Allow manual retry in case user accidentally gave wrong input
+			return &dealMakingError{
+				retry: types.DealRetryManual,
+				error: fmt.Errorf("commP mismatch, expected=%s, actual=%s", clientPieceCid, pieceCid),
+			}
+		} else {
```
dirkmc, line 37:
```suggestion
```

storagemarket/deal_commp.go
```diff
@@ -56,6 +65,13 @@ func (p *Provider) generatePieceCommitment(filepath string, pieceSize abi.Padded
 		var err error
 		pi, err = GenerateCommP(filepath)
 		if err != nil {
+			// Allow auto retry to cover cases where IO copy fails due to lack of space
+			if strings.Contains(err.Error(), "failed to write to CommP writer") {
```
dirkmc, line 69: Instead of doing string matching on the error string, let's use a custom error.
See for example ErrDealHandlerNotFound
https://github.com/filecoin-project/boost/blob/015b65463ca4499bbbb726a1f1477473bd56d12c/gql/resolver.go#L178

dirkmc:
Could you please also add some cases to TestDealRestartAfterManualRecoverableErrors (or create a new test if that's simpler) to verify that the changes you've made work correctly. The existing tests are in storagemarket/provider_test.go

dirkmc (approved):

PR #900, opened 2022-10-18
Include size with piece dir offset

dirkmc:


extern/boostd-data/couchbase/db.go
```diff
@@ -259,66 +259,101 @@ func (db *DB) AllRecords(ctx context.Context, pieceCid cid.Cid) ([]model.Record,
...
-// GetOffset
-func (db *DB) GetOffset(ctx context.Context, pieceCid cid.Cid, m multihash.Multihash) (uint64, error) {
-	ctx, span := tracing.Tracer.Start(ctx, "db.get_offset")
+// GetOffsetSize
+func (db *DB) GetOffsetSize(ctx context.Context, pieceCid cid.Cid, m multihash.Multihash) (*model.OffsetSize, error) {
+	ctx, span := tracing.Tracer.Start(ctx, "db.get_offset_size")
 	defer span.End()
 
-	var val string
+	var val *offsetSize
```
nonsense, line 340: Is that right? Shouldn't it be `var val offsetSize`?

dirkmc: Thanks, fixed

nonsense (approved):

Issue #910, opened 2022-10-19
Make leveldb / couchbase parameters configurable


dirkmc:

dirkmc:
Fixed by https://github.com/filecoin-project/boost/pull/914

PR #914, opened 2022-10-21
piece directory config

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/910

### DIFF
--- a/extern/boostd-data/ldb/service.go
+++ b/extern/boostd-data/ldb/service.go
@@ -33,9 +33,9 @@ func NewStore(repopath string) *Store {
 	return &Store{repopath: repopath}
 }
 
-func (s *Store) Start(_ context.Context) error {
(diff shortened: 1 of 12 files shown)


extern/boostd-data/ldb/service.go
```diff
@@ -50,13 +50,11 @@ func (s *Store) Start(_ context.Context) error {
...
 	}
 
-	if s.repopath == "" {
-		// used by tests - prepare db with a cursor
-		log.Debug("preparing db with next cursor")
-		s.db.SetNextCursor(context.Background(), 100)
-	}
+	// Prepare db with a cursor
+	log.Debug("preparing db with next cursor")
+	s.db.SetNextCursor(ctx, 100)
```
nonsense, line 55: This should only be done once when we `init` the database, not on every `Start`.

PR #915, opened 2022-10-23
Add wallet sign to boost cli

jsmetana:
A small thing that I think could be useful. I've adapted it from lotus.


dirkmc:
Thanks for the contribution 🙏 

Could you tell us about your use case - what are you using this functionality for in boost?

jsmetana:
We're trying to build a project, in which the users are going to have boost and we want to verify user accounts by their filecoin wallets, the signature can prove ownership of the wallet/account.

And for ex. Slingshot used this as well for address verification.

dirkmc (approved):

Issue #893, opened 2022-10-14
Integrate Piece Directory into booster-bitswap


dirkmc:
booster-bitswap needs to make two calls to the underlying blockstore implementation:
- `GetSize` to get the size of the block (and check if the blockstore has the block)
- `Get` to get the block data

At the moment these calls go from booster-bitswap to boostd to the sealing subsystem.

We should update booster-bitswap to
- Call the Piece Directory to get the block's size and offset (using the Piece Directory Client)
- Fetch the block data directly from the sealing subsystem (instead of going through boostd)

PR #917, opened 2022-10-24
Integrate piece directory into booster-bitswap

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/893

TODO:
- [x] Handle a GetSize request against a piece index that was imported from the dagstore (and does not have size information)


cmd/booster-http/run.go
```diff
@@ -133,37 +119,8 @@ var runCmd = &cli.Command{
 			log.Info("Tracing exporter enabled")
 		}
 
-		maddr, err := storageService.ActorAddress(ctx)
```
dirkmc, line 136: I moved a bunch of code from booster-http into a shared library (shared with booster-bitswap)

cmd/booster-bitswap/init.go
```diff
@@ -24,10 +24,6 @@ func configureRepo(cfgDir string, createIfNotExist bool) (peer.ID, crypto.PrivKe
 		return "", nil, fmt.Errorf("%s is a required flag", FlagRepo.Name)
 	}
 
-	if err := os.MkdirAll(cfgDir, 0744); err != nil {
```
dirkmc, line 27: This change fixes a problem where a directory called ~ could be created in the current working directory

cmd/lib/api.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/filecoin-project/lotus/markets/sectoraccessor"
+	lotus_modules "github.com/filecoin-project/lotus/node/modules"
+	"github.com/filecoin-project/lotus/node/modules/dtypes"
+	"github.com/filecoin-project/lotus/node/repo"
+	"github.com/filecoin-project/lotus/storage/paths"
+	"github.com/filecoin-project/lotus/storage/sealer"
+	logging "github.com/ipfs/go-log/v2"
+)
+
+func GetFullNodeApi(ctx context.Context, ai string, log *logging.ZapEventLogger) (v1api.FullNode, jsonrpc.ClientCloser, error) {
```
dirkmc, line 27: The code in this file is just moved from booster-http so it can be shared with booster-bitswap

extern/boostd-data/couchbase/db.go
```diff
@@ -19,6 +19,10 @@ import (
 
 const bucketName = "piecestore"
 
+// maxCasRetries is the number of times to retry an update operation when
+// there is a cas mismatch
+const maxCasRetries = 10
```
nonsense, line 24: We should instrument a counter with the CAS retries, so that we know how often this gets hit. If we start getting a lot of CAS retries, we need to revise the optimistic locking, and maybe do pessimistic locks.

extern/boostd-data/couchbase/db.go
```diff
@@ -375,6 +420,23 @@ func (db *DB) GetOffsetSize(ctx context.Context, pieceCid cid.Cid, m multihash.M
...
+// Attempt to perform an update operation. If the operation fails due to a
+// cas mismatch, retry several times before giving up.
+func (db *DB) withCasRetry(f func() error) error {
+	var err error
+	for i := 0; i < maxCasRetries; i++ {
+		err = f()
+		if err == nil {
+			return nil
+		}
+		if !errors.Is(err, gocb.ErrCasMismatch) {
```
nonsense, line 432: I suggest as a start we add a log line here indicating that we got a CAS error.

nonsense (approved):

extern/boostd-data/couchbase/db.go
dirkmc: Agreed 👍 

I'm not expecting conflicts very often as the modification code paths are unlikely to be invoked simultaneously by multiple threads but I think it's a good idea to log CAS errors.

I'll add in a follow-up PR as this one is quite big already.

Issue #924, opened 2022-11-02
Clean up old deal logs


dirkmc:
Boost keeps detailed logs for each deal in `<repo>/boost.logs.db`

We should add a process that periodically cleans up old logs:
- The period and maximum log age should be configurable
- Either all logs or no logs should be cleaned up per deal (we shouldn't clean up half the logs for a deal)

PR #929, opened 2022-11-03
Periodic logsDB cleanup based on configured duration

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/924

### DIFF
--- a/db/logs.go
+++ b/db/logs.go
@@ -63,3 +63,14 @@ func (d *LogsDB) Logs(ctx context.Context, dealID uuid.UUID) ([]DealLog, error) 
 
 	return dealLogs, nil
 }
+
+func (d *LogsDB) CleanupLogs(ctx context.Context, duration int) error {
+
+	t := time.Now()
+	td := t.AddDate(0, 0, -1*duration)
+
+	qry := "DELETE from DealLogs WHERE DealUUID IN (SELECT DealUUID FROM DealLogs WHERE CreatedAt < ?)"
+
+	_, err := d.db.ExecContext(ctx, qry, td)
+	return err
+}
--- a/db/logs.go
+++ b/db/logs.go
@@ -64,12 +64,12 @@ func (d *LogsDB) Logs(ctx context.Context, dealID uuid.UUID) ([]DealLog, error) 
 	return dealLogs, nil
 }
 
-func (d *LogsDB) CleanupLogs(ctx context.Context, duration int) error {
+func (d *LogsDB) CleanupLogs(ctx context.Context, daysOld int) error {
 
 	t := time.Now()
-	td := t.AddDate(0, 0, -1*duration)
+	td := t.AddDate(0, 0, -1*daysOld)
 
-	qry := "DELETE from DealLogs WHERE DealUUID IN (SELECT DealUUID FROM DealLogs WHERE CreatedAt < ?)"
+	qry := "DELETE from DealLogs WHERE DealUUID IN (SELECT DISTINCT DealUUID FROM DealLogs WHERE CreatedAt < ?)"
 
 	_, err := d.db.ExecContext(ctx, qry, td)
 	return err
--- a/db/logs_test.go
+++ b/db/logs_test.go
@@ -3,6 +3,7 @@ package db
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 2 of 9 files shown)


db/logs.go
```diff
@@ -63,3 +63,14 @@ func (d *LogsDB) Logs(ctx context.Context, dealID uuid.UUID) ([]DealLog, error)
 
 	return dealLogs, nil
 }
+
+func (d *LogsDB) CleanupLogs(ctx context.Context, duration int) error {
+
+	t := time.Now()
+	td := t.AddDate(0, 0, -1*duration)
+
+	qry := "DELETE from DealLogs WHERE DealUUID IN (SELECT DealUUID FROM DealLogs WHERE CreatedAt < ?)"
```
dirkmc, line 72: Most databases have an upper limit on the number of values that can be in an IN clause. So just in case I'd suggest adding DISTINCT to the sub-select:
```suggestion
	qry := "DELETE from DealLogs WHERE DealUUID IN (SELECT DISTINCT DealUUID FROM DealLogs WHERE CreatedAt < ?)"
```

db/logs.go
```diff
@@ -63,3 +63,14 @@ func (d *LogsDB) Logs(ctx context.Context, dealID uuid.UUID) ([]DealLog, error)
 
 	return dealLogs, nil
 }
+
+func (d *LogsDB) CleanupLogs(ctx context.Context, duration int) error {
```
dirkmc, line 67: nit: suggest renaming the duration parameter to make clear the unit is days:
```suggestion
func (d *LogsDB) CleanupLogs(ctx context.Context, daysOld int) error {
```

db/logs_test.go
```diff
@@ -31,3 +32,32 @@ func TestLogsDB(t *testing.T) {
...
+
+	err = ldb.InsertLog(ctx, &DealLog{DealUUID: deal.DealUuid, CreatedAt: time.Now(), LogLevel: "INFO", LogParams: "params", LogMsg: "Test", Subsystem: "Sub"})
+	req.NoError(err)
+
+	logs, err := ldb.Logs(ctx, deal.DealUuid)
+	req.NoError(err)
+	req.Len(logs, 1)
+
+	// Provide a negative duration to delete everything from today
+	err = ldb.CleanupLogs(ctx, -1)
```
dirkmc, line 57: I'd suggest that you
- insert two logs, one with created at set to more than one day ago, and with created at set to the current time
- call `ldb.CleanupLogs(ctx, 1)`
- verify that only the older log was removed

node/config/def.go
```diff
@@ -107,6 +107,7 @@ func DefaultBoost() *Boost {
 			HttpTransferMaxConcurrentDownloads: 20,
 			HttpTransferStallTimeout:           Duration(5 * time.Minute),
 			HttpTransferStallCheckPeriod:       Duration(30 * time.Second),
+			DealLogDuration:                    0,
```
dirkmc, line 110: By default lets turn on log cleanup and set it to a month:
```suggestion
			DealLogDurationDays:                    30,
```

node/config/types.go
```diff
@@ -215,6 +215,10 @@ type DealmakingConfig struct {
 	// - advertise bitswap records to the content indexer
 	// - list bitswap in available transports on the retrieval transport protocol
 	BitswapPeerID string
+
+	// The deal logs older than DealLogDuration days are deleted from the logsDB
+	// to keep the size of logsDB in check. Set the value as "0" to disable log cleanup
+	DealLogDuration int
```
dirkmc, line 221:
```suggestion

	// The deal logs older than DealLogDurationDays are deleted from the logsDB
	// to keep the size of logsDB in check. Set the value as "0" to disable log cleanup
	DealLogDurationDays int
```

storagemarket/logs/log.go
```diff
@@ -82,3 +82,27 @@ func paramsWithDealID(dealId uuid.UUID, kvs ...interface{}) []interface{} {
 	kvs = append([]interface{}{"id", dealId}, kvs...)
 	return kvs
 }
+
+func (d *DealLogger) LogCleanup(ctx context.Context, logDuration int) {
+
+	now := time.Now().Truncate(time.Hour)
+
+	// Create a ticker with a one second tick
```
dirkmc, line 90:
```suggestion
	// Create a ticker with a one hour tick
```

storagemarket/logs/log.go
```diff
@@ -82,3 +82,27 @@ func paramsWithDealID(dealId uuid.UUID, kvs ...interface{}) []interface{} {
...
+	ticker := time.NewTicker(time.Hour)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			now = now.Add(time.Hour)
+			d.logger.Debugf("Cleaning logs older than %d from logsDB ", logDuration)
```
dirkmc, line 98: We can log this at the info level as it's only once per hour
```suggestion
		case <-ticker.C:
			d.logger.Infof("Cleaning logs older than %d days from logsDB ", logDuration)
```

storagemarket/logs/log.go
```diff
@@ -82,3 +82,27 @@ func paramsWithDealID(dealId uuid.UUID, kvs ...interface{}) []interface{} {
 	kvs = append([]interface{}{"id", dealId}, kvs...)
 	return kvs
 }
+
+func (d *DealLogger) LogCleanup(ctx context.Context, logDuration int) {
+
+	now := time.Now().Truncate(time.Hour)
```
dirkmc, line 88: Seems like the `now` variable is not being used
```suggestion
```

dirkmc (commented):
A few suggestions, but overall looks great 👍

dirkmc (approved):

Issue #830, opened 2022-09-22
booster-http: add transparent gzip support


dirkmc:
https://pkg.go.dev/github.com/NYTimes/gziphandler#readme-usage

See discussion at https://github.com/filecoin-project/boost/discussions/826

PR #932, opened 2022-11-04
feat: add transparent gzip support for booster-http response

LexLuthr:
resolves https://github.com/filecoin-project/boost/issues/830


cmd/booster-http/http_test.go
```diff
@@ -0,0 +1,84 @@
...
+		BaseContext: func(listener net.Listener) context.Context {
+			return httpServer.ctx
+		},
+	}
+
+	go func() {
+		if err := httpServer.server.ListenAndServe(); err != http.ErrServerClosed {
+			log.Fatalf("http.ListenAndServe(): %w", err)
+		}
+	}()
```
dirkmc, line 58: The idea of using the mock is that you shouldn't need to copy all the startup code. You can just call httpServer.Start()

Then in the mock, add an EXPECT() for the UnsealSectorAt method that returns a reader over a file, similar to the code here:
https://github.com/filecoin-project/boost/pull/573/files#diff-45d2a3a8bb169fc4f527befb3677309c5bcb522cc8830a8bd88344714b9793d9R96

You may also need to mock out a couple of other calls to make everything work.

cmd/booster-http/server.go
```diff
@@ -302,27 +305,45 @@ func serveContent(w http.ResponseWriter, r *http.Request, content io.ReadSeeker,
...
+		w.Header().Set("Content-Encoding", "gzip")
+		if r.Method == "HEAD" {
+			// For an HTTP HEAD request ServeContent doesn't send any data (just headers)
+			http.ServeContent(w, r, "", time.Time{}, content)
+			alog("%s\tHEAD %s", color.New(color.FgGreen).Sprintf("%d", http.StatusOK), r.URL)
+			return
+		}
+
+		// Send the content
+		http.ServeContent(gzipwriter, r, "", lastModified, content)
```
dirkmc, line 335: This block of code is the same in both parts of the if / else statement.
I'd suggest instead doing something like:
```
var content http.ResponseWriter
if strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
    ...
    content = gziphandler....
}
if r.Method == "HEAD" {
    // For an HTTP HEAD request ServeContent doesn't send any data (just headers)
    http.ServeContent(w, r, "", time.Time{}, content)
    alog("%s\tHEAD %s", color.New(color.FgGreen).Sprintf("%d", http.StatusOK), r.URL)
    return
}

// Send the content
http.ServeContent(content, r, "", lastModified, content)
```

dirkmc (commented):
Looking good so far..

dirkmc:
Looks great 👍 

I'd suggest one more change: instead of generating a random file, create a fixtures directory for the test with
- a raw file
- the raw file gzipped

Then in the test
- return a reader over the raw file from the mock
- read the bytes from response.Body
- confirm that the bytes are the same as the gzipped file

dirkmc (approved):

PR #933, opened 2022-11-04
Couchbase record sharding

dirkmc:
The size of the multihash -> offset/size map is sometimes too large to fit in a couchbase value (max size 20mb)

So this PR splits the map into shards, and adds a shard prefix to the couchbase key.

### DIFF
--- a/extern/boostd-data/couchbase/db.go
+++ b/extern/boostd-data/couchbase/db.go
@@ -8,13 +8,13 @@ import (
 	"strconv"
 	"time"
 
-	"golang.org/x/sync/errgroup"
-
 	"github.com/couchbase/gocb/v2"
 	"github.com/filecoin-project/boost/tracing"
 	"github.com/filecoin-project/boostd-data/model"
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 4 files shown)


nonsense (approved):

extern/boostd-data/couchbase/db.go
```diff
@@ -310,136 +320,229 @@ func (db *DB) AddDealForPiece(ctx context.Context, pieceCid cid.Cid, dealInfo mo
...
 	ctx, span := tracing.Tracer.Start(ctx, "db.add_index_records")
 	defer span.End()
 
-	mhToOffsetSize := make(map[string]offsetSize)
-	for _, rec := range recs {
-		mhToOffsetSize[rec.Cid.Hash().String()] = offsetSize{
-			Offset: fmt.Sprintf("%d", rec.Offset),
-			Size:   fmt.Sprintf("%d", rec.Size),
-		}
+	if len(recs) > maxRecsPerPiece {
```
nonsense, line 499: We should be logging / emitting the `recs` value, so that we have an idea of how large various indices are.

PR #946, opened 2022-11-14
feat: HTTP retrieval by block CID

LexLuthr:


willscott:
Where is the logic here that would prevent serving blocks that should not be retrievable by provider policy?

cmd/booster-http/server.go
```diff
@@ -66,12 +69,17 @@ func (s *HttpServer) pieceBasePath() string {
 	return s.path + "/piece/"
 }
 
+func (s *HttpServer) blockBasePath() string {
+	return s.path + "/block/"
```
willscott, line 73: lets have a spec for this interface?

In particular, there are a set of semantics that have become fairly precise around how the IPFS gateways serve a block over HTTP.

We should, unless we have very strong reason otherwise, attempt to replicate those semantics.

Otherwise, we end up having to maintain support for this alternative endpoint with slightly different semantics. that we haven't fully documented anywhere.

dirkmc: I agree we should replicate the gateway semantics - could you point us to any existing docs about gateway semantics that we can take a look at?

dirkmc:
We need to implement provider policy for booster-http, but that's true of all endpoints, so I think it's outside the scope of this PR

cmd/booster-http/server.go
willscott: https://github.com/ipfs/specs/blob/main/http-gateways/TRUSTLESS_GATEWAY.md

dirkmc: So it seems like what's missing is:
1. Set Content-Type header to `application/vnd.ipld.raw`
2. Set Content-Disposition header to `attachment; filename="<cid>.car"`

Is there anything else I'm missing?

willscott: we should consider path semantics of `/ipfs/<cid>` rather than `/block/<cid>` as well

dirkmc: Sorry I'm not familiar enough with the spec to know what you mean by path semantics - can you point us to a specific section?

LexLuthr: We probably need "X-Content-Type-Options" set as "nosniff" as well

willscott: > data is requested by CID, only supported path is /ipfs/{cid}

just that the implemented handler is at the path "/block/" and the gateway spec is expecting it to be at "/ipfs/"

dirkmc: Oh I see - sure that seems like an easy change to make

LexLuthr: All requested changes done. @willscott and @dirkmc Can you review again?

cmd/booster-http/server.go
```diff
@@ -194,6 +210,57 @@ func (s *HttpServer) handlePieceRequest(w http.ResponseWriter, r *http.Request)
...
+	}
+
+	filename := r.URL.Query().Get("filename")
+	if filename == "" {
+		filename = blockCidStr
+	}
+
+	utf8Name := url.PathEscape(filename)
+	asciiName := url.PathEscape(regexp.MustCompile("[[:^ascii:]]").ReplaceAllLiteralString(filename, "_"))
+	w.Header().Set("Content-Disposition", fmt.Sprintf("%s; filename=\"%s\"; filename*=UTF-8''%s", "attachment", asciiName, utf8Name))
```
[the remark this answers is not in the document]
LexLuthr: Are talking about the filename logic in case use did not specify one? Or filename within the header itself?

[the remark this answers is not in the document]
LexLuthr: I see what you mean. Let me remove the filename input altogether to avoid this. Filename would be same as block Cid to mitigate this.

cmd/booster-http/server.go
```diff
@@ -194,6 +210,57 @@ func (s *HttpServer) handlePieceRequest(w http.ResponseWriter, r *http.Request)
...
+	w.Header().Set("Content-Type", "application/vnd.ipld.raw")
```
dirkmc, line 249: It looks like the Content-Type gets overridden in the parameter passed to `serveContent()`

LexLuthr: Fixed

cmd/booster-http/server.go
```diff
@@ -65,12 +68,17 @@ func (s *HttpServer) pieceBasePath() string {
 	return s.path + "/piece/"
 }
 
+func (s *HttpServer) blockBasePath() string {
+	return s.path + "/ipfs/"
```
LaurenSpiegel, line 72: why ipfs?

LaurenSpiegel:
Could you please do a related PR to the docs so we can see how this appears from a user perspective along with the piece cid endpoints and query params?

cmd/booster-http/server.go
```diff
@@ -194,6 +210,51 @@ func (s *HttpServer) handlePieceRequest(w http.ResponseWriter, r *http.Request)
...
+	data, err := s.api.GetBlockByCid(ctx, blockCid)
+	if err != nil {
+		msg := fmt.Sprintf("server error getting data for block '%s': %s", blockCidStr, err.Error())
+		writeError(w, r, http.StatusInternalServerError, msg)
+		stats.Record(ctx, metrics.HttpBlockByCid500ResponseCount.M(1))
+		return
+	}
+
+	utf8Name := url.PathEscape(blockCidStr)
+	asciiName := url.PathEscape(regexp.MustCompile("[[:^ascii:]]").ReplaceAllLiteralString(blockCidStr, "_"))
```
dirkmc, line 247: Instead of doing all the escaping, can you just use blockCid.String()?

cmd/booster-http/http_test.go
```diff
@@ -96,3 +97,30 @@ func TestHttpGzipResponse(t *testing.T) {
...
+	data := []byte("Hello World!")
+
+	mockHttpServer.EXPECT().GetBlockByCid(gomock.Any(), gomock.Any()).AnyTimes().Return(data, nil)
+
+	// Create a client and make request with Encoding header
+	client := new(http.Client)
+	request, err := http.NewRequest("GET", "http://localhost:7777/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi", nil)
+	require.NoError(t, err)
+
+	response, err := client.Do(request)
```
dirkmc, line 117: Let's also check that the expected response headers are sent and that they are correct

cmd/booster-http/server.go
LexLuthr: I might be wrong but I think this due to how Go treats a string as slice of bytes(runes). It would still try to use the special characters so forcing them to escape would be a safer choice.
BTW, I picked this directly from Kubo.
https://github.com/ipfs/kubo/blob/1d5e46ac97dbd9862818b191517a4dc04186a26e/core/corehttp/gateway_handler.go#L688-L692

dirkmc (approved):

dirkmc:
@LexLuthr some other PRs got merged into `nonsense/piece-store`. Could you please resolve the conflicts and merge when ready.

PR #953, opened 2022-11-15
feat: piecedirectory cleanup

LexLuthr:


LexLuthr:
We probably will need to refactor the couchbase tested so they can succeed on circleCI. Even if I enable remote_docker in circleCI, the tests keep failing. One, because multiple tests require to spin up couchbase and thus required ports get blocked. Other reason is a bit weird but I think sleep is not being honored so couchbase connection starts before services are setup correctly. 
Seems to be working locally.

extern/boostd-data/ldb/service.go
```diff
@@ -334,6 +335,98 @@ func (s *Store) IndexedAt(ctx context.Context, pieceCid cid.Cid) (time.Time, err
...
+
+	defer func(now time.Time) {
+		log.Debugw("handled.remove-deal-for-piece", "took", time.Since(now).String())
+	}(time.Now())
+
+	s.Lock()
+	defer s.Unlock()
+
+	md, err := s.db.GetPieceCidToMetadata(ctx, pieceCid)
+	if err != nil {
```
dirkmc, line 353: Let's return nil if the error is a not found error

extern/boostd-data/svc/svc_test.go
```diff
@@ -46,27 +46,26 @@ var testCouchSettings = couchbase.DBSettings{
 }
 
 func TestService(t *testing.T) {
-	logging.SetLogLevel("*", "debug")
+	_ = logging.SetLogLevel("*", "debug")
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
dirkmc, line 51: Do these tests take 5 minutes?

extern/boostd-data/svc/svc_test.go
```diff
@@ -340,14 +344,131 @@ func compareIndices(subject, subjectDb index.Index) (bool, error) {
...
+		CarLength:   3,
+	}
+
+	// Add a deal for the piece
+	err = cl.AddDealForPiece(ctx, pieceCid, di)
+	require.NoError(t, err)
+
+	// Add the same deal a second time to test uniqueness
+	err = cl.AddDealForPiece(ctx, pieceCid, di)
+	require.NoError(t, err)
```
dirkmc, line 424: This is already tested elsewhere, let's try to keep this test to the minimum required to test the remove functionality

extern/boostd-data/client/client.go
```diff
@@ -27,6 +28,9 @@ type Store struct {
 		MarkIndexErrored          func(context.Context, cid.Cid, error) error
 		IndexedAt                 func(context.Context, cid.Cid) (time.Time, error)
 		PiecesContainingMultihash func(context.Context, mh.Multihash) ([]cid.Cid, error)
+		RemoveDealForPiece        func(context.Context, cid.Cid, uuid.UUID) error
+		RemovePieceMetadata       func(context.Context, cid.Cid) error
```
dirkmc, line 32: I'm not sure that we need the RemovePieceMetadata method - normally we should only ever need to call RemoveDealForPiece.
RemoveIndexes should only be called if something went wrong and the database got into an inconsistent state where we have indexes but not piece information, right?

extern/boostd-data/couchbase/db.go
```diff
@@ -553,6 +554,100 @@ func (db *DB) AddIndexRecords(ctx context.Context, pieceCid cid.Cid, recs []mode
...
+			err = getResult.Content(&cidStrs)
+			if err != nil {
+				return err
+			}
+			if len(cidStrs) <= 1 {
+				_, err = db.mhToPieces.Remove(mhStr, &gocb.RemoveOptions{Context: ctx})
+				if err != nil {
+					return err
+				}
+			}
```
dirkmc, line 624: I think we should still check to make sure that the piece cid in the array matches the piece cid we're trying to remove. So basically just remove these lines (as we do that check below)
```suggestion
```

extern/boostd-data/couchbase/db.go
```diff
@@ -553,6 +554,100 @@ func (db *DB) AddIndexRecords(ctx context.Context, pieceCid cid.Cid, recs []mode
...
+					cidStrs = cidStrs[:len(cidStrs)-1]
+				}
+			}
+
+			_, err = db.mhToPieces.Upsert(cbKey, cidStrs, &gocb.UpsertOptions{Context: ctx})
+			if err != nil {
+				return fmt.Errorf("removing multihash %s to piece %s: update doc: %w", mhStr, pieceCid, err)
+			}
+		}
+		// Add the piece cid to the set of piece cids
```
dirkmc, line 640:
```suggestion
```

dirkmc (commented):
This looks like a good start 👍 

There's a lot going on in this PR so I'd suggest we break this PR into smaller PRs and review them one-by-one:
1. Firstly let's create a PR that just does the remove operations
2. Let's create a separate PR branched from `nonsense/piece-store` that gets the couchbase tests working with docker. Note that we should try to use docker's go API rather than executing command-line commands from a shell.
3. Let's create a separate PR branched from `nonsense/piece-store` that has unrelated fixes (eg fixing the span context)
4. Let's move the GC implementation to another PR after the first PR has been merged.
5. Let's minimize changes to go.mod in each PR so it's clear why each dependency is changing.

extern/boostd-data/svc/svc_test.go
LexLuthr: CouchBase tests are taking longer. Not 5 minutes though. Since, I am removing all CB tests from this PR, I will reset all contexts to 5 Seconds.

extern/boostd-data/client/client.go
LexLuthr: I kept all 3 removal method to cover edge cases where one or more DB operations fail.
Example:
Remove last deal -> remove metadata(fails) -> remove indexes (succeeds)
So now we can remove metadata as indexes are already gone but metadata itself was left.

LexLuthr:
- I have "undo" most of the extra changes and this PR not focuses in cleanup API only. 
- All suggested changes were implemented. 

- https://github.com/filecoin-project/boost/pull/968 created for CB testing.

- I will create a new PR for unrelated fixes once this merged so there are no conflicts.

- I am holding off on creating the GC PR till I have clarity on deal UUID vs legacy deal ID as GC needs to remove both legacy and new deals. There are a couple of other points which needs clarification.

@dirkmc Can you review again?

extern/boostd-data/couchbase/service.go
```diff
@@ -300,3 +301,96 @@ func normalizeMultihashError(m mh.Multihash, err error) error {
 	}
 	return err
 }
+
+// Remove Single deal for pieceCID. If []Deals is empty then Metadata is removed as well
+func (s *Store) RemoveDealForPiece(ctx context.Context, pieceCid cid.Cid, dealUuid uuid.UUID) error {
+	log.Debugw("handle.remove-deal-for-piece", "piece-cid", pieceCid)
```
nonsense, line 307:
```suggestion
	log.Debugw("handle.remove-deal-for-piece", "piece-cid", pieceCid, "deal-uuid", dealUuid)
```

extern/boostd-data/ldb/service.go
```diff
@@ -371,3 +372,98 @@ func normalizeMultihashError(m mh.Multihash, err error) error {
 	}
 	return err
 }
+
+// Remove Single deal for pieceCID. If []Deals is empty then Metadata is removed as well
+func (s *Store) RemoveDealForPiece(ctx context.Context, pieceCid cid.Cid, dealUuid uuid.UUID) error {
+	log.Debugw("handle.remove-deal-for-piece", "piece-cid", pieceCid)
```
nonsense, line 378:
```suggestion
	log.Debugw("handle.remove-deal-for-piece", "piece-cid", pieceCid, "deal-uuid", dealUuid)
```

nonsense (approved):

extern/boostd-data/couchbase/db.go
```diff
@@ -665,3 +665,93 @@ func serviceName(svc gocb.ServiceType) string {
...
+					cidStrs[i] = cidStrs[len(cidStrs)-1]
+					cidStrs[len(cidStrs)-1] = ""
+					cidStrs = cidStrs[:len(cidStrs)-1]
+				}
+			}
+
+			_, err = db.mhToPieces.Replace(cbKey, cidStrs, &gocb.ReplaceOptions{Context: ctx})
+			if err != nil {
+				return fmt.Errorf("removing multihash %s to piece %s: update doc: %w", mhStr, pieceCid, err)
+			}
```
dirkmc, line 747: This block needs to be wrapped in a compare-and-swap. See AddDealForPiece for an example.
You can read more about compare-and-swap at https://docs.couchbase.com/go-sdk/current/howtos/concurrent-document-mutations.html

extern/boostd-data/couchbase/service.go
```diff
@@ -300,3 +301,96 @@ func normalizeMultihashError(m mh.Multihash, err error) error {
...
+			fmt.Errorf("Failed to remove the Metadata after removing the last deal: %w", err)
+		}
+		return nil
+	}
+
+	err = s.db.SetPieceCidToMetadata(ctx, pieceCid, md)
+	if err != nil {
+		return err
+	}
+
```
dirkmc, line 344: I'd suggest moving this code to db.go and putting it inside a compare-and-swap

extern/boostd-data/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/couchbase/gocb/v2 v2.5.0
-	github.com/docker/docker v20.10.7+incompatible
+	github.com/docker/docker v20.10.17+incompatible
```
dirkmc, line 7: There are a bunch of changes to versions in go.mod and go.sum. Let's not change any versions unless we need to.

extern/boostd-data/svc/svc_test.go
```diff
@@ -340,14 +346,129 @@ func compareIndices(subject, subjectDb index.Index) (bool, error) {
...
+
+	_, err = cl.GetPieceDeals(ctx, pieceCid)
+	require.ErrorContains(t, err, "not found")
+
+	_, err = cl.GetOffsetSize(ctx, pieceCid, mhash)
+	require.ErrorContains(t, err, "not found")
+
+	_, err = cl.GetRecords(ctx, pieceCid)
+	require.ErrorContains(t, err, "not found")
```
dirkmc, line 473: Let's also add checks that PiecesContainingMultihash and IsIndexed work as expected

extern/boostd-data/go.mod
LexLuthr: I have changed the dependency back but require{} are automatically updated by go mod tidy as per requirements.

extern/boostd-data/couchbase/db.go
```diff
@@ -665,3 +666,155 @@ func serviceName(svc gocb.ServiceType) string {
...
+	ctx, span := tracing.Tracer.Start(ctx, "db.remove_metadata")
+	defer span.End()
+
+	return db.withCasRetry("remove-metadata-for-piece", func() error {
+		var getResult *gocb.GetResult
+		k := toCouchKey(pieceCid.String())
+		getResult, err := db.pcidToMeta.Get(k, &gocb.GetOptions{Context: ctx})
+		if err != nil {
+			return fmt.Errorf("getting piece cid to metadata for piece %s: %w", pieceCid, err)
+		}
```
dirkmc, line 681: Here we should check if the error is a not found error. If so, we should just return nil.

extern/boostd-data/couchbase/db.go
```diff
@@ -665,3 +666,155 @@ func serviceName(svc gocb.ServiceType) string {
...
+
+		// Get each value in the map
+		for mhStr := range recMap {
+			cbKey := toCouchKey(mhStr)
+			err = db.withCasRetry("remove-piece", func() error {
+				getResult, err := db.mhToPieces.Get(cbKey, &gocb.GetOptions{Context: ctx})
+				if err != nil {
+					return nil
+				}
```
dirkmc, line 738: If the key is not found, let's assume another process removed it, and continue (skip this key)

extern/boostd-data/couchbase/db.go
```diff
@@ -665,3 +666,155 @@ func serviceName(svc gocb.ServiceType) string {
...
+	ctx, span := tracing.Tracer.Start(ctx, "db.remove_deals")
+	defer span.End()
+
+	return db.withCasRetry("remove-deal-for-piece", func() error {
+		var getResult *gocb.GetResult
+		k := toCouchKey(pieceCid.String())
+		getResult, err := db.pcidToMeta.Get(k, &gocb.GetOptions{Context: ctx})
+		if err != nil {
+			return fmt.Errorf("getting piece cid to metadata for piece %s: %w", pieceCid, err)
+		}
```
dirkmc, line 787: If the err is a not found error, let's just return nil

extern/boostd-data/couchbase/service.go
dirkmc: Marking as unresolved as it doesn't look like the logic has changed

extern/boostd-data/couchbase/db.go
LexLuthr: This is wrapped in a cas retry func() so, putting a continue statement won't work. I can return the original error here and check the error outside the func() and then continue without calling the remove. WDYT?

dirkmc: I think it makes more sense to check if the error is a not found error, and in that case return nil
Then withCasRetry will also return nil.
We should also check the error returned from db.pieceOffsets.Remove and if it's a not found error, ignore it.

extern/boostd-data/couchbase/db.go
```diff
@@ -784,11 +786,14 @@ func (db *DB) RemoveIndexes(ctx context.Context, pieceCid cid.Cid, recordCount i
...
 				if err != nil {
-					return nil
+					return err
 				}
 				var cidStrs []string
 				err = getResult.Content(&cidStrs)
 				if err != nil {
+					if isNotFoundErr(err) {
+						return nil
+					}
```
dirkmc, line 796: I think this check should actually be on line 789

extern/boostd-data/couchbase/db.go
```diff
@@ -715,3 +715,164 @@ func serviceName(svc gocb.ServiceType) string {
...
+					return nil
+				}
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (db *DB) RemoveDeals(ctx context.Context, dealId string, pieceCid cid.Cid) error {
```
dirkmc, line 835: This method should only remove one deal right? If so let's call it RemoveDeal

extern/boostd-data/ldb/service.go
```diff
@@ -384,3 +384,102 @@ func normalizeMultihashError(m mh.Multihash, err error) error {
...
+	defer span.End()
+
+	defer func(now time.Time) {
+		log.Debugw("handled.remove-piece-metadata", "took", time.Since(now).String())
+	}(time.Now())
+
+	s.Lock()
+	defer s.Unlock()
+
+	if err := s.db.RemoveMetadata(ctx, pieceCid); err != nil {
```
dirkmc, line 448: Let's rename the db method to RemovePieceMetadata to be consistent with the name in the service

piecedirectory/piecedirectory.go
```diff
@@ -325,25 +327,16 @@ func (ps *PieceDirectory) BuildIndexForPiece(ctx context.Context, pieceCid cid.C
...
-
-	//if !wasLast {
-	//return nil
-	//}
-
-	//// Remove piece indexes
-	//if err := ps.deleteIndexForPiece(pieceCid); err != nil {
-	//return fmt.Errorf("deleting index for piece %s: %w", pieceCid, err)
-	//}
+func (ps *PieceDirectory) DeleteDealForPiece(ctx context.Context, pieceCid cid.Cid, dealUuid string) error {
```
dirkmc, line 330: Let's rename DeleteDealForPiece to RemoveDealForPiece to be consistent with the service

extern/boostd-data/couchbase/db.go
dirkmc: This is not resolved

extern/boostd-data/couchbase/db.go
```diff
@@ -715,3 +715,164 @@ func serviceName(svc gocb.ServiceType) string {
...
+			return fmt.Errorf("failed removing index for piece %s: %w", pieceCid, err)
+		}
+
+		_, err = db.pcidToMeta.Remove(k, &gocb.RemoveOptions{
+			Context: ctx,
+			Cas:     getResult.Cas(),
+		})
+		if err != nil {
+			return fmt.Errorf("removing piece %s metadata: %w", pieceCid, err)
+		}
```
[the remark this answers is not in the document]
dirkmc: This is not resolved

dirkmc (approved):

Issue #950, opened 2022-11-14
feature: Output datacap in `boost wallet list`


magik6k:
It would be really neat to see remaining datacap in the `boost wallet list` output.

PR #959, opened 2022-11-16
feat: show datacap in `boost wallet list`

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/950

### DIFF
--- a/cmd/boost/wallet_cmd.go
+++ b/cmd/boost/wallet_cmd.go
@@ -207,6 +207,9 @@ var walletList = &cli.Command{
 					}
 				} else {
 					wallet[dataCapKey] = "n/a"
+					if cctx.Bool("json") {
+						wallet[dataCapKey] = nil
+					}
 				}
 
 				wallets = append(wallets, wallet)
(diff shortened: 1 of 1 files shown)


cmd/boost/wallet_cmd.go
```diff
@@ -198,6 +199,19 @@ var walletList = &cli.Command{
 						wallet[marketLockedKey] = marketLockedValue
 					}
 				}
+				dcap, err := api.StateVerifiedClientStatus(ctx, addr, types.EmptyTSK)
+				if err == nil {
+					wallet[dataCapKey] = dcap
+					if !cctx.Bool("json") && dcap == nil {
+						wallet[dataCapKey] = "X"
```
dirkmc, line 206: Is is possible for dcap to be nil when err is not nil?

LexLuthr: There are 2 scenarios here.
1. Address does exists on chain. This would result in a error for the query.
2. Address is on chain but there is no data cap assigned to that address yet or it is now over.  This scenario could be dcap to be nil when err is not nil.

dirkmc: I guess I would expect that dcap is zero if there is no data cap assigned to that address... but if it's nil then this logic LGTM 👍

dirkmc (approved):

Issue #951, opened 2022-11-15
Support exposing booster bitswap peer directly to indexer via extended providers


hannahhoward:
# What

The indexer now supports "extended providers", a mechanism for adding new peers and protocols to a family of advertisements:
https://github.com/filecoin-project/index-provider/pull/292

Through this, we'd like to enable two new features --
1. SPs should be able to expose their bitswap peer ID directly to the external world if they prefer, rather than going through the boost protocol proxy
2. For those continuing to use the protocol proxy (cause it's perhaps easier, and maybe they don't want to go through the hassle of a complicated configuration), stop re-advertising every deal when bitswap goes on or off.

# Implementation Outline

Currently, we have the following Bitswap configs:

```golang
type DealmakingConfig struct {
        // ...
        
    	// The peed id used by booster-bitswap. To set, copy the value
	// printed by running 'booster-bitswap init'. If this value is set,
	// Boost will:
	// - listen on bitswap protocols on its own peer id and forward them
	// to booster bitswap
	// - advertise bitswap records to the content indexer
	// - list bitswap in available transports on the retrieval transport protocol
	BitswapPeerID string
         
         //...
}
```

I propose adding two additional configs

```golang
type DealmakingConfig struct {
        // ...
        
    	// The peed id used by booster-bitswap. To set, copy the value
	// printed by running 'booster-bitswap init'. If this value is set,
	// Boost will:
	// - listen on bitswap protocols on its own peer id and forward them
	// to booster bitswap
	// - advertise bitswap records to the content indexer
	// - list bitswap in available transports on the retrieval transport protocol
	BitswapPeerID string
         

         // Public multiaddresses for booster-bitswap
         // If empty, booster-bitswap is assumed to be running privately
         // Boost will listen on bitswap protocols on its own peer id and forward them to booster bitswap via the protocol proxy
         // If public addresses are set, Boost will announce the booster bitswap peer id directly to the indexer as an extended provider
         BitswapPublicAddresses []string
         
          // If operating in public mode, in order to announce Booster-Bitswap as an extended provder, this value must point to a
          // a file containing the booster bitswap peer id's private key. Can be left blank when operating with protocol proxy
         BitswapPrivKeyFile string
}
```

With these in place, we will need to:
- delete any per deal bitswap announcements made (the per deal is now always graphsync only)
- advertise the bitswap peer id as an extended provider when operating in public mode
- advertise with no context id to add bitswap as a protocol when operating in proxied mode

[...]

PR #967, opened 2022-11-20
Support exposing bitswap publicly

hannahhoward:
# Goals

Implements #951 

Uses index provider families to switch to advertising bitswap as an extended provider in the indexer-- this allows a provider to expose booster-bitswap directly as a public node.

# Implementation

- unfortunately, this has two dependent PRs for the time being, cause the indexprovider package needed a major update which triggered other deps updates -- https://github.com/filecoin-project/lotus/pull/9688 & https://github.com/filecoin-project/go-fil-markets/pull/766
- per #951 description, adds two config vars for public addresses and the location of a private key for booster bitswap
- if public addresses are not present, proxy approach still works:
   - adds BItswap peer id to config of protocol proxy
   - advertises Bitswap records for the main peer id
   - NOTE: rather than announcing bitswap per deal any more, we announce bitswap as a supported protocol on an extended provider with the same peer ID as boost
- if public address are present, annouces them as extended provider
   - extended provider for bitswap is announced in indexer with peerKey + addrs of the bitswap node
   - booster-bitswap should be started w/o forwarding
   - also makes sure these addresses are advertised instead of the main peer in the support transport endpoint
   
# For discussion

- This PR could probably lose the Lotus update if we moved the index provider code we are depending on over to boost. Why are we relying on this code from Lotus? It seems like it should live in Boost.
- Do we need to bother with trying to unannounce individual bitswap records?


indexprovider/wrapper.go
```diff
@@ -36,55 +40,107 @@ var shardRegMarker = ".boost-shard-registration-complete"
...
+			extendedProviders: eps,
+
+			enabled: !isDisabled,
 		}
 		// announce all deals on startup in case of a config change
 		lc.Append(fx.Hook{
 			OnStart: func(ctx context.Context) error {
 				go func() {
-					_ = w.IndexerAnnounceAllDeals(ctx)
+					_ = w.AnnounceExtendedProviders(ctx)
```
willscott, line 111: this change (from calling `IndexerAnnounceAllDeals` to calling `AnnounceExtendedProviders`) will now just update the multiaddrs / if bitswap is running.
Before, that method on startup would try to reconcile what was announced with the known deals.
where will we do that now?

willscott (commented):
This seems generally reasonable.

@ischasny should also review

go.mod
```diff
@@ -27,33 +27,33 @@ require (
...
 	github.com/filecoin-project/go-jsonrpc v0.1.8
 	github.com/filecoin-project/go-legs v0.4.9
 	github.com/filecoin-project/go-padreader v0.0.1
 	github.com/filecoin-project/go-paramfetch v0.0.4
-	github.com/filecoin-project/go-state-types v0.9.1
+	github.com/filecoin-project/go-state-types v0.9.9
 	github.com/filecoin-project/go-statestore v0.2.0
-	github.com/filecoin-project/index-provider v0.8.1
-	github.com/filecoin-project/lotus v1.18.0-rc3
+	github.com/filecoin-project/index-provider v0.9.2-0.20221119192528-622ddb90df0e
```
ischasny, line 37: [v0.9.2](https://github.com/filecoin-project/index-provider/releases/tag/v0.9.2) has been cut, please use it

indexprovider/wrapper.go
```diff
@@ -36,55 +40,107 @@ var shardRegMarker = ".boost-shard-registration-complete"
...
 	return w.enabled
 }
 
+func (w *Wrapper) AnnounceExtendedProviders(ctx context.Context) error {
+
+	if len(w.extendedProviders) == 0 {
+		return nil
+	}
+	adBuilder := xproviders.NewAdBuilder(w.h.ID(), w.h.Peerstore().PrivKey(w.h.ID()), w.h.Addrs())
+	adBuilder.WithExtendedProviders(w.extendedProviders...)
```
ischasny, line 130: If you want a bitswap endpoint to also be added for the main provider, don't forget to call `.WithMetadata(...)` on the builder.

hannahhoward:
moved the indexer code cause it was the right thing to do, but still not able to avoid a Lotus deps update

node/config/types.go
```diff
@@ -218,6 +218,16 @@ type DealmakingConfig struct {
 	// - list bitswap in available transports on the retrieval transport protocol
 	BitswapPeerID string
 
+	// Public multiaddresses for booster-bitswap
+	// If empty, booster-bitswap is assumed to be running privately
+	// Boost will listen on bitswap protocols on its own peer id and forward them to booster bitswap via the protocol proxy
+	// If public addresses are set, Boost will announce the booster bitswap peer id directly to the indexer as an extended provider
```
dirkmc, line 224:
```suggestion
	// Public multiaddresses for booster-bitswap.
	// If empty, booster-bitswap is assumed to be running privately.
	// Boost will listen on bitswap protocols on its own peer id and forward them to booster-bitswap via the protocol proxy.
	// If public addresses are set, Boost will announce the booster-bitswap peer id directly to the indexer as an extended provider.
```

node/config/types.go
```diff
@@ -218,6 +218,16 @@ type DealmakingConfig struct {
...
+	BitswapPublicAddresses []string
+
+	// If operating in public mode, in order to announce Booster-Bitswap as an extended provder, this value must point to a
+	// a file containing the booster bitswap peer id's private key. Can be left blank when operating with protocol proxy
```
dirkmc, line 228:
```suggestion
	// If operating in public mode, in order to announce booster-bitswap as an extended provider, this value must point to a
	// a file containing the booster-bitswap peer id's private key. Can be left blank when operating with protocol proxy.
```

node/modules/retrieval.go
```diff
@@ -51,9 +51,20 @@ func NewTransportsListener(cfg *config.Boost) func(h host.Host) (*lp2pimpl.Trans
 			})
 		}
 		if cfg.Dealmaking.BitswapPeerID != "" {
+			addrs := h.Addrs()
+			if len(cfg.Dealmaking.BitswapPublicAddresses) > 0 {
```
dirkmc, line 55: Let's add a comment explaining what this code is doing (and why we need it)

indexprovider/wrapper.go
```diff
@@ -36,55 +40,107 @@ var shardRegMarker = ".boost-shard-registration-complete"
...
+		var eps []xproviders.Info
+
+		// add bitswap extended provider as needed
+		if cfg.Dealmaking.BitswapPeerID != "" {
+			meta := metadata.Default.New(metadata.Bitswap{})
+			mbytes, err := meta.MarshalBinary()
+			if err != nil {
+				return nil, err
+			}
+			if len(cfg.Dealmaking.BitswapPublicAddresses) > 0 {
```
dirkmc, line 76: Let's add a comment explaining what this code is doing (and why we need it)

indexprovider/wrapper.go
```diff
@@ -36,55 +40,107 @@ var shardRegMarker = ".boost-shard-registration-complete"
...
 		if cfg.DAGStore.RootDir == "" {
 			cfg.DAGStore.RootDir = filepath.Join(r.Path(), defaultDagStoreDir)
 		}
 
 		_, isDisabled := prov.(*DisabledIndexProvider)
+
+		var eps []xproviders.Info
+
+		// add bitswap extended provider as needed
+		if cfg.Dealmaking.BitswapPeerID != "" {
```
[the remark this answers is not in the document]
hannahhoward: would rather not jump anything? we'd need to handle multiple private ids too.

indexprovider/wrapper.go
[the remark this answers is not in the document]
hannahhoward: no. In this case, you still use the bitswap peer id but it now uses the protocol proxy to forward that id.

indexprovider/wrapper.go
hannahhoward: That method just got added to the start with booster-bitswap previous release, so I don't think it's all that important. We could run it though for other reasons?

go.mod
hannahhoward: yep

Issue #960, opened 2022-11-17
Consolidating booster-http endpoints


jacobheun:
The goal of this is to clarify and create consistency with the initial endpoints supported for booster-http now that we've done a couple iterations of functionality. This is a followup of https://github.com/filecoin-project/boost/pull/946 and https://github.com/filecoin-project/boost/issues/888.

## Endpoint Overview
A user has a GIF stored on IPFS, it’s 1.9MB and totals 9 blocks. Their GIF has been aggregated with other content in a car file via an Aggregator and stored on Filecoin.

```bash
$ ipfs dag stat {cid}
Size: 1907628, NumBlocks: 9
```

The user, and relevant stakeholders (SP’s) may retrieve the data via the following methods:

**Method 1**. The user wants to get their gif back via the gateway as is, they can curl the CID without headers. This is a trusted return as you cannot verify the data until you have all of it. This is also subject to [IPLD codec deserialization](https://ipld.io/docs/codecs/) and *could* fail if the codec is unknown. In which case, the user may wish to get the CAR.

```bash
$ curl http://{booster-http.url}/ipfs/{cid} --output cool.gif
```

**Method 2**. If the user wants their data back in a CAR and handle IPLD deserialization themselves, they can add the `car` format in the header. This is particularly useful if Boost doesn't know the IPLD codec of the data:

```bash
$ curl -H "Accept:application/vnd.ipld.car;" http://{booster-http.url}/ipfs/{cid} --output cool.gif.car
```

**Method 3**. If the user wants their data back block by block, they can retrieve and parse the first block of the GIF with the `raw` ipld format. This would be useful if the user wants to iterate over the blocks, but is likely a low priority use case for Filecoin. It could be supported to match the [IPFS Trustless gateway spec](https://github.com/ipfs/specs/blob/main/http-gateways/TRUSTLESS_GATEWAY.md) though.

```bash
$ curl -H "Accept:application/vnd.ipld.raw;" http://{booster-http.url}/ipfs/{cid} --output {cid}.block
```

**Method 4**. In Filecoin we also have Pieces, which is a padded car file of the original deal. Users may wish to retrieve the whole piece byte stream to reduce risk of incompatible commp, such as when replicating deals.
- [x] https://github.com/filecoin-project/boost/issues/970

```bash
$ curl http://{booster-http.url}/piece/{cid} --output {cid}.piece
```

[...]

PR #975, opened 2022-11-21
fix: update current http urls as per #960

LexLuthr:
Fixes part of https://github.com/filecoin-project/boost/issues/960

### DIFF
--- a/cmd/booster-http/server.go
+++ b/cmd/booster-http/server.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bufio"
 	"context"
 	"encoding/json"
 	"errors"
@@ -24,10 +23,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
-	"github.com/ipld/go-car/v2"
-	"github.com/ipld/go-car/v2/index"
 	"github.com/multiformats/go-multihash"
-	"github.com/multiformats/go-varint"
 	"go.opencensus.io/stats"
 )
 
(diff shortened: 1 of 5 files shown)


dirkmc (commented):
I suggest we
- remove the `format` parameter from the /piece endpoint, as it can only ever have one value so it's redundant
- remove all code related to building car files
- note that we should also remove any methods from HttpServerApi and its implementation that are no longer used

We can always find this code in the commit history if we need it later

cmd/booster-http/server.go
```diff
@@ -151,122 +124,59 @@ func (s *HttpServer) handleIndex(w http.ResponseWriter, r *http.Request) {
...
+	startTime := time.Now()
+	ctx, span := tracing.Tracer.Start(r.Context(), "http.piece_cid")
+	defer span.End()
+	stats.Record(ctx, metrics.HttpPieceByCidRequestCount.M(1))
+
 	q, err := url.ParseQuery(r.URL.RawQuery)
 	if err != nil {
 		msg := fmt.Sprintf("parsing query: %s", err.Error())
 		writeError(w, r, http.StatusBadRequest, msg)
+		stats.Record(ctx, metrics.HttpPayloadByCid400ResponseCount.M(1))
```
jacobheun, line 146: I'm not familiar with all of the metrics vars, but as we're only doing piece retrieval by piece cid shouldnt this be `HttpPieceByCid400ResponseCount` or similar?

Same for the below values.

LexLuthr:
@jacobheun Please let me know what's the final call on Dirk's comment. It makes sense to remove that param as it is not adding any value.

jacobheun:
@LexLuthr agreed on removing it as there's only 1 option.

dirkmc:
Let's remove PiecesContainingMultihash and GetMaxPieceOffset from HttpServerApi as they are no longer used.
Let's also remove the implementations from serverApi

dirkmc (approved):

PR #976, opened 2022-11-21
Retrieval Log page

dirkmc:
Screenshot 2022-11-21 at 4 50 30 PM
Screenshot 2022-11-21 at 4 50 42 PM


willscott:
I wonder if there a way for the default log view to limit itself to successes and failures over attempts at only whole-piece retrievals.

For retrievals of single blocks/small-dags, it seems like aggregate views are going to be more useful as a primary observation point - because the number of interactions will lead to difficulty in picking out trends from only a small number.

I wonder if @brendalee  is the right person to help us think about the most useful product outcomes we should be thinking about surfacing here?

Issue #1003, opened 2022-11-23
`AlwaysKeepUnsealedCopy = false` is not being honored.


Reiers:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature or enhancement request. If it is, please open a [new idea discussion](https://github.com/filecoin-project/boost/discussions/new?category=ideas) instead. New feature and enhancement requests would be entertained by the boost team after a thorough discussion only.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to boost.

### Boost component

- [X] boost daemon - storage providers
- [ ] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [X] Other

### Boost Version

```text
boost version 1.5.0+git.dbedffb
```

### Describe the Bug

**Hi Boost team 👋** 

After looking at logs, testing, talking with clients etc. in regards to issue https://github.com/filecoin-project/lotus/issues/9396
(Snap Deals always creates an unsealed copy even though "AlwaysKeepUnsealedCopy = false")

We think we have found the reason for why `AlwaysKeepUnsealedCopy = false` is not being honored:
https://github.com/filecoin-project/boost/blob/3c68f89975498c23fed70aad5dc928027a2ed475/storagemarket/provider.go#L591

Not all SP's or clients wants fast retrieval and should be able to opt out.

### Logging Information

```text
lotus-miner storage find 16273
In FILECOIN_SERVER (Unsealed, Cache, Update, UpdateCache)
	Sealing: false; Storage: true
	Local (/nfs/TechHedgeUnit3)
	URL: http://192.168.10.80:8001/remote/unsealed/s-t022352-16273 <---
```

### Repo Steps

1. Check that your configuration on the lotus miner is correct.
2. Run `lotus-miner sectors snap-up <sectornumb>`
3. Send a deal, import and seal deal into the snapped up sector
4. See that an unsealed copy was created in the long term storage.

jacobheun:
[...]

PR #1004, opened 2022-11-24
fix: allow client to set fast-retrieval

LexLuthr:
fixes https://github.com/filecoin-project/boost/issues/1003


storagemarket/types/deal_state.go
```diff
@@ -58,6 +58,9 @@ type ProviderDealState struct {
 
 	// NBytesReceived is the number of bytes Received for this deal
 	NBytesReceived int64
+
+	// Keep unsealed copy of the data
+	FastRetrieval bool
```
dirkmc, line 63: We need to add some code to make sure the FastRetrieval flag gets written to / read from the database when the ProviderDealState is written to the DB

LexLuthr: I will add that part

nonsense (approved):

Issue #1006, opened 2022-11-24
Boostd panics after a few hours


stuberman:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature or enhancement request. If it is, please open a [new idea discussion](https://github.com/filecoin-project/boost/discussions/new?category=ideas) instead. New feature and enhancement requests would be entertained by the boost team after a thorough discussion only.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to boost.

### Boost component

- [X] boost daemon - storage providers
- [ ] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### Boost Version

```text
lotus-miner version
Daemon: 1.19.0-rc2+mainnet+git.2520b1644+api1.5.0
Local: lotus-miner version 1.19.0-rc2+mainnet+git.2520b1644

boostd -version
boostd version 1.5.1-rc1+git.f3fd05a
```

### Describe the Bug

After upgrading Lotus and Boost I see a boostd panic after a few hours. see log below

Here is the start string I use
`nohup boostd run --pprof >> /market/logs/boost.log 2>&1 &`

### Logging Information

```text
2022-11-24T15:20:09.965Z	ERROR	retrieval	impl/provider.go:434	Retrieval query: GetAsk: GetRetrievalPricingInput: failed to fetch storage deal state: 1 error occurred:
	* deal 7841268 not found - deal may not have completed sealing before deal proposal start epoch, or deal may have been slashed


2022-11-24T15:20:10.089Z	ERROR	retrieval	impl/provider.go:434	Retrieval query: GetAsk: GetRetrievalPricingInput: failed to fetch storage deal state: 1 error occurred:
	* deal 7841268 not found - deal may not have completed sealing before deal proposal start epoch, or deal may have been slashed


panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #1007, opened 2022-11-24
check for nil for evt.Response

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/1006


LexLuthr (approved):
Let's the bump the version to v1.5.1-rc2 so I can cut a release.

PR #994, opened 2022-11-23
Support peer filtering for booster-bitswap for endpoint

hannahhoward:
# Goals

fix #847 start on #954

# Implementation

- Extract periodic fetching logic from block filter to multifilter class
- Implement peer filter based on specs from @s0nik42
- Add peer filter endpoint as cli url

# For discussion

- Pending further discussion of additional retrieval configs with @s0nik42
- At somepoint booster bitswap will need a config file
- Otherwise GTG I think

PR #1020, opened 2022-11-30
Retrieval API Based Filter Configs

hannahhoward:
# Goal

Per ongoing discussions with @s0nik42, we're updating the endpoint that previously transmitted peer allow/deny lists to send additional agreed upon configuration for retrieval.

The api structure is now as follows:

```javascript
{
	"AllowDenyList": {
			"PeerIDs": ["Qma9T5YraSnpRDZqRR4krcSJabThc8nwZuJV3LercPHufi", "QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N"]
	},
        "UnderMaintenance": true,
        "StorageProviderLimits": {
	        "Bitswap": {
                        "SimultaneousRequests": 10,
                        "SimultaneousRequestsPerPeer": 4,
			"MaxBandwidth": "10mb" // human readable size metric, per second
		}
	}
}
```

# Implementations

- uses some newly added bitswap apis to get the information needed to manage limits
- implements a bandwidth measurement tool
- updates multifilter to pass on additional information to individual filters
- renames peer filter to remote config filter

# For Discussion

Blocked on #994 , https://github.com/ipfs/go-bitswap/pull/596


cmd/booster-bitswap/server.go
```diff
@@ -58,12 +64,26 @@ func (s *BitswapServer) Start(ctx context.Context, proxy *peer.AddrInfo) error {
...
+		server.WithPeerBlockRequestFilter(func(p peer.ID, c cid.Cid) bool {
+			ss := filters.ServerState{}
+			// TODO: this really isn't ideal - we should find a way to get these values passed into the peer filter directly
+			if s.server != nil {
+				ss = filters.ServerState{
+					TotalRequestsInProgress:   s.server.TotalWants(),
+					RequestsInProgressForPeer: s.server.WantCountForPeer(p),
+				}
+			}
+			fulfill, err := s.filter.FulfillRequest(p, c, ss)
```
dirkmc, line 78: Suggest we log any errors here

cmd/booster-bitswap/server.go
```diff
@@ -58,12 +64,26 @@ func (s *BitswapServer) Start(ctx context.Context, proxy *peer.AddrInfo) error {
...
+		server.MaxOutstandingBytesPerPeer(1 << 20),
+		server.WithPeerBlockRequestFilter(func(p peer.ID, c cid.Cid) bool {
+			ss := filters.ServerState{}
+			// TODO: this really isn't ideal - we should find a way to get these values passed into the peer filter directly
+			if s.server != nil {
+				ss = filters.ServerState{
+					TotalRequestsInProgress:   s.server.TotalWants(),
+					RequestsInProgressForPeer: s.server.WantCountForPeer(p),
+				}
+			}
```
dirkmc, line 77: I'm not sure that wants correspond to requests in progress. IIRC the want list is the list of blocks that a client wants from the server. If the server doesn't have the block, the want remains in the want list until it's removed by the client with a cancel message.

hannahhoward: I figured this one out I think and removed the need for the bitswap PR as a bonus!

hannahhoward:
closed in favor of #1030

dirkmc:
I'd suggest we keep this pull request open until it's approved. It makes it easier to discuss the PR.

I'm not sure that checking wantlists and sent messages (block / have / dont-have) exactly maps to "simultaneous requests". For example in certain circumstances if the server doesn't have a block it simply doesn't respond. The problem is that bitswap is not a request / response protocol.

For this reason it may not make sense to have a limit on the number of simultaneous requests. If the goal is to make sure that one peer doesn't hog all the resources, bitswap is already designed to distribute blocks in a manner that is fair to all peers. So I'd suggest we just drop the simultaneous requests limits (but keep the bandwidth limit).

dirkmc:
With respect to versioning, I'm not sure I was clear in what I suggested. Copying from the PR description, the api structure is now as follows:
```
{
	"AllowDenyList": {
			"PeerIDs": ["Qma9T5YraSnpRDZqRR4krcSJabThc8nwZuJV3LercPHufi", "QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N"]
	},
        "UnderMaintenance": true,
        "StorageProviderLimits": {
	        "Bitswap": {
                        "SimultaneousRequests": 10,
                        "SimultaneousRequestsPerPeer": 4,
			"MaxBandwidth": "10mb" // human readable size metric, per second
		}
	}
}
```

I'm suggesting we add a version field to this structure, so that if (for example) we add "Graphsync" to "StorageProviderLimits" we can update the version, and it's easier for the process that parses this structure to understand which fields are expected.

hannahhoward:
@dirkmc 

- I looked into limiting bandwidth at the libp2p level. Libp2p has limits for connections and streams, but it actually doesn't have the kind of per-protocol bandwidth limiting you're talking about. 

- If we simply limit outgoing bandwidth without rejecting requests, the we'd end up in a situation where no one gets any responses quickly, with the task queue just growing, potentially indefinitely. It's certainly possible to use MessageSent hook as a way to slowdown outgoing bandwidth usage if we wanted to go that route, but I think rejecting requests is less failure prone.

- Re: request limits -- so you raise a good point about not sending blocks at all. I saw the global SendDontHave (set true on our instance) but I didn't see the entry based SendDontHave. I guess that's the case where it wouldn't send anything. There are probably ways around that if we need.
     - The not smart fix is just adding a timeout based deletion of simultaneous requests in the booster-bitswap code
     - If we're concerned with "correctness" -- the number of tasks in the peer's taskqueue is probably best representation of "simultaneous requests" -- that would require some modifications to that queue but could be done.

- Re: removing simultaneous limits entirely -- it seems like we should have involved you earlier in the discussions with @s0nik42 and Slingshot Evergreen. I'm not sure the best way to proceed. The SPs want these limits --- while the peer task queue provides a mechanism of "fairness" -- it doesn't provide what SPs are concerned about which is hard limiting their resource usage. Again, while I think we had hoped Libp2p could help here, it really just understands connections and streams. Especially cause Bitswap isn't request response, that doesn't provide much.

dirkmc:
> If we simply limit outgoing bandwidth without rejecting requests, the we'd end up in a situation where no one gets any responses quickly

Bitswap is designed to handle this use case - because it operates in a peer-to-peer environment, it needs to be resilient to getting spammed by a single peer. Essentially it will distribute blocks evenly amongst peers, so that no one peer can clog up the queue.

> it doesn't provide what SPs are concerned about which is hard limiting their resource usage. Again, while I think we had hoped Libp2p could help here, it really just understands connections and streams

I understand that SPs want to limit resource usage, that makes sense 👍 
As explained above, I don't think the limit should be applied to the number of concurrent requests, because of the nature of how bitswap works.
On the other hand I do think we should apply a bandwidth limit. If this is not currently available in libp2p I suggest we either add it there, or add a wrapper somewhere in our code or a new library.

I realize that's going to take time, so I suggest we punt on this requirement. Let's start testing booster-bitswap with SPs and work on bandwidth limiting in the mean time.

s0nik42:
To clarify about simultaneous requests, the correct phrasing should be "simultaneous TCP connections". So it's not based on a per request, but more per simultaneous TCP active connection.
I confirmed that during the SP call yesterday. Does that make a difference on your side ?

- Limiting the bandwidth at the libp2p level is probably good. 

What SP wants to protect themself from : 
- Out of Server ressource
- Out of network equipment resources (router/fw)
- Controlling how much bandwidth from the physical access is allocated to retrieval

dirkmc:
Thanks Julien, that makes sense 👍 
We already have a Connection Manager in libp2p that should be used for this purpose.

hannahhoward:
> To clarify about simultaneous requests, the correct phrasing should be "simultaneous TCP connections". So it's not based on a per request, but more per simultaneous TCP active connection.

OK that's helpful, we can certainly use this metric instead, and it's much simpler so that's great.

It's worth noting that currently the ConnManager won't let you change limits after its constructed, so this would be a change on CIDGravity wouldn't take effect until you rebooted booster-bitswap, unless we made a change in go-libp2p's conn manager. 

It also won't work the way you expect if you run booster bitswap proxied through boostd.  Connection limiting then is just boostd's job.

The same issues apply for the resource manager if we went that route. (only sets limits at startup, won't work the way you expect in proxied mode)

We can maybe doing this implementing a ConnectionGater instead if we want limits that change while running?

> Limiting the bandwidth at the libp2p level is probably good.

Ok, so we can certainly explore doing it through libp2p but it looks like a longer discussion that will take a while based off some early convos with the libp2p folks. 

The alternative is just to block the MessageSent hook I'm using already to limit bandwidth. It's an indirect approach but it has a much shorter implementation path. In fact, we use this exact approach to backpressure selector execution flow in go-graphsync, so I'm pretty sure there's some code I can copy that will do almost exactly what we want.

hannahhoward:
> Bitswap is designed to handle this use case - because it operates in a peer-to-peer environment, it needs to be resilient to getting spammed by a single peer. Essentially it will distribute blocks evenly amongst peers, so that no one peer can clog up the queue.

It's worth noting the peer task queue actually has pretty easily exploitable DDOS vulnerabilities. The notion of fairness is based on peer ID, which apparently are trivial to manufacture thousands of. So it only really avoids clogging when there are no malicious actors. I think that's why TCP level connection limiting makes more sense for SPs.

dirkmc:
I think something like the connection gater approach sounds good 👍 

Blocking the message sent hook sounds like a bit of a hack..

hannahhoward:
I mean, I'm not sure what would be wrong with blocking the message sent hook. It would work. again -- we use these mechanisms in graphsync.

hannahhoward:
Also though, another option is to simply tell SPs who care to put boost on a public node and configure bandwidth usage on a firewall.

hannahhoward:
Closing pending future discussion

PR #1023, opened 2022-12-01
add command to generate index for a given `piececid` and add that index to the new Piece Directory

nonsense:
This PR is adding a CLI command which fetches a specific `piececid`, generates an index for it, and pushes the records to the Piece Directory.

It support both:
- [x] fetching the `piececid` from the storage system if an unsealed copy exists in a sector
- [x] fetching the `piececid` from a specific car file

The PR is also cleaning up deprecated APIs for the deprecated `PieceStore`

---

Note that this PR is also modifying the `Piece Directory` to explicitly return a descriptive error if the storage system can't find an unsealed copy of a given `piececid`.

---

Address TODOs:

- [x] shall we `SetCarSize` when adding the index to the piece directory?
- [x] confirm that `OffsetSize.Size` value is actually correct within individual `model.Record` entries
- [ ] confirm change in behaviour in `PieceDirectory.GetPieceDeals` is OK (i.e the fact that we now return a specific error when an unsealed copy is not available)


cmd/boostd/genindex.go
```diff
@@ -0,0 +1,221 @@
...
+		if filepath := cctx.String("filepath"); filepath != "" {
+			fmt.Println("fetching piececid from car file: ", filepath)
+
+			r, err = carv2.OpenReader(filepath)
+			if err != nil {
+				return fmt.Errorf("open reader: %w", err)
+			}
+			defer r.Close()
+
+			if r.Version == 1 {
```
LexLuthr, line 108: Typically we deal with carV1 but if user supplies a carv2 we should accept it as well and directly take the indexed from it. WDYT?
And we don't want to do that then perhaps we should let the user know that they have supplied a cav2 and return from there.

piecedirectory/piecedirectory.go
```diff
@@ -109,6 +113,11 @@ func (ps *PieceDirectory) GetPieceDeals(ctx context.Context, pieceCid cid.Cid) (
 		return nil, fmt.Errorf("listing deals for piece %s: %w", pieceCid, err)
 	}
 
+	//TODO: confirm this change in behaviour is ok
+	if len(deals) == 0 {
```
LexLuthr, line 117: Wouldn't clean up prevent this from hapenning? If a piece has no deals, it should be cleaned up.

cmd/boostd/genindex.go
```diff
@@ -0,0 +1,221 @@
...
+				fi, err := os.Stat(filepath)
+				if err != nil {
+					return fmt.Errorf("os.stat err: %w", err)
+				}
+				headerDataSize = uint64(fi.Size())
+			}
+		} else {
+			fmt.Println("no car file specified, fetching piececid from the storage api")
```
LexLuthr, line 116: Can we simply expose  `buildIndexForPiece` method in piecedirectory and call it since this would be an internal process?

LexLuthr (commented):
Just a few questions

cmd/boostd/genindex.go
nonsense: > Typically we deal with carV1 but if user supplies a carv2 we should accept it as well and directly take the indexed from it.

I think we have agreed that we always re-generate indices and don't take them from the client even if they are there.

piecedirectory/piecedirectory.go
nonsense: How can a `piece` have no deals? I don't think this is possible.

I am not aware of any case where `err == nil && len(deals) == 0` but the one where we don't have an unsealed copy.

cmd/boostd/genindex.go
nonsense: Yes, all this code probably shouldn't belong on the CLI layer.

api/api.go
```diff
@@ -72,12 +71,6 @@ type Boost interface {
 	MarketPendingDeals(ctx context.Context) (lapi.PendingDealInfo, error)                                                                                                                //perm:write
 	SectorsRefs(context.Context) (map[string][]lapi.SealedRef, error)                                                                                                                    //perm:read
 
-	PiecesListPieces(ctx context.Context) ([]cid.Cid, error)                                 //perm:read
-	PiecesListCidInfos(ctx context.Context) ([]cid.Cid, error)                               //perm:read
-	PiecesGetPieceInfo(ctx context.Context, pieceCid cid.Cid) (*piecestore.PieceInfo, error) //perm:read
-	PiecesGetCIDInfo(ctx context.Context, payloadCid cid.Cid) (*piecestore.CIDInfo, error)   //perm:read
-	PiecesGetMaxOffset(ctx context.Context, pieceCid cid.Cid) (uint64, error)                //perm:read
-
```
dirkmc, line 80: Thanks for cleaning this up 👍

cmd/boostd/genindex.go
```diff
@@ -0,0 +1,221 @@
...
+		},
+		&cli.StringFlag{
+			Name:     "api-piece-directory",
+			Usage:    "the endpoint for the piece directory API",
+			Required: false,
+		},
+		&cli.StringFlag{
+			Name:     "piececid",
+			Usage:    "piececid to index",
+			Required: false,
```
dirkmc, line 50: Looking at the code in the Action it seems like piececid should be a required variable
Nit: suggest calling it piece-cid to match the casing of the other parameters.

cmd/boostd/genindex.go
```diff
@@ -0,0 +1,221 @@
...
+		},
+		&cli.StringFlag{
+			Name:     "filepath",
+			Usage:    "the path to the car",
+			Required: false,
+		},
```
dirkmc, line 56: Let's add a description here or in the Command description explaining when you might want to use the filepath parameter.

cmd/boostd/genindex.go
```diff
@@ -0,0 +1,221 @@
...
+
+		fmt.Println("adding index took", time.Since(addStart).String())
+
+		fmt.Printf("successfully added index (%d records) for piececid %s to the piece directory\n", len(recs), piececid)
+
+		return nil
+	},
+}
+
+func GetRecords(r *carv2.Reader, headerDataSize uint64) ([]model.Record, error) {
```
dirkmc, line 157: I think instead of this function we can just use `PieceDirectory.buildIndexForPiece()`

cmd/boostd/genindex.go
```diff
@@ -0,0 +1,221 @@
...
+		&cli.StringFlag{
+			Name:     "api-storage",
+			Usage:    "the endpoint for the storage node API",
+			Required: false,
+		},
+		&cli.StringFlag{
+			Name:     "api-piece-directory",
+			Usage:    "the endpoint for the piece directory API",
+			Required: false,
+		},
```
dirkmc, line 46: I think we can assume that boostd is running - so then we don't need to supply all these parameters, we can just add an API method to boostd and call it from the CLI

cmd/boostd/genindex.go
```diff
@@ -0,0 +1,221 @@
...
+		defer storageCloser()
+
+		pr := &piecedirectory.SectorAccessorAsPieceReader{SectorAccessor: sa}
+
+		pd := piecedirectory.NewPieceDirectory(pdClient, pr, cctx.Int("add-index-throttle"))
+
+		var r *carv2.Reader
+		var headerDataSize uint64
+
+		if filepath := cctx.String("filepath"); filepath != "" {
```
dirkmc, line 99: What's the use case for adding an index from a CAR file? If we can't retrieve the piece from the sealer then I don't think we want an index for it.

piecedirectory/piecedirectory.go
dirkmc: At the moment GetPieceDeals doesn't know anything about sealing - it just knows about the deals associated with a piece.
Note that it will return [ErrNotFound](https://github.com/filecoin-project/boost/blob/052c81a0fbd48e78d018bbaa5bf066202dff6a52/extern/boostd-data/svc/types/types.go#L14) if there is no metadata for a given piece cid (ie there are no deals and no index).

react/src/Inspect.js
```diff
@@ -138,27 +138,39 @@ function PieceStatus({pieceCid, pieceStatus, searchQuery}) {
...
+                        <th>Data Root CID</th>
+                        <td><strong>{rootCid}</strong></td>
+                    </tr>
+                ) : null}
+                {rootCid && !searchIsRootCid ? (
                     <tr key="data root cid">
                         <th>Data Root CID</th>
                         <td>{rootCid}</td>
                     </tr>
                 ) : null}
```
dirkmc, line 166:
```suggestion
                {rootCid ? (
                    <tr key="data root cid">
                        <th>Data Root CID</th>
                        <td>
                            { searchIsRootCid ? <strong>{rootCid}</strong> : rootCid }
                        </td>
                    </tr>
                ) : null}
```

react/src/Menu.js
```diff
@@ -28,6 +29,7 @@ export function Menu(props) {
             <DealPublishMenuItem />
             <DealTransfersMenuItem />
             <InspectMenuItem />
+            <PiecesMenuItem />
```
dirkmc, line 32: At the moment the data in PieceDirectory.js is mock data, so not sure if we want to expose it in the UI yet

cmd/boostd/genindex.go
nonsense: It was useful during debugging, but you're right, it probably doesn't make sense. Removing it.

cmd/boostd/genindex.go
nonsense: Removed the filepath option.

piecedirectory/piecedirectory.go
nonsense: At some point in the stack, we should detect that there are no unsealed deals. I am thinking this is a good place, because the API of returning `nil` error and `nil` DealInfo, doesn't make sense and is not expected.

When calling this API I expect to either get deals, or get an error basically.

nonsense: @dirkmc you're right this is the wrong place to add this.

PR #1032, opened 2022-12-06
Piece Directory Doctor

dirkmc:
TODO:
- [ ] leveldb implementation of piece tracking (couchbase implementation is done)
- [x] Clean up UI (eg back button doesn't work from piece detail page)
- [ ] Test in production


extern/boostd-data/couchbase/service.go
```diff
@@ -255,24 +260,51 @@ func (s *Store) ListPieces(ctx context.Context) ([]cid.Cid, error) {
...
-func normalizePieceCidError(pieceCid cid.Cid, err error) error {
-	if err == nil {
-		return nil
-	}
-	if isNotFoundErr(err) {
-		return fmt.Errorf("piece %s: %s", pieceCid, types.ErrNotFound)
-	}
-	return err
+//const pieceCheckFrequencyMillis = 60 * 60 * 1000
+const pieceCheckFrequencyMillis = 1000
```
nonsense, line 264: Shouldn't this be set to 1 day -- i.e. 3600 * 24 * 1000 ?

dirkmc: Yes, we should change these values once we finish testing 👍

piecedirectory/test_util.go
```diff
@@ -0,0 +1,118 @@
+package piecedirectory
```
nonsense, line 1: Nit but maybe name this file `util_test.go` so that it is not compiled in the default binary if these are only used for tests.

dirkmc: Unfortunately I don't think we can use the methods from a different package if the filename ends with `_test.go`

nonsense (approved):

Issue #1041, opened 2022-12-14
booster-bitswap: Add flag for badbits filter url


dirkmc:
We should add a flag to booster-bitswap to allow the user to specify a custom badbits filter url:
`--badbits-denylist=<url>`

This also will help with testing booster-bitswap.

PR #1043, opened 2022-12-15
custom badbits filter

LexLuthr:
fixes https://github.com/filecoin-project/boost/issues/1041

### DIFF
--- a/cmd/booster-bitswap/filters/filters.go
+++ b/cmd/booster-bitswap/filters/filters.go
@@ -28,6 +28,7 @@ const expectedListGrowth = 128
 // FetcherForHTTPEndpoint makes an fetcher that reads from an HTTP endpoint
 func FetcherForHTTPEndpoint(endpoint string, authHeader string) Fetcher {
 	return func(ifModifiedSince time.Time) (bool, io.ReadCloser, error) {
+		// Allow empty badbits list
 		if endpoint == "" {
 			return false, nil, nil
 		}
@@ -134,17 +135,12 @@ func NewMultiFilter(
 	requestCounter RequestCounter,
 	apiFilterEndpoint string,
 	apiFilterAuth string,
-	customBadBitsDenyList string,
+	BadBitsDenyList string,
 ) *MultiFilter {
 	filters := []FilterDefinition{
 		{
 			CacheFile: filepath.Join(cfgDir, "denylist.json"),
 			Fetcher:   FetcherForHTTPEndpoint(BadBitsDenyList, ""),
-			Handler:   NewBlockFilter(),
-		},
-		{
-			CacheFile: filepath.Join(cfgDir, "customdenylist.json"),
-			Fetcher:   FetcherForHTTPEndpoint(customBadBitsDenyList, ""),
 			Handler:   NewBlockFilter(),
 		},
 	}
--- a/cmd/booster-bitswap/filters/filters.go
+++ b/cmd/booster-bitswap/filters/filters.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/benbjohnson/clock"
 	"github.com/ipfs/go-cid"
-	peer "github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peer"
 )
 
 // UpdateInterval is the default interval at which the public list is refected and updated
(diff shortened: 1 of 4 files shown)


cmd/booster-bitswap/filters/filters.go
```diff
@@ -131,13 +131,19 @@ func NewMultiFilter(
...
 		{
 			CacheFile: filepath.Join(cfgDir, "denylist.json"),
 			Fetcher:   FetcherForHTTPEndpoint(BadBitsDenyList, ""),
 			Handler:   NewBlockFilter(),
 		},
+		{
+			CacheFile: filepath.Join(cfgDir, "customdenylist.json"),
+			Fetcher:   FetcherForHTTPEndpoint(customBadBitsDenyList, ""),
+			Handler:   NewBlockFilter(),
+		},
```
willscott, line 146: can this just be 1 deny list, where it defaults to the canonical `BadBits` location when not over-ridden?

LexLuthr: @willscott We don't want to override the default one. I would prefer to keep the default one always on and deny additional ones based on the SP's input. This would be useful in cases SPs themselves have identified bad content. The default Badbists list is already verified and should always be denied to avoid copyright issues.

LexLuthr:
`GOLOG_LOG_LEVEL=remote-blockstore=debug,booster=debug,engine=debug,bitswap-server=debug booster-bitswap --repo ~/lexluthr/.bitswap run --api-boost=$BOOST_API_INFO --port 8889 --badbits-denylist https://www.jsonkeeper.com/b/UUJZ`

```
ls -la ../.bitswap/
total 2248
drwxr--r-- 2 nonsense nonsense    4096 Dec 15 00:23 .
drwxrwxr-x 4 nonsense nonsense    4096 Dec 15 00:16 ..
-rw------- 1 nonsense nonsense      79 Dec 15 00:23 customdenylist.json
-rw------- 1 nonsense nonsense 2281692 Dec 15 00:23 denylist.json
-rw------- 1 nonsense nonsense      68 Dec 15 00:16 libp2p.key
```

```
 less ../.bitswap/customdenylist.json 

[{"anchor":"7f26e7813526dfa529020b7611a62e8f2c496558f80a82bc24a95d4e4105125d"}]
```

ipfs get bafykbzacebmouqhg33afbz2sb57ivdfxpw3si6w7q7or32k53axuhhvi3myms  

`DEBUG	engine	decision/engine.go:718	Bitswap engine: block denied access	{"local": "12D3KooWKwGVaK4bBrLFhuXjZUHtB3mHZ3WhmTkfFaS9s81w8ZCW", "from": "12D3KooWJqqmmfa9SExfr3yuAVaFU8X2LpZLSSGaanFGmkwdYfNS", "cid": "bafykbzacebmouqhg33afbz2sb57ivdfxpw3si6w7q7or32k53axuhhvi3myms", "sendDontHave": false}`

cmd/booster-bitswap/filters/filters.go
willscott: * agree we want SPs to be able to identify bad content and block it.
* can the flag for 'bad bits' that can be configured be an array of urls? when not specified it can be defaulted to `[OfficialBadBits]`
* The ability to override the default one is important for decentralization, "you don't have to rely on centralized infrastructure to be an SP"

LexLuthr: That makes sense. Let me redesign with the ability to override the default list if required.

dirkmc:
I'd suggest for now let's keep it simple and just override the default list location.

cmd/booster-bitswap/filters/filters.go
hannahhoward: huh I totally never considered the option of multiple filters but the multifilter would totally support that. But I agree, I'd rather it  be a replacable array rather than a single optional add.

hannahhoward: and I don't like initializing a filter in the absense of a URL

cmd/booster-bitswap/server.go
```diff
@@ -13,7 +13,7 @@ import (
 	blockstore "github.com/ipfs/go-ipfs-blockstore"
 	nilrouting "github.com/ipfs/go-ipfs-routing/none"
 	"github.com/libp2p/go-libp2p/core/host"
-	peer "github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peer"
```
hannahhoward, line 16: thanks for these corrections

hannahhoward (changes requested):
Agree with @willscott  comments.

Recommend:
- NewMultifilter takes an array of urls for block filtering
- Default value is [official badbits]

dirkmc (approved):
LGTM 👍

PR #1045, opened 2022-12-16
couchbase / piece-directory tests

nonsense:


piecedirectory/couchbase_test.go
```diff
@@ -0,0 +1,255 @@
...
+			"8092":  struct{}{},
+			"8093":  struct{}{},
+			"8094":  struct{}{},
+			"8095":  struct{}{},
+			"8096":  struct{}{},
+			"11210": struct{}{},
+			"11211": struct{}{},
+		},
+	}, &container.HostConfig{
+		PortBindings: map[nat.Port][]nat.PortBinding{
```
willscott, line 55: how much harder would it be to make these not fixed ports on the host so that multiple runs won't conflict?

piecedirectory/couchbase_test.go
```diff
@@ -0,0 +1,255 @@
...
+	start := time.Now()
+	for {
+		fullPath := "http://127.0.0.1:8091" + path
+		resp, err := http.Get(fullPath)
+		if resp != nil && resp.Body != nil {
+			_ = resp.Body.Close()
+		}
+		if err == nil && resp != nil {
+			if resp.StatusCode < 300 {
+				return
```
willscott, line 225: if there's no response, the status code might be 0?

piecedirectory/couchbase_test.go
nonsense: all tests should really use a single instance of couchbase, we shouldn't start and stop it in between tests, as that's a quite slow process...

if cleanup is not done properly, i am not sure starting a bunch of instances on a different port is really nice either, i think we'd rather just get a conflict.

nonsense: if we want to be really clean, then it makes sense to use custom and non-default couchbase ports, but thats really not a pain point right now.

dirkmc (approved):
Looks great ✅

dirkmc:
I'm merging the work that has been done so far so that we can get CI green again. I've added a ticket for the remaining work at https://github.com/filecoin-project/boost/issues/1053

PR #1049, opened 2022-12-26
boostd add cmd: set-ask and get-ask

llifezou:
boost does not have a command to set prices, for example `lotus-miner storage-deals set-ask`. If using boost, the `lotus-miner storage-deals set-ask` will not be available because lotus's market is disabled. But when I don't want to use the UI, I think the command line should have a command that allows the operation.

### DIFF
--- a/cmd/boostd/main.go
+++ b/cmd/boostd/main.go
@@ -44,6 +44,7 @@ func main() {
 			dummydealCmd,
 			dataTransfersCmd,
 			retrievalDealsCmd,
+			storageDealsCmd,
 			indexProvCmd,
 			importDataCmd,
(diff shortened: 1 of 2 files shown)


dirkmc:
@llifezou thank you for contributing to the project 🙏 

Currently we are recommending that users use the UI to set the price.
If that's not possible we already have a graphql endpoint that the user can call to set the prices. See https://boost.filecoin.io/graphql-api for more information on how to use the graphql API. The particular API call for setting prices is `storageAskUpdate`

free1139:
@dirkmc Is there an executable sample of 'storageAskUpdate'?

PR #1051, opened 2023-01-04
feat: client flag to not announce deals

LexLuthr:


dirkmc (commented):
Looks great so far 👍 

As part of this PR, let's also add the fast retrieval and announce to IPNI fields on the deal detail page. This will have the side effect of making it easier to debug what happens when an old client (without knowledge of these flags) makes a request to a new version of boostd. Let me know if you need help with that.

Let's also add a test to provider_test.go with two subtests:
1. Executes a deal with the AnnounceToIPNI flag set to false
2. Executes a deal with the AnnounceToIPNI flag set to true

The code for the test itself should be quite short. It needs to
- set up the harness
- start
- execute
- wait for the deal to reach the IndexedAndAnnounced stage
- verify whether the deal was announced

However it may take some digging to understand how to set the parameters up and how to mock out the deal announcement code. Again let me know if you need help with that.

storagemarket/types/dealcheckpoints/checkpoints.go
```diff
@@ -10,6 +10,7 @@ const (
 	Published
 	PublishConfirmed
 	AddedPiece
+	Indexed
```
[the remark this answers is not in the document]
LexLuthr: I added the new checkpoint to streamline the UI. Otherwise, we will see the status in UI and client deal-status as "IndexedAndAnnounced" which might confuse folks when they chose not to announce. Let me know if that doesn't make sense and I will undo the checkpoint.

cmd/boost/deal_cmd.go
```diff
@@ -86,6 +86,11 @@ var dealFlags = []cli.Flag{
 		Name:  "wallet",
 		Usage: "wallet address to be used to initiate the deal",
 	},
+	&cli.BoolFlag{
+		Name:  "AnnounceToIPNI",
```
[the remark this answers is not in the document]
LexLuthr: Resolving this as we are renaming it to "skip-announce-to-ipni"

storagemarket/types/dealcheckpoints/checkpoints.go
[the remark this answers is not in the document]
dirkmc: Let's just change the string that we display in the UI to "Indexed".
I think it's better not to change the deal status because clients might already be relying on the name of the checkpoint.

cmd/boost/deal_cmd.go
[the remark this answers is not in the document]
dirkmc: I think we should leave the CLI command parameter as "announce-to-ipni".
We're changing the name of the field that is sent over the network. But I think for the CLI it's less confusing to have `--announce-to-ipni=true` rather than `--skip-announce-to-ipni=false`.

db/migrations_tests/deals_fast_retrieval_test.go
```diff
@@ -22,11 +22,15 @@ func TestDealFastRetrieval(t *testing.T) {
...
 	deals, err := db.GenerateNDeals(1)
 	req.NoError(err)
 
+	// Add AnnounceToIPNI to allow tests to works
+	_, err = sqldb.Exec(`ALTER TABLE Deals ADD AnnounceToIPNI BOOL`)
+	require.NoError(t, err)
+
 	// Insert the deals in DB
 	err = dealsDB.Insert(ctx, &deals[0])
 	require.NoError(t, err)
```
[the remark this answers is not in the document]
LexLuthr: I used the Insert column method to avoid writing a very lengthy manual DB INSERT. Also, if we manually add the entry to DB, we are bypassing all the deal insert code. This method seemed the most clean to me.
Can you please let me know why manual INSERT makes more sense?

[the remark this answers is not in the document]
dirkmc: At the beginning of the test we run migrations up to a certain point. But the database management code is written to work with the latest database structure. So if we use the `dealsDB.Insert(...)` method it will fail because the code references fields that do not exist in the database structure until all migrations have been run.

One way around that is to write an ALTER TABLE statement as you have done. However in that case you are effectively doing the same thing as running all the database migrations. Each time we add a new migration you will need to add more code to each of these migration tests to manually perform the migration.

So instead I think we should write a manual INSERT statement that works with the structure of the database at the point the migration has reached.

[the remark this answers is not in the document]
LexLuthr: That makes sense. I will update both tests using the ALTER statement.

storagemarket/provider_test.go
```diff
@@ -1103,6 +1103,61 @@ func TestDealVerification(t *testing.T) {
 	}
 }
 
+func TestIPNIAnnounce(t *testing.T) {
```
dirkmc, line 1106: The "announce" and "don't announce" subtests have almost identical code, so I'd suggest that we refactor them so that each subtest calls the same method with a boolean parameter indicating whether to announce or not. Take a look at TestSimpleDealHappy for an example

storagemarket/provider_test.go
```diff
@@ -1103,6 +1103,61 @@ func TestDealVerification(t *testing.T) {
...
+
+	t.Run("announce", func(t *testing.T) {
+		// setup the provider test harness
+		harness := NewHarness(t)
+		// start the provider test harness
+		harness.Start(t, ctx)
+		defer harness.Stop()
+
+		td := harness.newDealBuilder(t, 1).withAllMinerCallsNonBlocking().withNormalHttpServer().build()
+		td.params.SkipAnnounceToIPNI = false
```
dirkmc, line 1117: Rather than setting `td.params.SkipAnnounceToIPNI` explicitly, I'd suggest we follow the pattern of how the other parameters are set: add a `withAnnouncement(bool)` method to the builder that sets `td.params.SkipAnnounceToIPNI` appropriately

storagemarket/provider_test.go
```diff
@@ -1103,6 +1103,61 @@ func TestDealVerification(t *testing.T) {
...
+		check := false
+		for _, m := range dlogs {
+			if strings.Contains(m.LogParams, "announcement-cid") {
+				check = true
+				break
+			} else {
+				continue
+			}
+		}
+		require.True(t, check)
```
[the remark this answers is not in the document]
dirkmc: I tried out setting the number of times the method is expected to be called to 0 if SkipAnnounceToIPNI is true, and 1 if SkipAnnounceToIPNI is false, but I couldn't get it to work. Let me take another look in the morning.

storagemarket/provider_test.go
dirkmc: Resolved in [be5958e](https://github.com/filecoin-project/boost/pull/1051/commits/be5958e0b657e897f70c73589710d16557cb71ef)

storagemarket/provider_test.go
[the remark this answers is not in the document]
dirkmc: Resolved in [be5958e](https://github.com/filecoin-project/boost/pull/1051/commits/be5958e0b657e897f70c73589710d16557cb71ef)

storagemarket/provider_test.go
dirkmc: Resolved in [be5958e](https://github.com/filecoin-project/boost/pull/1051/commits/be5958e0b657e897f70c73589710d16557cb71ef)

dirkmc:
@LexLuthr I fixed the tests, so the only remaining part is the CLI changes.

Please let me know when you've tested out the CLI changes and the UI changes.

gql/resolver.go
```diff
@@ -372,6 +372,14 @@ func (dr *dealResolver) IsVerified() bool {
 	return dr.ProviderDealState.ClientDealProposal.Proposal.VerifiedDeal
 }
 
+func (dr *dealResolver) KeepUnsealedCopy() bool {
```
dirkmc, line 375: I think you may also need to add these fields to `gql/schema.graphql` and to `react/src/gql.js`

db/migrations_tests/storagetagged_set_host_test.go
```diff
@@ -31,6 +31,10 @@ func TestStorageTaggedSetHost(t *testing.T) {
 	_, err := sqldb.Exec(`ALTER TABLE Deals ADD FastRetrieval BOOL`)
 	require.NoError(t, err)
 
+	// Add AnnounceToIPNI to allow tests to works
+	_, err = sqldb.Exec(`ALTER TABLE Deals ADD AnnounceToIPNI BOOL`)
+	require.NoError(t, err)
+
```
dirkmc, line 37: Let's remove these two ALTER TABLE commands and instead add an INSERT like you did for deals_fast_retrieval_test.go

storagemarket/types/dealcheckpoints/checkpoints.go
```diff
@@ -20,7 +20,7 @@ var names = map[Checkpoint]string{
 	Published:           "Published",
 	PublishConfirmed:    "PublishConfirmed",
 	AddedPiece:          "AddedPiece",
-	IndexedAndAnnounced: "IndexedAndAnnounced",
+	IndexedAndAnnounced: "Indexed",
```
dirkmc, line 23: Let's leave this as IndexedAndAnnounced because clients may be relying on it to have that name

storagemarket/provider_test.go
```diff
@@ -629,7 +535,7 @@ func TestDealAutoRestartAfterAutoRecoverableErrors(t *testing.T) {
 			require.NotNil(t, dh)
 
 			//Check for fast retrieval
-			require.False(t, td.params.FastRetrieval)
+			require.False(t, !td.params.RemoveUnsealedCopy)
```
dirkmc, line 538:
```suggestion
			require.True(t, td.params.RemoveUnsealedCopy)
```

LexLuthr:
gql
```
{
  "data": {
    "deals": {
      "deals": [
        {
          "ID": "7f7384e7-9a56-4c64-bd47-464e1b771fa9",
          "KeepUnsealedCopy": true,
          "AnnounceToIPNI": true
        }
      ]
    }
  }
}
```

Screenshot 2023-01-06 at 6 51 06 PM

cmd/boost/deal_cmd.go
```diff
@@ -78,14 +78,19 @@ var dealFlags = []cli.Flag{
...
 	},
 	&cli.StringFlag{
 		Name:  "wallet",
 		Usage: "wallet address to be used to initiate the deal",
 	},
+	&cli.BoolFlag{
+		Name:  "announce-to-ipni",
+		Usage: "indicates that deal index should be announced to the IPNI(Network Indexer)",
+		Value: true,
+	},
```
ribasushi, line 93: These are new, never before existing CLI options: give them the same treatment as the protocol - explicitly asking for the less desirable thing
`--remove-unsealed-copy`
`--skip-ipni-announce`

documentation/en/api-v1-methods.md
```diff
@@ -464,7 +465,8 @@ Response:
   "Err": "string value",
   "Retry": "auto",
   "NBytesReceived": 9,
-  "FastRetrieval": true
+  "FastRetrieval": true,
+  "AnnounceToIPNI": true
```
ribasushi, line 469: These seem wrong...

gql/schema.graphql
```diff
@@ -45,6 +45,8 @@ type Deal {
   PieceCid: String!
   PieceSize: Uint64!
   IsVerified: Boolean!
+  AnnounceToIPNI: Boolean!
+  KeepUnsealedCopy: Boolean!
```
ribasushi, line 49: Shouldn't the API match the protocol?

cmd/boost/deal_cmd.go
dirkmc: On the CLI we can set a default value more easily, so I would favour using a clearer name. To me boolean names are clearer when they're positive like "--announce-to-ipni" rather than negative "--skip-ipni-announce".

documentation/en/api-v1-methods.md
dirkmc: These are auto-generated by the docs tool

gql/schema.graphql
dirkmc: This is just for display on the front end, and again I think it's clearer to use a positive name for booleans

cmd/boost/deal_cmd.go
brendalee: Thanks Dirk and Riba! While the negative names (remove-unsealed-copy and skip-ipni-announce) are definitely annoying, I think we do want to keep the defaults of these booleans to false to align with general developer experience. In the case where an SP skips setting these flags all together, having the flags as false would be aligned with what they would expect. 

Unfortunately... if we want the defaults to be false, then the names will be negative... :(

documentation/en/api-v1-methods.md
brendalee: We should update the FastRetrieval naming to be aligned with the KeepUnsealedCopy name we are using. @dirkmc - if there's somewhere we are still using FastRetrieval, let's remove it to reduce confusion...

cmd/boost/deal_cmd.go
dirkmc: When a user doesn't supply these flags on the command line, they will default to the value we chosen for them.
In this case we have coded the parameter name to be `--announce-to-ipni` and the default value to be `true`.

Some example commands and the result:
Flag value | Example Command | Whether index will be announced
---------- | ------- | -----
`true` | `boost deal --announce-to-ipni=true` | `Yes`
`false` | `boost deal --announce-to-ipni=false` | `No`
Not set | `boost deal` | `Yes`

Note that the Command Line Interface is distinct from the [Propose Storage Deal Protocol](https://boost.filecoin.io/boost-architecture/libp2p-protocols#propose-storage-deal-protocol). Most SPs probably won't need to know of the protocol, they will just interact with the CLI tool.

documentation/en/api-v1-methods.md
dirkmc: Ok we can do that - note though that it will be a breaking change, so we would need to change the Propose Storage Deal Protocol major version (rather than just the minor version). This will require a little more work on our end.

cmd/boost/deal_cmd.go
RobQuistNL: SP's doing deals with CLI are probably only testing and preparing to use the actual protocol though

dirkmc: We talked sync and decided we will use the "negative" naming:
`--remove-unsealed-copy` (default false)
`--skip-ipni-announce` (default false)

documentation/en/api-v1-methods.md
dirkmc: Sorry my mistake - these variable names are API methods that are called by the boost client.
I don't believe anyone outside our team is using these API methods so it should be fine to change them. But the naming is unrelated to this PR so I'd suggest we address it in a follow-up PR so that we can merge this one.

brendalee: that works for me, thank you!

dirkmc (approved):

PR #1059, opened 2023-01-09
fix: vacuum db after log cleanup

LexLuthr:


db/logs.go
```diff
@@ -72,5 +72,12 @@ func (d *LogsDB) CleanupLogs(ctx context.Context, daysOld int) error {
 	qry := "DELETE from DealLogs WHERE DealUUID IN (SELECT DISTINCT DealUUID FROM DealLogs WHERE CreatedAt < ?)"
 
 	_, err := d.db.ExecContext(ctx, qry, td)
+	if err != nil {
+		return err
+	}
+
+	// Vacuum DealLogs to reduce the log file size. This OP requires free space of "2 * current file size" to succeed
+	qry = "VACUUM DealLogs"
+	_, err = d.db.ExecContext(ctx, qry, td)
```
dirkmc, line 81: I don't think you need to provide the `td` parameter here.

dirkmc (commented):
Please test this out locally before merging.

dirkmc (commented):
I'd suggest we don't actually error out if the vacuum command fails. Let's just log an error.

LexLuthr:
I have updated to just log the error instead of erroring out.

dirkmc (approved):

Issue #1053, opened 2023-01-05
Implement piece directory NextPiecesToCheck leveldb implementation


dirkmc:
Implement the remaining tasks in https://github.com/filecoin-project/boost/pull/1045 and uncomment `testNextPieces` in `doctor_test.go`

dirkmc:
Fixed by https://github.com/filecoin-project/boost/pull/1062

PR #1062, opened 2023-01-09
implement NextPiecesToCheck for LevelDB

nonsense:
Fixes https://github.com/filecoin-project/boost/issues/1053

---

This PR is adding a basic implementation of the `NextPiecesToCheck` for the LevelDB adapter.

### DIFF
--- a/extern/boostd-data/couchbase/db.go
+++ b/extern/boostd-data/couchbase/db.go
@@ -726,7 +726,7 @@ func (db *DB) NextPiecesToCheck(ctx context.Context) ([]cid.Cid, error) {
 }
 
 // The minimum frequency with which to check pieces for errors (eg bad index)
-const MinPieceCheckPeriod = time.Second
+var MinPieceCheckPeriod = 30 * time.Second
 
 // Work out how frequently to check each piece, based on how many pieces
 // there are: if there are many pieces, each piece will be checked
--- a/piecedirectory/doctor_test.go
+++ b/piecedirectory/doctor_test.go
@@ -7,13 +7,16 @@ import (
 	"testing"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/filecoin-project/boostd-data/client"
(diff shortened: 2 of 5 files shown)


extern/boostd-data/couchbase/db.go
```diff
@@ -726,7 +726,7 @@ func (db *DB) NextPiecesToCheck(ctx context.Context) ([]cid.Cid, error) {
 }
 
 // The minimum frequency with which to check pieces for errors (eg bad index)
-const MinPieceCheckPeriod = time.Second
+var MinPieceCheckPeriod = 30 * time.Second
```
nonsense, line 729: Bumping up to 30 seconds, as we don't want to check pieces every second - even 30 seconds is probably aggressive in the Filecoin realm.

Changing to `var` so that we can update this value in tests.

piecedirectory/doctor_test.go
```diff
@@ -44,56 +64,166 @@ func testPieceDoctor(ctx context.Context, t *testing.T, bdsvc *svc.Service, port
...
-
-	// TODO: uncomment once the leveldb implementation is complete
-	//t.Run("next pieces", func(t *testing.T) {
-	//	testNextPieces(ctx, t, cl, pieceCheckPeriod)
-	//})
 }
 
 // Verify that after a new piece is added
 // - NextPiecesToCheck immediately returns the piece
 // - NextPiecesToCheck returns the piece every pieceCheckPeriod
```
nonsense, line 82: @dirkmc I think it is hard to implement this in LevelDB, unless we add book-keeping for any new piece, which I'd like to avoid.

New pieces are added in a sorted table in LevelDB. What I've implemented is in-memory offset tracking and the doctor going batch by batch over all entries in the table. This is slightly different to the Couchbase implementation, but I think it is good enough. WDYT?

dirkmc: Yes I think it's fine for them to have different implementations 👍

dirkmc (approved):

PR #1063, opened 2023-01-10
fix panic in index provider

dirkmc:


nonsense (approved):

indexprovider/wrapper.go
```diff
@@ -198,7 +199,11 @@ func (w *Wrapper) AnnounceBoostDeal(ctx context.Context, pds *types.ProviderDeal
 		return cid.Undef, fmt.Errorf("failed to get proposal cid from deal: %w", err)
 	}
 
-	annCid, err := w.prov.NotifyPut(ctx, propCid.Bytes(), fm)
+	addrInfo := &peer.AddrInfo{
```
masih, line 202: No need to set this every time; it is inferred from the defaults configured at engine instantiation [if](https://github.com/ipni/index-provider/blob/main/engine/engine.go#L362-L365) `nil`.

Issue #1047, opened 2022-12-19
BUG: Offline deals doesn't trigger Deal Filter


s0nik42:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature or enhancement request. If it is, please open a [new idea discussion](https://github.com/filecoin-project/boost/discussions/new?category=ideas) instead. New feature and enhancement requests would be entertained by the boost team after a thorough discussion only.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to boost.

### Boost component

- [X] boost daemon - storage providers
- [ ] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### Boost Version

```text
boost --version
boost version 1.5.1-rc2+git.2c5604e
```

### Describe the Bug

A new behavior has been introduced by this commit (not sure if its a bug or intedted): 

https://github.com/filecoin-project/boost/commit/6fc504212c7b86fdb03d8cdc2e2f32fb3648d26d#diff-8d035260e0e563b27f9c0e1618df37accd637edeb18024932ebcbd5256614d7eR404-R415
 
With this new code, offline deals bypass the deal filter. Which is not the expected behavior. 

We expect offline deals to be treated as any types of deals (http/graphsync) and going through the deal filter.

### Logging Information

```text
.
```

### Repo Steps

1. Run '...'
2. Do '...'
3. See error '...'
...

PR #1067, opened 2023-01-11
fix: run deal filter for offline deals

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1047


dirkmc (approved):

s0nik42:
I confirm that it works as expected with CIDgravity

PR #1070, opened 2023-01-11
Remove bitswap peer/bandwidth filters

hannahhoward:
# Goals

Remove bandwidth and peer limit functionality

# Implementation

Thought we'd be able to just revert a commit, but it was actually too tied up in code for other PRs, so I just removed the relevant code.

The current structure of the API is as follows (cc @s0nik42):

```javascript
{
	"AllowDenyList": {
			"PeerIDs": ["Qma9T5YraSnpRDZqRR4krcSJabThc8nwZuJV3LercPHufi", "QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N"]
	},
        "UnderMaintenance": true,
}
```

We will put  back in storage provider limits when we support simultaneous connections. Any extraneous values in an API response should not prevent it from functioning.


cmd/booster-bitswap/filters/configfilter.go
```diff
@@ -21,44 +19,29 @@ const AllowList PeerListType = "allowlist"
...
-type RequestCounter interface {
-	StateForPeer(p peer.ID) requestcounter.ServerState
-}
-
 type remoteConfig struct {
 	peerListType                   PeerListType
 	peerList                       map[peer.ID]struct{}
 	underMaintenance               bool
 	maxSimultaneousRequests        uint64
 	maxSimultaneousRequestsPerPeer uint64
```
dirkmc, line 27: We can also remove all the max simultaneous requests code now, right?

Issue #1014, opened 2022-11-29
Add filter by Checkpoint


dirkmc:
Currently it's possible to filter the list of deals by a search query, that matches against peer id, payload cid etc.
However it's not possible to filter by Checkpoint.

We should add an optional parameter to the deals endpoint to allow users to filter by checkpoint:
```
deals(query: String, checkpoint: String, cursor: ID, offset: Int, limit: Int): DealList!
```

jacobheun:
The UI currently shows the 'State' rather than the Checkpoint, so based on a sync discussion we'll aim to expose the ability to filter on `State` rather than checkpoint. This will also give more granularity for later stages of deals.

jacobheun:
After reviewing the internals of the code here, State is a computed property so for simplicity of this filtering by Checkpoint is the most direct path. Pulling some feedback from slack some requests to be able to filter by are:

* Filter all deals that are currently transferring, or waiting for offline import
* Filter for all deals that have been Accepted (and that are offline deals)

I'm reviewing the UX for the api and UI and will update when I have more thoughts on best path forward for these.

PR #1097, opened 2023-01-19
Add the ability to filter certain properties in the deal list

jacobheun:
## Purpose
Adds the ability to include filters when search for deals in addition to the existing query functionality. This should address all scenarios mentioned in https://github.com/filecoin-project/boost/issues/1014. 

### Example curls:
**Get all offline deals pending transfer**
```sh
curl -s -X POST -H "Content-Type: application/json" -d '{"query":"query { deals (filter: { Checkpoint: Accepted, IsOffline: true }) { deals { ID  Message} } }"}' http://localhost:8080/graphql/query
```

**Get all deals currently in the sealing pipeline**
```sh
curl -s -X POST -H "Content-Type: application/json" -d '{"query":"query { deals (filter: { Checkpoint: IndexedAndAnnounced }) { deals { ID  Message} } }"}' http://localhost:8080/graphql/query
```

### Current Supported Properties
I've started with a subset of properties to filter by based on common scenarios. VerifiedDeal may be less useful atm, but there may be some useful applications there outside of the requested scenarios.

* Checkpoint
* IsOffline
* TransferType (Not currently supported in the UI as we havent had requests)
* IsVerified (Not currently supported in the UI as we havent had requests)

## Work remaining
- [X] Hook up the graphql resolver to pass filters from the client
- [x] Fix the checkpoint filter so it's optional
- [x] Add support to the UI to select the filters

**Disclaimer**: Still getting up to speed on go, so please suggest more idiomatic usage where applicable. 

Resolves https://github.com/filecoin-project/boost/issues/1014

## Demos
**Querying with filters via the GQL UI**

https://user-images.githubusercontent.com/639834/214308975-6e8952dc-a98b-4b9b-bce8-d38821817b4c.mp4

**Querying with filters via the Web UI**

https://user-images.githubusercontent.com/639834/214659019-6c527fe9-2bb8-467a-b521-f91ac44170f7.mp4

### DIFF
--- a/db/deals.go
+++ b/db/deals.go
@@ -229,13 +229,25 @@ func (d *DealsDB) BySignedProposalCID(ctx context.Context, proposalCid cid.Cid) 
 	return d.scanRow(row)
 }
 
-func (d *DealsDB) Count(ctx context.Context, query string) (int, error) {
+func (d *DealsDB) Count(ctx context.Context, query string, filter map[string]interface{}) (int, error) {
 	whereArgs := []interface{}{}
 	where := "SELECT count(*) FROM Deals"
 	if query != "" {
 		searchWhere, searchArgs := withSearchQuery(query)
 		where += " WHERE " + searchWhere
 		whereArgs = append(whereArgs, searchArgs...)
+	}
+
+	if filter != nil {
+		filterWhere, filterArgs := withSearchFilter(filter)
+
+		if query != "" {
+			where += " AND "
+		} else {
+			where += " WHERE "
+		}
+		where += filterWhere
+		whereArgs = append(whereArgs, filterArgs...)
 	}
 	row := d.db.QueryRowContext(ctx, where, whereArgs...)
 
@@ -252,7 +264,7 @@ func (d *DealsDB) ListCompleted(ctx context.Context) ([]*types.ProviderDealState
 	return d.list(ctx, 0, 0, "Checkpoint = ?", dealcheckpoints.Complete.String())
 }
 
-func (d *DealsDB) List(ctx context.Context, query string, cursor *graphql.ID, offset int, limit int) ([]*types.ProviderDealState, error) {
+func (d *DealsDB) List(ctx context.Context, query string, filter map[string]interface{}, cursor *graphql.ID, offset int, limit int) ([]*types.ProviderDealState, error) {
 	where := ""
 	whereArgs := []interface{}{}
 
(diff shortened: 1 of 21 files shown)


db/deals.go
```diff
@@ -272,9 +284,40 @@ func (d *DealsDB) List(ctx context.Context, query string, cursor *graphql.ID, of
...
+		// If the filterField is in the filter and it's not empty, append
+		value, ok := filter[filterField]
+		if ok {
+			st := filterField + " = ?"
+			statements = append(statements, st)
+			whereArgs = append(whereArgs, value)
+		}
+	}
+
+	where := "(" + strings.Join(statements, " AND ") + ")"
```
dirkmc, line 316: We may want to return "" for the case that there are no fields in the filter

gql/resolver.go
```diff
@@ -134,7 +142,22 @@ func (r *resolver) Deals(ctx context.Context, args dealsArgs) (*dealListResolver
...
+	}
+	if args.Filter.IsOffline.Set && args.Filter.IsOffline.Value != nil {
+		filter["IsOffline"] = args.Filter.IsOffline.Value
+	}
+	if args.Filter.TransferType.Set && args.Filter.TransferType.Value != nil {
+		filter["TransferType"] = args.Filter.TransferType.Value
+	}
+	if args.Filter.VerifiedDeal.Set && args.Filter.VerifiedDeal.Value != nil {
+		filter["VerifiedDeal"] = args.Filter.VerifiedDeal.Value
+	}
```
dirkmc, line 158: I think the code may end up simpler if we add a FilterOptions struct as a parameter to DealList at the db layer that looks similar to filterArgs (but where each field is a pointer so that it can be nil). That way we also don't expose the db fields at the resolver layer.

db/deals.go
```diff
@@ -229,14 +236,26 @@ func (d *DealsDB) BySignedProposalCID(ctx context.Context, proposalCid cid.Cid)
...
 	if query != "" {
 		searchWhere, searchArgs := withSearchQuery(query)
 		where += " WHERE " + searchWhere
 		whereArgs = append(whereArgs, searchArgs...)
 	}
+
+	if filter != nil {
+		filterWhere, filterArgs := withSearchFilter(*filter)
+
+		if query != "" {
```
dirkmc, line 251: I think we need to check if filterWhere == "" also

db/deals.go
```diff
@@ -272,9 +291,51 @@ func (d *DealsDB) List(ctx context.Context, query string, cursor *graphql.ID, of
 		whereArgs = append(whereArgs, searchArgs...)
 	}
 
+	if filter != nil {
+		if where != "" {
```
dirkmc, line 295: I think we need to check if filterWhere == "" also

dirkmc (approved):
Looks great! 🙌 
Just one nit to fix and then feel free to merge.

db/deals.go
jacobheun: I can add it to be safe but I don't think it should be possible to get to that state, as if filter is not nil it should always yield a valid query statement.

PR #1100, opened 2023-01-19
increase poll intervals

nonsense:


dirkmc:
Mostly looks good - the only two I'm not sure about:
- On the deal transfers page I think let's keep it to 1s. It's not an expensive call (all data is in memory) and it's nice to see the numbers updating frequently
- On the deal list page it's also nice to get frequent updates, particularly when deals are transferring. Maybe we can do some caching on the back-end if it's too expensive.

nonsense:
I reverted the deal transfers intervals.

I still think for the Deal List page 10sec. is probably OK. We generally have not more than 100 deals per day, so refreshing every second seems excessive.

dirkmc (approved):

PR #1101, opened 2023-01-20
Couchbase test using docker compose

airenas:
Runs tests from docker compose.
## Done
- [x] added docker compose file that starts couchbase container and go tests
- [x] refactored `svc/setup_couchbase_test_util.go`:
   - [x] moved to `couchbase/setup_test_util.go` - as the test here also needs cb, and can't access svc because cyclic deps
   - [x] dropped docker initialization code
   - [x] uses `COUCHBASE_HOST` env to access couchbase service, if not set then uses `localhost`
   - [x] did another refactoring for init clusters, now tries to check for dummy bucket before initializing in order not to fail on a second initialization
- [x] Unskipped tests:
  - [x] TestService/couchbase
  - [x] TestSharding
## Not done  
- [ ] Unskip
  - [ ] TestCleanup/couchbase - is failing, was failing before my changes also, but with a different error, need your response on it
  - [ ] TestServiceFuzz - failing also
- [ ] configure CI

 ## Running
 ### from CI or bash inside containers
 `docker compose up --build --exit-code-from go-tests`

 ### from ide
 `docker compose up couchbase` and run test from ide, test expects couchbase to be on localhost:8091
  ### from bash without container
 `docker compose up -d couchbase && go test -v --count=1 ./...`
  ### drop cached couchbase clusters
 `docker compose down`


nonsense:
Closing in favor of swapping Couchbase with YugabyteDB.

PR #1102, opened 2023-01-20
refactor piece directory -> local index directory

dirkmc:
For now I focused on the externally-facing information, like error messages, and command and parameter names.
Changing all the internal naming would be a lot more work and I'm not sure if it's worth it.

### DIFF
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -38,12 +38,12 @@ type Boost struct {
 	// The connect string for the sealing RPC API (lotus miner)
 	SealerApiInfo string
 	// The connect string for the sector index RPC API (lotus miner)
-	SectorIndexApiInfo string
-	Dealmaking         DealmakingConfig
-	Wallets            WalletsConfig
-	Graphql            GraphqlConfig
-	Tracing            TracingConfig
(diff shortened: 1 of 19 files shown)


node/config/types.go
```diff
@@ -261,34 +261,34 @@ type StorageConfig struct {
...
 }
 
-type PieceDirectoryCouchbaseBucketConfig struct {
+type LocalIndexDirectoryCouchbaseBucketConfig struct {
 	// Bucket setting RAMQuotaMB
 	RAMQuotaMB uint64
 }
 
-type PieceDirectoryCouchbaseConfig struct {
+type LocalIndexDirectoryCouchbaseConfig struct {
```
nonsense, line 269: We have to at least contact @TippyFlitsUK when we merge this, as we'll have to update config for all Boost nodes that are already running Piece Directory. I know we haven't released it yet, but there are at least 3 Boost nodes that are already on it.

nonsense (approved):

node/config/types.go
dirkmc: I don't think @TippyFlitsUK is actually running piece directory in production yet - he just ran the migration script

PR #1131, opened 2023-01-26
boost retrieve cmd

nonsense:
Add `boost retrieve` cmd, similar to how `lotus retrieve` works.

TODO:
- [x] retrieve over graphsync when price is set to 0
- [x] make sure we get an error when trying to retrieve over graphsync when price is set to > 0
- [x] make sure that trying to retrieve content that provider doesn't have returns a meaningful error
- [ ] make sure that trying to retrieve content that provider doesn't have doesn't hang


cmd/boost/retrieve_cmd.go
```diff
@@ -0,0 +1,468 @@
...
+			return fmt.Errorf("please specify a CID to retrieve")
+		}
+
+		dmSelText := textselector.Expression(cctx.String(flagDmPathSel.Name))
+
+		miners, err := parseMiners(cctx)
+		if err != nil {
+			return err
+		}
+
```
dirkmc, line 106: I think we can simplify by removing the multi-miner support

cmd/boost/retrieve_cmd.go
```diff
@@ -0,0 +1,468 @@
...
+		flagDmPathSel,
+		flagCar,
+	},
+	Action: func(cctx *cli.Context) error {
+
+		// Store config dir in metadata
+		ddir, err := homedir.Expand("~/.boost-client")
+		if err != nil {
+			fmt.Println("could not set config dir: ", err)
+		}
```
dirkmc, line 91: Do we also need to call `node.Setup` from cli/node/node.go like we do in boost deal_cmd.go?

cmd/boost/retrieve_cmd.go
```diff
@@ -0,0 +1,468 @@
...
+			Client:     fc,
+			Cid:        c,
+			Candidates: candidates,
+			SelNode:    selNode,
+		}}
+
+		stats, err := node.RetrieveFromBestCandidate(cctx.Context, networks)
+		if err != nil {
+			return err
+		}
```
dirkmc, line 193: I suggest we remove all the candidate stuff for simplicity.

cmd/boost/retrieve_cmd.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	return c, closer, nil
+}
+
+type Node struct {
+	Host host.Host
+
+	Datastore  datastore.Batching
+	DHT        *dht.IpfsDHT
```
dirkmc, line 350: Do we need a DHT?

cmd/boost/retrieve_cmd.go
```diff
@@ -0,0 +1,468 @@
...
+	Blockstore blockstore.Blockstore
+	Bitswap    *bitswap.Bitswap
```
dirkmc, line 352: I don't think we need Bitswap

cmd/boost/retrieve_cmd.go
```diff
@@ -0,0 +1,468 @@
...
+
+	Wallet *wallet.LocalWallet
+}
+
+func setup(ctx context.Context, cfgdir string) (*Node, error) {
+	peerkey, err := loadOrInitPeerKey(keyPath(cfgdir))
+	if err != nil {
+		return nil, err
+	}
+
```
dirkmc, line 362: Most of this can probably be replaced with node.Setup in cli/node/node.go

retrieve/libp2ptransfermanager.go
```diff
@@ -0,0 +1,432 @@
...
+	listener   func(uint, ChannelState)
+	unsub      func()
+}
+
+type libp2pTransferManagerOpts struct {
+	xferTimeout     time.Duration
+	authCheckPeriod time.Duration
+}
+
+func newLibp2pTransferManager(dtServer libp2pCarServer, ds datastore.Batching, authDB *httptransport.AuthTokenDB, opts libp2pTransferManagerOpts) *libp2pTransferManager {
```
dirkmc, line 52: I don't think we need libp2pTransferManager - it's used to manage incoming requests for data in storage deals

dirkmc (commented):
Looks good so far 👍 

I'd suggest we try to simplify a little more before merging

cmd/boost/retrieve_cmd.go
nonsense: I've added Blockstore and Datastore to the Node, although just Retrieve needs them. Let me know if you're okay with this... I thought it is a good idea because we do keep state for them in the repository.

nonsense:
@dirkmc I think I addressed all comments, and further simplified the code.

cli/node/node.go
```diff
@@ -55,9 +61,22 @@ func Setup(cfgdir string) (*Node, error) {
 		return nil, err
 	}
 
+	bstoreDatastore, err := flatfs.CreateOrOpen(blockstorePath(cfgdir), flatfs.NextToLast(3), false)
+	bstore := blockstore.NewBlockstoreNoPrefix(bstoreDatastore)
+	if err != nil {
+		return nil, fmt.Errorf("blockstore could not be opened (it may be incompatible after an update - try running the clear blockstore subcommand to delete the blockstore and try again): %v", err)
+	}
```
dirkmc, line 68:
```suggestion
		return nil, fmt.Errorf("blockstore could not be opened: %w", err)
	}
```

cmd/boost/retrieve_cmd.go
```diff
@@ -0,0 +1,321 @@
...
+var flagDmPathSel = &cli.StringFlag{
+	Name:  "datamodel-path-selector",
+	Usage: "a rudimentary (DM-level-only) text-path selector, allowing for sub-selection within a deal",
+}
+
+var retrieveCmd = &cli.Command{
+	Name:        "retrieve",
+	Usage:       "Retrieve a file by CID from a miner",
+	Description: "Retrieve a file by CID from a miner. If desired, multiple miners can be specified as fallbacks in case of a failure (comma-separated, no spaces).",
```
dirkmc, line 65:
```suggestion
```

retrieve/rep/event.go
```diff
@@ -0,0 +1,193 @@
+package rep
```
dirkmc, line 1: I don't think we're actually subscribing to these events anywhere so we can probably just delete this entire rep package

retrieve/retrieve.go
```diff
@@ -0,0 +1,1091 @@
...
+
+	logRetrievalProgressEvents bool
+}
+
+type GetPieceCommFunc func(ctx context.Context, payloadCid cid.Cid, bstore blockstore.Blockstore) (cid.Cid, uint64, abi.UnpaddedPieceSize, error)
+
+type Lp2pDTConfig struct {
+	Server          httptransport.ServerConfig
+	TransferTimeout time.Duration
+}
```
dirkmc, line 81: I believe Lp2pDTConfig is for storage deals, so we can probably remove

retrieve/retrieve.go
```diff
@@ -0,0 +1,1091 @@
...
+		return fmt.Errorf("failed to read response: %w", err)
+	}
+
+	return nil
+}
+
+type TransferType string
+
+const (
+	BoostTransfer     TransferType = "boost"
```
dirkmc, line 382: Looks like BoostTransfer is unused

retrieve/retrieve.go
```diff
@@ -0,0 +1,1091 @@
...
+	GraphsyncTransfer TransferType = "graphsync"
+)
+
+type ChannelState struct {
```
dirkmc, line 386: ChannelState, TransferStatus() and ChannelStateConv() are unused

retrieve/retrieve.go
```diff
@@ -0,0 +1,1091 @@
...
+		MarketAvailable:       types.FIL(avail),
+		VerifiedClientBalance: vcstatus,
+	}, nil
+}
+
+type LockFundsResp struct {
+	MsgCid cid.Cid
+}
+
+func (c *Client) CheckChainDeal(ctx context.Context, dealid abi.DealID) (bool, *api.MarketDeal, error) {
```
dirkmc, line 517: Unused

retrieve/retrieve.go
```diff
@@ -0,0 +1,1091 @@
...
+			"queryResponse:UnsealPrice", tevent.QueryResponse().UnsealPrice)
+	case rep.RetrievalEventFailure:
+		logadd("errorMessage", tevent.ErrorMessage())
+	case rep.RetrievalEventSuccess:
+		logadd("receivedSize", tevent.ReceivedSize())
+	}
+	log.Debugw("retrieval-event", kv...)
+}
+
+func ChannelIDFromString(id string) (*datatransfer.ChannelID, error) {
```
dirkmc, line 1036: Unused

dirkmc (commented):
Awesome! Just about there, let's remove a few more of these unused methods.

I wonder if we should delete everything from the blockstore after use? Otherwise it's just going to keep growing, and we'd need to add commands to manage it.

retrieve/retrieve.go
nonsense: Wonder why the linter is not catching this? It does catch "some" unused methods.

dirkmc:
Also should we move this under the retrievalmarket folder? Maybe retrievalmarket/client

retrieve/retrieve.go
```diff
@@ -0,0 +1,1091 @@
...
+		return nil, err
+	}
+	minerOwner, err := c.minerOwner(ctx, miner)
+	if err != nil {
+		return nil, err
+	}
+	return c.RetrieveContentFromPeerWithProgressCallback(ctx, minerPeer.ID, minerOwner, proposal, progressCallback)
+}
+
+func (c *Client) RetrieveContentFromPeerWithProgressCallback(
```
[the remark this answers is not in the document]
nonsense: This is actually used.

retrieve/rep/event.go
nonsense: I think we do for the query ask. I tried to remove it once and I failed, I could give it another shot.

nonsense: Removed rep.

dirkmc (approved):

PR #1140, opened 2023-01-27
update jsonrpc to v0.2.0

dirkmc:
TODO:
- [ ] Test in production


nonsense (approved):

nonsense:
Bumped to 0.2.1 as it is needed in the Lotus API, so we might as well test directly 0.2.1

PR #1144, opened 2023-01-30
Use more efficient encoding for couchbase indices

dirkmc:
Changes
- multihash -> piece cid index
  - Changes the multihash encoding from hex to base85 (not binary because couchbase keys must be utf8 conformant)
  - Changes the list of piece cids to be binary
- piece offsets map (piece cid -> map of multihash -> offset / length)
  - Changes the multihash encoding from hex to base64
  - Note that the map itself must not be in binary format, because we need to be able to get an individual key from the map, and in couchbase the key must be utf8-conformant.
  - Note also that we use base64 instead of base85 because the [couchbase path specification](https://docs.couchbase.com/go-sdk/current/howtos/subdocument-operations.html#path-syntax) has a more restricted character set than the [couchbase key specification](https://docs.couchbase.com/server/current/learn/data/data.html#keys)

Note that I also tried using base122 encoding but it violated the [couchbase key specification](https://docs.couchbase.com/server/current/learn/data/data.html#keys) so I switched to base85.

Comparison of indices in dagstore vs couchbase:

couchbase-index-sizes

```
$ du -h ~/.boost/dagstore 
2,0G	/root/.boost/dagstore/index            # couchbase: piece-dir.piece-offsets     2.63GiB
4,0K	/root/.boost/dagstore/transients
5,1G	/root/.boost/dagstore/datastore        # couchbase: piece-dir.mh-to-pieces      4.69GiB
7,1G	/root/.boost/dagstore    
```

Indices in couchbase before this PR was applied:

couchbase-compression-ps

TODO:
- [x] Test on local devnet
- [x] Test on a production miner


nonsense (approved):

dirkmc:
TODO:
- [ ] Use last 24 bytes for multihash keys
  - Collision probability (birthday problem): 2^(192/2)
- [ ] After reducing the multihash key length, check the size of the `map[multihash] -> offset/size` values in couchbase. If they are small enough we may be able to reduce the shard key to 1 byte.
- [ ] Try using base85 encoding instead of base64
- [ ] Add a Version field to the piece metadata
- [ ] Remove "u:" prefix from keys

jacobheun:
What's the latest diff in size from the PR description snapshot?

dirkmc:
I'm running it now, I'll post results in a about an hour

dirkmc:
I added updated screenshots to the description and the [comment describing the most recent set of changes](https://github.com/filecoin-project/boost/pull/1144#issuecomment-1410481883)

PR #1152, opened 2023-02-01
chore: remove dead code

ribasushi:


jacobheun (approved):
Dead indeed! usage was removed in https://github.com/filecoin-project/boost/issues/960.

Issue #1084, opened 2023-01-16
docker devnet -- M1 compatibility and lotus update


LaurenSpiegel:
@airenas and @nonsense , would appreciate your input on these:

1) Lotus's dockerfile has been updated. Is there anything we should do in our setup? See  https://github.com/filecoin-project/lotus/pull/9625

2) @hannahhoward was unable to get the docker to build on an M1 mac with either the make docker/all command or the make docker/all ffi_from_source=1

Error:
unknown-linux-gnu/lib/rustlib/aarch64-unknown-linux-gnu/lib/libcompiler_builtins-6421404651457840.rlib" "-Wl,-Bdynamic" "-lhwloc" "-lOpenCL" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-L" "/usr/local/rustup/toolchains/nightly-2022-05-09-aarch64-unknown-linux-gnu/lib/rustlib/aarch64-unknown-linux-gnu/lib" "-o" "/opt/filecoin/extern/filecoin-ffi/rust/target/debug/deps/filcrypto-bdd4d906adefd5ff" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro,-znow" "-nodefaultlibs"
#12 445.0   = note: collect2: fatal error: ld terminated with signal 9 [Killed]
#12 445.0           compilation terminated.
#12 445.0           
#12 445.0 
#12 445.3 error: could not compile `filcrypto` due to previous error
#12 445.7 + rm -f /tmp/tmp.wqNf8h8H48
#12 445.7 make[1]: *** [Makefile:14: .install-filcrypto] Error 101
#12 445.8 make[1]: Leaving directory '/opt/filecoin/extern/filecoin-ffi'
#12 445.8 make: *** [Makefile:41: build/.filecoin-install] Error 2

jacobheun:
Yeah we need to update our setup, Lotus has consolidated some images so we need to update for that.

jacobheun:
Looks like there's an issue with M1s in the latest lotus setup https://github.com/filecoin-project/lotus/issues/9821

airenas:
> 2\. @hannahhoward was unable to get the docker to build on an M1 mac with either the make docker/all command or the make docker/all ffi_from_source=1

[...]

PR #1157, opened 2023-02-03
Devnet images using Lotus from DockerHub

airenas:
Implemented as requested in  (#1084).
Now building images the first time for devnet should be twice faster.

### Changes
- tested with an official lotus image: [v1.19.0-debug](https://hub.docker.com/layers/filecoin/lotus-all-in-one/v1.19.0-debug/images/sha256-295f057ad0fbae829b168592d486c91e801bcf053e9be95f217869acbb5b8d44?context=explore). The image works out of the box, without any other changes
- updated Makefile to use the official lotus docker image instead of building locally as the default case
- ~~refactored makefile to use a new lotus dockerfile~~, but rollbacked. The tag 1.19.0 (default now) of the lotus repo does not contain the new Dockerfile, so the building did not work out-of-the-box
- updated readme

#### with the official lotus image 
Works only for amd64 platforms
```bash
make docker/all
# with another image, e.g.: filecoin/lotus-all-in-one:nightly-debug
make docker/all lotus_version=nightly
```
#### with specific version of lotus
```bash
make docker/all lotus_version=v1.17.0 build_lotus=1
# for M1, M2
make docker/all lotus_version=v1.17.0 build_lotus=1 ffi_from_source=1
```
#### from lotus source
```bash
make docker/all lotus_src_dir=../lotus
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,36 +200,53 @@ docsgen-openrpc-boost: docsgen-openrpc-bin
 
 ## DOCKER IMAGES
 docker_user?=filecoin
-lotus_version?=1.19.0-rc1
+lotus_version?=v1.19.0
 lotus_src_dir?=
 ffi_from_source?=0
+build_lotus?=0
 ifeq ($(lotus_src_dir),)
-    lotus_src_dir=/tmp/lotus-$(lotus_version)
-    lotus_checkout_dir=$(lotus_src_dir)
+	ifeq ($(build_lotus),1)
+# v1: building lotus image with provided lotus version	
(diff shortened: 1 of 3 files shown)


jacobheun:
I was able to do the make build with the latest 1.20.0-rc1 from lotus and that was much quicker (I'm using AMD). When I ran `docker compose up -d` everything got stuck in `Starting`. I'll try this again later today, might have just been an issue with my setup.

jacobheun (approved):
Looks good, the issue I was having earlier I believe was due to recent changes in the lotus 1.20.0-rc1. I was able to run this against 1.19.0 without any issues.

Issue #1142, opened 2023-01-30
Improve resiliency of booster-bitswap denylist fetch


jacobheun:
A user reported in Slack that booster-bitswap failed to launch on initial startup. The underlying issue was that the badbits denylist which is automatically downloaded from the badbits website was truncated, presumably from an interrupted connection, resulting in a malformed json file. This prevents booster-bitswap from starting until the file is fixed / downloaded.

## Error log
```
2023-01-27T10:23:51.865-0800    INFO    booster booster-bitswap/run.go:187      Using boost API at ws://10.32.32.20:1288/rpc/v0
Error: starting block filter: parsing denylist: unexpected EOF
```

![Image of truncated denylist from user reporting](https://user-images.githubusercontent.com/639834/215457367-fb2e9620-73d6-4de2-99e0-c81baa8c3a55.png)

## Workaround
If the file was downloaded locally into the booster-bitswap repo, ie `~/.booster-bitswap/denylist[i].json`, it can be deleted as the startup command uses this cached file if it exists, otherwise it will reattempt a download.

PR #1159, opened 2023-02-06
Fetch filters from remote if local cache is invalid

ischasny:
On startup immediately refresh local caches from remote if they are invalid. That might happen due to a previously broken connection that has prevented the process from starting up.

Fixes https://github.com/filecoin-project/boost/issues/1142

### DIFF
--- a/cmd/booster-bitswap/filters/filters.go
+++ b/cmd/booster-bitswap/filters/filters.go
@@ -93,12 +93,20 @@ func (f *filter) update() error {
 		return nil
 	}
 	defer stream.Close()
-	// open the cache file
-	cache, err := os.OpenFile(f.CacheFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	// write cache to the temp file first
+	tmpFile := f.CacheFile + ".tmp"
+	cache, err := os.OpenFile(tmpFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return fmt.Errorf("opening cache file: %w", err)
 	}
-	defer cache.Close()
+	defer func() {
+		// rename the temp file only if the operation succeeds
+		if cache.Close() == nil && err == nil {
+			_ = os.Rename(tmpFile, f.CacheFile)
+		} else {
+			_ = os.Remove(tmpFile)
+		}
+	}()
 	forkedStream := io.TeeReader(stream, cache)
 	f.lastUpdated = fetchTime
 	err = f.Handler.ParseUpdate(forkedStream)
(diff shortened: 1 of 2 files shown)


dirkmc:
Thanks a lot for diving in to fix this bug!

I wonder if it would make sense to simply not write to the cache file until after it has successfully been parsed? Then we don't need to worry about adding extra logic in the `Start()` method.

ischasny:
> Thanks a lot for diving in to fix this bug!
> 
> I wonder if it would make sense to simply not write to the cache file until after it has successfully been parsed? Then we don't need to worry about adding extra logic in the `Start()` method.

As I understand it was a deliberate implementation choice to write [alongside](https://github.com/filecoin-project/boost/blob/main/cmd/booster-bitswap/filters/filters.go#L102) reading as the denylist might be quite large (let me know if I misunderstood the design intent). I can change to write in bulk but then there will be memory implication.

dirkmc:
Good point - I wonder if we could write to a temp file, then rename it to the cache file on success?

ischasny:
Good point - will add that soon. Would you not want to still verify cache correctness on start though? So that the process can self recover for example if data format is going to change in the future?

ischasny:
> I think I'd rather keep the caching logic in one place, and then if we change formats in future we can add a migration stage at startup

Cool - pushed the updates in

cmd/booster-bitswap/filters/filters.go
```diff
@@ -93,12 +93,21 @@ func (f *filter) update() error {
...
-	cache, err := os.OpenFile(f.CacheFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	// write cache to the temp file first
+	tmpFile := f.CacheFile + ".tmp"
+	cache, err := os.OpenFile(tmpFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return fmt.Errorf("opening cache file: %w", err)
 	}
-	defer cache.Close()
+	defer func() {
+		cache.Close()
```
dirkmc, line 103: Should we check the return value of cache.Close() also?

dirkmc (commented):
Looks great!
Just one question then I think we're ready for merge.

cmd/booster-bitswap/filters/filters.go
ischasny: Great spot, fixed.

dirkmc:
hmm looks like `TestMultiFilterCacheShouldNotBeSavedIfDataWasInvalid` is now failing 🤔

ischasny:
> hmm looks like `TestMultiFilterCacheShouldNotBeSavedIfDataWasInvalid` is now failing 🤔

Yes, on it - small mistake

dirkmc (approved):
Looks great, thanks Ivan! ✅

PR #1167, opened 2023-02-08
feat: use an expiry lru cache for the index backed blockstore

jacobheun:
## Summary
This switches over the dagstore index backed blockstore to use an expiry based LRU cache instead of just an LRU. The existing LRU will never prune unless it fills up and evicts. With the new cache, expired entries will automatically be removed, and in the event the cache hits its max size, the shards closest to expiry will be pruned first.

* Adds a config option, `DealmakingConfig.BlockstoreCacheMaxShards` to set the max shards in the cache. The previous value was hard coded at `100`, which the default has been set to, but we may want to lower this as it can cause process memory to grow quite large. Too low can cause in progress retrievals to fail as shards in use can be pruned.
  * EDIT: reduced to `20` to match default simultaneous retrieval limits
* Adds a config option, `DealmakingConfig.BlockstoreCacheExpiry` to set the duration a cached shard can be idle for until pruned, default is `10 minutes`. Any time the cache is accessed it will be refreshed, so this could be set much lower if memory usage is a problem.
  * EDIT: reduced to `30 seconds`

### Memory Usage Before
Prior to the change with retrieval load on the system, followed by allowing the system to idle, heap space increases and is not reclaimed. 

Looking at a diff of the heap space allocation we can see an increase in the piece reader memory pipeline despite the system being idle.

### Memory Usage After
With these changes, allowing the system to idle after retrievals we can see in use heap decrease back to pre load levels.

Performing a diff of heap utilization we can see no significant markers of memory increase...

... and a clear reduction in piece reader heap between time in load and idle time

## TODO
- [x] fix ci
- [x] bump dagstore version once released

## Dependencies
- [x] requires https://github.com/filecoin-project/dagstore/pull/150

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -232,6 +232,7 @@ require (
 	github.com/jackpal/go-nat-pmp v1.0.2 // indirect
 	github.com/jbenet/go-temp-err-catcher v0.1.0 // indirect
 	github.com/jbenet/goprocess v0.1.4 // indirect
+	github.com/jellydator/ttlcache/v2 v2.11.1 // indirect
 	github.com/jessevdk/go-flags v1.4.0 // indirect
 	github.com/joeshaw/multierror v0.0.0-20140124173710-69b34d4ec901 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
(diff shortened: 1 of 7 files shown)


dirkmc (approved):
Looks good 👍 
Feel free to merge once the dagstore release is ready.

PR #1183, opened 2023-02-14
initial boost / fvm contract deals support

nonsense:
This PR is adding a chain monitor for the `DealProposalCreate` event as described at:
- https://www.notion.so/pl-strflt/WIP-Deal-Client-Contract-FRC-458e625f13b14c70bfdfe7ed64007b6c
- https://github.com/filecoin-project/FIPs/discussions/604

Given that this is not a final FRC, this functionality might change in the near future.

---

By default the monitor is disabled, users should enable it by updating a few values in `config.toml`

---

Based on `DealClient` solidity contract at: https://github.com/nonsense/fevm-hardhat-kit/blob/nonsense/deal-client-contract/contracts/basic-deal-client/DealClient.sol

---

MUST FIX:
- [ ] fix/review `client` mismatch

NICE TO HAVE:
- [ ] fix CBOR generation task
- [ ] once the FRC is somewhat stable add an itest


cmd/boostd/run.go
```diff
@@ -43,7 +46,8 @@ var runCmd = &cli.Command{
 			}()
 		}
 
-		fullnodeApi, ncloser, err := lcli.GetFullNodeAPIV1(cctx)
+		subCh := gateway.NewEthSubHandler()
```
dirkmc, line 49: Is there any performance implication of subscribing to these events? Wondering if we should check if fvm deals are enabled in config before subscribing.

storagemarket/contract_deal_monitor.go
```diff
@@ -0,0 +1,269 @@
...
+						//TODO: maybe add to pv1?? RemoveUnsealedCopy: paramsAndVersion.RemoveUnsealedCopy,
+						RemoveUnsealedCopy: false,
+						SkipIPNIAnnounce:   pv1.SkipIpniAnnounce,
+					}
+
+					log.Infow("received contract deal proposal", "id", proposal.DealUUID, "client-peer", dpc.Client)
+
+					reason, err := c.prov.ExecuteLibp2pDeal(context.Background(), &proposal, "")
+					if err != nil {
+						log.Warnw("contract deal proposal failed", "id", proposal.DealUUID, "err", err, "reason", reason.Reason)
```
dirkmc, line 200: Should we return in this if statement?

storagemarket/contract_deal_monitor.go
```diff
@@ -0,0 +1,269 @@
...
+							Size:   pv1.CarSize,
+						},
+						//TODO: maybe add to pv1?? RemoveUnsealedCopy: paramsAndVersion.RemoveUnsealedCopy,
+						RemoveUnsealedCopy: false,
+						SkipIPNIAnnounce:   pv1.SkipIpniAnnounce,
+					}
+
+					log.Infow("received contract deal proposal", "id", proposal.DealUUID, "client-peer", dpc.Client)
+
+					reason, err := c.prov.ExecuteLibp2pDeal(context.Background(), &proposal, "")
```
dirkmc, line 198: nit: We can probably leave the name as ExecuteDeal as it's called from the contract deal monitor path as well as from libp2p

storagemarket/types/contract_deal_proposal_types.go
```diff
@@ -0,0 +1,33 @@
...
+	StoragePricePerEpoch abi.TokenAmount
+
+	ProviderCollateral abi.TokenAmount
+	ClientCollateral   abi.TokenAmount
+
+	Version string
+	Params  []byte
+}
+
+type ParamsVersion1 struct {
```
dirkmc, line 29: nit: if ParamsVersion1 is specific to contracts, suggest we prefix it with Contract (or similar) to distinguish it from DealParams (used in libp2p deal proposals)

dirkmc (commented):
Really amazing work! 🙌

cmd/boostd/run.go
nonsense: I don't think there is any performance implication, I think this is just creating the handler in memory, but the actual subscription happens elsewhere.

storagemarket/contract_deal_monitor.go
```diff
@@ -0,0 +1,269 @@
...
+							},
+						},
+						DealDataRoot: rootCid,
+						Transfer: types.Transfer{
+							Type:   "http",
+							Params: []byte(fmt.Sprintf(`{"URL":"%s"}`, pv1.LocationRef)),
+							Size:   pv1.CarSize,
+						},
+						//TODO: maybe add to pv1?? RemoveUnsealedCopy: paramsAndVersion.RemoveUnsealedCopy,
+						RemoveUnsealedCopy: false,
```
nonsense, line 192: @dirkmc we should add `RemoveUnsealedCopy` into `ParamsVersion1`, right?

dirkmc: Yes I think that makes sense 👍

PR #1185, opened 2023-02-15
Introduce deal proposal protocol v1.2.1

### LINKED ISSUE #1184: Request to add client and server logic for protocol version awareness
Work was completed to add deal sealing and indexer announcement configurability behaviors here: 
https://github.com/filecoin-project/boost/pull/1051#top 

in support of this FRC: 
https://github.com/filecoin-project/FIPs/blob/f30244781db571d3b0420f414be28e6d84cfa1f9/FRCs/frc-0056.md

However, in order for this functionality to actually work and be usable we need the client and server to recognize the protocol version (including the new version v1.2.1).

The desirable behavior is that the protocol will default to the earliest version either the client or server supports.

dirkmc:
Fixes #1184

The new deal proposal protocol supports two extra fields as described in
https://github.com/filecoin-project/FIPs/blob/f30244781db571d3b0420f414be28e6d84cfa1f9/FRCs/frc-0056.md

### DIFF
--- a/storagemarket/deal_execution.go
+++ b/storagemarket/deal_execution.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"runtime/debug"
 	"time"
 
 	"github.com/filecoin-project/boost/storagemarket/types"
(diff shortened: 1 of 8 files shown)


jacobheun (approved):
Looks good, we should make sure to update the docs with 1.2.1 info https://boost.filecoin.io/boost-architecture/libp2p-protocols.

PR #1186, opened 2023-02-15
move markets code from lotus to boost

dirkmc:
Port https://github.com/filecoin-project/boost/pull/1085 to `main`


jacobheun (approved):
This should be good to go straight into main now, instead of the refactor branch. It doesn't look like there will be any conflicts changing the base branch over, but I'll let you handle that merge train 🚄

Issue #1207, opened 2023-02-21
Very slow offline deal acceptance on v1.5.1


hyunmoon:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature or enhancement request. If it is, please open a [new idea discussion](https://github.com/filecoin-project/boost/discussions/new?category=ideas) instead. New feature and enhancement requests would be entertained by the boost team after a thorough discussion only.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to boost.

### Boost component

- [X] boost daemon - storage providers
- [ ] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### Boost Version

```text
boostd --version
boostd version 1.5.1+git.5aadde7

lotus-miner --version
lotus-miner version 1.19.0+mainnet+git.64059ca87

```

### Describe the Bug

**How to reproduce:**

On boost v1.5.1, using boost client, send an offline deal to your own boost daemon node.
The deal will take a very long time to be accepted (or rejected).
For me, each deal took about 15 seconds to be processed.

**How I fixed it:**

1. I downgraded boost and boostd to v1.5.0.
2. I also changed ConfigVersion = 4 to ConfigVersion = 3 in config.toml to downgrade.

After that, I could see a few dozens offline deals getting processed per second.

### Logging Information

```text
-
```

### Repo Steps

-

hyunmoon:
I couldn't see any deals on v1.5.0 (lotus-miner not responding to an offline deal import)
So I upgraded to v1.5.1-rc5 and I get the following error:
`Error: creating node: starting node: starting storage provider: failed to list completed deals: scanning deal row: sql: Scan error on column index 6, name "FastRetrieval": sql/driver: couldn't convert <nil> (<nil>) into type bool`

hyunmoon:
Everything works fine on boost v1.5.0 and lotus v1.8.0

marco-storswift:
[...]

PR #1209, opened 2023-02-22
feat: cache sealing pipeline status

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1207

Suggested design by DIrk:

if there is no deal filter enabled in config, don't call the method at all

if the deal filter is enabled
> check the cache
> if the cache is empty
> > call the method
> > store the results in the cache
>> expire the result after x time (where x is configurable)
> return the result

### DIFF
--- a/storagemarket/provider.go
+++ b/storagemarket/provider.go
@@ -158,7 +158,7 @@ func NewProvider(cfg Config, sqldb *sql.DB, dealsDB *db.DealsDB, fundMgr *fundma
 	}
 
 	if cfg.SealingPipelineCacheTimeout < 0 {
-		cfg.SealingPipelineCacheTimeout = 30
+		cfg.SealingPipelineCacheTimeout = 30 * time.Second
 	}
 
 	return &Provider{
--- a/storagemarket/provider.go
+++ b/storagemarket/provider.go
@@ -157,8 +157,9 @@ func NewProvider(cfg Config, sqldb *sql.DB, dealsDB *db.DealsDB, fundMgr *fundma
 		cfg.MaxConcurrentLocalCommp = 1
 	}
 
-	// Set the timeout to 30 seconds
-	cfg.SealingPipelineCacheTimeout = 30
+	if cfg.SealingPipelineCacheTimeout < 0 {
+		cfg.SealingPipelineCacheTimeout = 30
+	}
 
 	return &Provider{
 		ctx:       ctx,
--- a/storagemarket/provider.go
+++ b/storagemarket/provider.go
@@ -69,7 +69,7 @@ type Config struct {
 	// Cleanup deal logs from DB older than this many number of days
 	DealLogDurationDays int
 	// Cache timeout for Sealing Pipeline status
-	SealingPipelineCacheTimeout int
+	SealingPipelineCacheTimeout time.Duration
 	StorageFilter               string
 }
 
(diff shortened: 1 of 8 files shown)


storagemarket/provider_loop.go
```diff
@@ -225,9 +227,11 @@ func (p *Provider) processOfflineDealProposal(ds *smtypes.ProviderDealState, dh
 		return aerr
 	}
 
-	// Run deal through the filter
-	if aerr := p.runDealFilters(ds); aerr != nil {
-		return aerr
+	// Run deal through the filter if deal filter is set
+	if p.config.StorageFilter != "" {
```
dirkmc, line 231: Let's move this check if the storage filter is set inside runDealFilters

storagemarket/provider_dealfilter.go
```diff
@@ -54,8 +55,49 @@ func (p *Provider) getDealFilterParams(deal *types.ProviderDealState) (*dealfilt
...
-		SealingPipelineState: *sealingStatus,
+		SealingPipelineState: p.spsCache.Status,
 		FundsState:           *fundsStatus,
 		StorageState:         *storageStatus,
 	}, nil
 }
+
+// sealingPipelineStatus updates the SealingPipelineCache to reduce constant sealingpipeline.GetStatus calls
+// to the lotus-miner. This is to speed up the deal filter processing
+func (p *Provider) sealingPipelineStatus() {
```
dirkmc, line 66: Because this call is quite expensive, I'd suggest that we don't do in continuously.

Instead I suggest that when there's a call to get the sealing pipeline status
- check the cache
  - if the value has not expired
    - return the value
    - in the background run a process to get a new value
  - if the value has expired
    - get a new value and wait for it before returning

LexLuthr: I have tweaked the logic a little more. Since we are processing one deal at a time, there is no chance of clash if we don't use goroutines
Filter requests sealing status:

- Check cache
     - if expired or cache error is true 
          - fetch new status (keep the caller waiting)
          - This would also work when Boost starts
          - Return the new status to caller and update cache
     - if cache is not expired
          - Return the cached status to caller (no waiting)
          - Don't trigger a refresh in case we have multiple deals processing at the same time
               - This would prevent the scenario where goroutine and sealingPipelineStatus() both try to update the cache
               - Or we can use locks to prevent this (more complexity)

dirkmc: The problem with the code as implemented is that we're repeatedly calling sealingpipeline.GetStatus, even if there are no deal proposals being made. We have seen that this can be a very expensive call (sometimes takes ~30s). So I would prefer to change the logic to only call sealingpipeline.GetStatus when it's needed. That's what I've proposed above.

LexLuthr: Ahh! Sorry for the confusion in that case.
I have corrected the logic. Pushed new commit with changes.

storagemarket/provider.go
```diff
@@ -147,6 +157,10 @@ func NewProvider(cfg Config, sqldb *sql.DB, dealsDB *db.DealsDB, fundMgr *fundma
 		cfg.MaxConcurrentLocalCommp = 1
 	}
 
+	if cfg.SealingPipelineCacheTimeout < 0 {
+		cfg.SealingPipelineCacheTimeout = 30
```
dirkmc, line 161:
```suggestion
		cfg.SealingPipelineCacheTimeout = 30 * time.Second
```

dirkmc (approved):
Just one comment, please fix and then merge when ready

Issue #1208, opened 2023-02-21
Must allow ConsiderOnlineStorageDeals to accept offline storage deals


hyunmoon:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature or enhancement request. If it is, please open a [new idea discussion](https://github.com/filecoin-project/boost/discussions/new?category=ideas) instead. New feature and enhancement requests would be entertained by the boost team after a thorough discussion only.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to boost.

### Boost component

- [X] boost daemon - storage providers
- [X] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### Boost Version

```text
lotus-miner --version
lotus-miner version 1.19.0+mainnet+git.64059ca87

boostd --version
boostd version 1.5.1+git.5aadde7
```

### Describe the Bug

On boost v1.5.1, ConsiderOnlineStorageDeals must be enable to accept an offline deal.
This problem was fixed after downgrading to v1.5.0.

### Logging Information

```text
-
```

### Repo Steps

Try sending an offline deal to yourself on v1.5.1.

LexLuthr:
This is coming from https://github.com/filecoin-project/boost/blob/e8d1bedc53f67b2b43011a558ea61a8c657022ea/node/modules/dealfilter.go#L52-L55. We send an empty string in transfer type for offline deals. 
Before v1.5.1, we were not running offline deals through the filter and thus this was not caught.

LexLuthr:
Seems to be coming from https://github.com/filecoin-project/boost/blob/e8d1bedc53f67b2b43011a558ea61a8c657022ea/node/modules/dealfilter.go#L41-L44

The logic is 
   transfer.type is not "manual" = true
   !online deal = true
   This trigger an early return.
   
This would also cause to never reject the offline deals even when turned off.
[...]

PR #1210, opened 2023-02-22
fix: BasicDealFilter logic

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1208
This PR also extend the deal params exposed to the external deal filters like CIDGravity. Before this an offline deal was sending a false value for IsOffline to external filter.

### DIFF
--- a/storagemarket/provider_dealfilter.go
+++ b/storagemarket/provider_dealfilter.go
@@ -46,9 +46,6 @@ func (p *Provider) getDealFilterParams(deal *types.ProviderDealState) (*dealfilt
 		ClientDealProposal: deal.ClientDealProposal,
 		DealDataRoot:       deal.DealDataRoot,
 		Transfer:           deal.Transfer,
-		IsOffline:          deal.IsOffline,
-		SkipIPNIAnnounce:   !deal.AnnounceToIPNI,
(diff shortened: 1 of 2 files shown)


storagemarket/provider_dealfilter.go
```diff
@@ -46,6 +46,9 @@ func (p *Provider) getDealFilterParams(deal *types.ProviderDealState) (*dealfilt
 		ClientDealProposal: deal.ClientDealProposal,
 		DealDataRoot:       deal.DealDataRoot,
 		Transfer:           deal.Transfer,
+		IsOffline:          deal.IsOffline,
+		SkipIPNIAnnounce:   !deal.AnnounceToIPNI,
+		RemoveUnsealedCopy: !deal.FastRetrieval,
```
dirkmc, line 51: Let's do this in a separate PR as it requires a version change on the deal filter JSON object

dirkmc (approved):

storagemarket/provider_dealfilter.go
LexLuthr: Sure. I will remove it with a commit. Running a devnet at the moment to check the original problem. Will push the commit after.

Issue #1215, opened 2023-02-23
DealProposal FRC integration


nonsense:
This is a follow up issue to:
- https://github.com/filecoin-project/boost/issues/1161
- https://github.com/filecoin-project/boost/pull/1183

---

TODO:
- [x] fix/review client mismatch
- [ ] add an itest which exercises contract deals and eth_ subscriptions and eth_ state calls
- [x] finalize type for `ContractDealProposal` and add `RemoveUnsealedCopy` as part of initial params v1.
- [x] boost UI - visualize all contract deal proposals to increase visibility for all deals that are made via on-chain proposals
- [x] consider if `from` should stay as a config value, or we should infer it somehow from the miner actor ID (as `lotus evm stat` does)
- [ ] ...

PR #1216, opened 2023-02-23
FRC DealProposal integration

nonsense:
An initial DealProposal FRC has been merged into Boost already. This PR is a follow-up to solidify the remaining items as described in https://github.com/filecoin-project/boost/issues/1215 as well as any changes that might come up in the actual FRC.

---

TODO:
- [x] test with local devnet - tested with lotus `v1.20.0-rc2-boost`
- [ ] ideally add an itest which exercises the current agreed-upon contract/ABI and we detect properly the event on boost side


storagemarket/types/contract_deal_proposal_types.go
```diff
@@ -7,11 +7,14 @@ import (
...
 )
 
+//go:generate cbor-gen-for --map-encoding ContractDealProposal ContractParamsVersion1
+
 type ContractDealProposal struct {
 	PieceCID     cid.Cid
 	PieceSize    abi.PaddedPieceSize
 	VerifiedDeal bool
 	Client       address.Address
+	Provider     string
```
nonsense, line 17: `Provider` should probably go into the `ContractParamsVersion1`, given that it is optional and given that we might want to change to a slice in the future, etc.

storagemarket/helper.go
```diff
@@ -191,5 +191,9 @@ func (c *ChainDealManager) CheckDealEquality(ctx context.Context, tok ctypes.Tip
 		p1.ProviderCollateral.Equals(p2.ProviderCollateral) &&
 		p1.ClientCollateral.Equals(p2.ClientCollateral) &&
 		p1.Provider == p2.Provider &&
-		p1ClientID == p2ClientID, nil
+		p1ClientID == p2ClientID
+
+	log.Warnw("check deal quality", "result", res, "p1clientid", p1ClientID, "p2clientid", p2ClientID, "label_equality", p1.Label.Equals(p2.Label), "provider_equality", p1.Provider == p2.Provider)
```
dirkmc, line 196: Maybe log this at the debug level

dirkmc (approved):

storagemarket/helper.go
nonsense, line 196:
```suggestion
	log.Debugw("check deal quality", "result", res, "p1clientid", p1ClientID, "p2clientid", p2ClientID, "label_equality", p1.Label.Equals(p2.Label), "provider_equality", p1.Provider == p2.Provider)
```

PR #1217, opened 2023-02-23
fix: bug bypassing basic filters

LexLuthr:
Basic filters are now working as expected.

Online = false and offline = true

Screenshot 2023-02-24 at 1 31 54 AM

online = true and offline = false
Screenshot 2023-02-24 at 1 33 42 AM

### DIFF
--- a/storagemarket/provider_loop.go
+++ b/storagemarket/provider_loop.go
@@ -69,6 +69,9 @@ type acceptError struct {
 	reason string
 }
 
+// we still need to call the BasicDealFilter() even when external deal filter is not set.
+// Once BasicDealFilter() completes the checks like are we accepting online deal, verified deal etc.
+// Then it runs the external "cmd" filter. Thus, runDealFilters is not optional of any type of deal
 func (p *Provider) runDealFilters(deal *types.ProviderDealState) *acceptError {
 
 	// run custom storage deal filter decision logic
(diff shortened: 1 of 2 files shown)


storagemarket/provider_loop.go
```diff
@@ -70,9 +70,6 @@ type acceptError struct {
 }
 
 func (p *Provider) runDealFilters(deal *types.ProviderDealState) *acceptError {
-	if p.config.StorageFilter == "" {
-		return nil
-	}
```
dirkmc, line 75: Can you explain the change here?
I don't understand why we would make the call to get deal filter params if the StorageFilter is not set?

LexLuthr: So, it would seem we still need to call the [BasicDealFilter()](https://github.com/filecoin-project/boost/blob/d70db4757cdc3906e21a36a27467b30b9d0f819e/node/modules/dealfilter.go#L13) even when external deal filter is not set. The removed part of the code skipped that section and basic deal filter was not running. Once BasicDealFilter() completes the checks like we have enough storage, funds etc on SP side, then [it runs "cmd" filter](https://github.com/filecoin-project/boost/blob/d70db4757cdc3906e21a36a27467b30b9d0f819e/node/modules/dealfilter.go#L92-L94).

dirkmc: Ok that makes sense, thanks 👍 

In that case I would change the code in `getDealFilterParams` so that if the storage filter is not set, it doesn't get the funds or storage status either.

In other words at the top of `getDealFilterParams`:
```
if p.config.StorageFilter == "" {
	return &dealfilter.DealFilterParams{
                FundsState: <zero value for funds status>,
                StorageState: <zero value for storage state> ,
        }
}
```

Please also add a comment to the code explaining what you said in your github comment above.

LexLuthr:
Updated the logic as per suggestion. Tested on devnet, working as expected.

Screenshot 2023-02-24 at 4 13 36 PM

dirkmc (approved):

dirkmc:
Nice work 🙌

PR #1221, opened 2023-02-25
hyperspace build target for boost

nonsense:


LexLuthr:
@nonsense Do we want to merge ntwk/hyperspace in main? I was thinking about keeping because it might become tricky to consolidate dependencies as both branches of Lotus will diverge between v1.20.0-X and lotus - ntwk/hyperspace. They are keeping both branch separate and I am not aware of any plan to merge as of now.

PR #1223, opened 2023-02-27
add start-epoch-head-offset

anjor:


anjor:
Fixes https://github.com/filecoin-project/boost/issues/1224

nonsense:
Thank you @anjor !

anjor:
yay, thanks for the quick merge, and pushing the commit to rename -- i agree with that change!

PR #1229, opened 2023-02-28
chore: enable feature and enhancement report

LexLuthr:
When we were launching Boost, it was decided that we will entertain feature/enhancement requests only via discussion. After multiple iterations, we have seen that requests coming in have been reasonable and useful.

If everyone in the team agrees, I would like to enable direct filing of feature requests and enhancement requests at this point. I will also update the docs to point to these templates.


dirkmc (approved):

dirkmc:
LGTM 👍 
I think it makes sense to enable feature / enhancement requests

PR #1231, opened 2023-02-28
fix homedir expansion

nonsense:


cmd/boost/retrieve_cmd.go
```diff
@@ -85,6 +86,12 @@ var retrieveCmd = &cli.Command{
 		}
 
 		cfgdir := cctx.String(cmd.FlagRepo.Name)
+
+		cfgdir, err = homedir.Expand(cfgdir)
+		if err != nil {
+			return nil, fmt.Errorf("expanding homedir: %w", err)
+		}
+
 		node, err := clinode.Setup(cfgdir)
```
nonsense, line 95: `cfgdir` is expanded within clinode.Setup, but it is also currently used in non-expanded form further down in this function at:

`fc, err := rc.NewClient(node.Host, api, node.Wallet, addr, bstore, ds, cfgdir)`

---

So this PR is expanding the cfgdir immediately when we get it from the flag.

LexLuthr (approved):

dirkmc (approved):

Issue #1125, opened 2023-01-26
Retrievals hang at stage 'PaymentChannelSkip (Ongoing)'


jacobheun:
@jcace reported this issue in #boost-help in filecoin slack.

## Summary
Retrieval attempts to Boost leveraging `lotus client retrieve` will sometimes hang indefinitely or until the client times out. This may require a restart of the process to get it unstuck, but the issue can recur after restarts. 

Initial reported versions: 
```
lotus version 1.19.0+mainnet+git.64059ca87.dirty
boostd version 1.5.0+git.dbedffb
```

### Note from the user
>I have a very strange retrieval issue.  Consider these two different retrieval commands:
a) lotus client retrieve --provider f01963614 QmfB9Sz8UYnTd8o8BVHrmGcrRcqDgu8jM8vFXUnFsTGbGE --car
b) lotus client retrieve --provider f01886797 QmaQpw28KgCjS8PRCDwSmn5jaoScYN4RYpqWxERtmqEsnE --car
a) works perfect, transfer starts right away, but b) just hangs at PaymentChannelSkip (Ongoing) and never progresses. 
It will eventually time out.

>What’s strange about this is I control both SP nodes. In fact, they are both physically 
running on the same hardware. They are on the same internet connection. 
They are storing files on the same physical disks.

>Both CIDs are in a sector with an unsealed copy, and I’ve verified that it exists. 
The only difference is that node b) currently has all my sealing workers connected 
to it, whereas node a) does not - thus, unsealing is not possible. However, looking into the code 
it would seem that the [IsUnsealed](https://github.com/filecoin-project/lotus/blob/e68b8e96c7c28f4ff65f4b64254717549d158e87/storage/sealer/piece_provider.go#L52) 
function should be able to find the sector just fine. The lotus-miner instance has the storage directly attached to it in both instances.

> Boost/Miner configs between the two nodes are almost identical - did not modify anything to do with retrievals. 
What could possibly cause this?

## References: 
* https://filecoinproject.slack.com/archives/C03CKDLEWG1/p1673319661213899
* https://filecoinproject.slack.com/archives/C03CKDLEWG1/p1673040967787369

Related https://github.com/filecoin-project/boost/issues/846

jcace:
Thanks for creating this issue @jacobheun !!

davidd8:
[...]

PR #1233, opened 2023-02-28
feat: add job to cancel stalled retrievals

jacobheun:
## Summary
As part of debugging https://github.com/filecoin-project/boost/issues/1125 we determined that stalled retrievals were never being cleaned up by the go-fil-markets code. This was originally intended to allow resumption of retrieval, however, there was never a cap put on the duration and retrievals would require manual termination. This behavior doesn't scale.

There is refactoring that is needed for the go-fil-markets retrieval codepaths as a separate effort, but in the interim, this will provide a short term fix for canceling stalled retrieval.

This new job runs every 5 minutes and will cancel all retrievals in the Retrieval Log that have not been updated in the past 30 minutes (configurable). 30 minutes should be a reasonable default for clients to attempt to reconnect and resume transfer if desired.

## Caveats of this approach
I have noted in the config type docs that the RetrievalLog duration must exceed the stalled retrieval duration in order for this to function properly. The reason the RetrievalLog is being used here for pruning is that it is significantly more efficient than querying the go-fil-markets statemachines, as the latter loads everything into memory. Once we have refactored the legacy retrieval code, we can migrate to using a more thorough check for stalled retrieval than the logs db.

## Devnet Testing
I set up my local devnet to prune stalled retrievals older than 5 minutes and executed 100 retrievals and abruptly terminated their connection from the client. After the 5 minute stall window was passed, all deals were executed.

Screenshot 2023-02-28 at 2 12 54 PM

Screenshot 2023-02-28 at 2 20 10 PM

### DIFF
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -176,7 +176,11 @@ type DealmakingConfig struct {
 	// The amount of time to keep deal proposal logs for before cleaning them up.
 	DealProposalLogDuration Duration
 	// The amount of time to keep retrieval deal logs for before cleaning them up.
+	// Note RetrievalLogDuration should exceed the StalledRetrievalDuration as the
+	// logs db is leveraged for pruning stalled retrievals.
 	RetrievalLogDuration Duration
+	// The amount of time stalled retrieval deals will remain open before being canceled.
+	StalledRetrievalDuration Duration
 
 	// A command used for fine-grained evaluation of storage deals
 	// see https://boost.filecoin.io/configuration/deal-filters for more details
--- a/node/config/types.go
+++ b/node/config/types.go
@@ -176,11 +176,11 @@ type DealmakingConfig struct {
 	// The amount of time to keep deal proposal logs for before cleaning them up.
 	DealProposalLogDuration Duration
 	// The amount of time to keep retrieval deal logs for before cleaning them up.
-	// Note RetrievalLogDuration should exceed the StalledRetrievalDuration as the
+	// Note RetrievalLogDuration should exceed the StalledRetrievalTimeout as the
 	// logs db is leveraged for pruning stalled retrievals.
 	RetrievalLogDuration Duration
 	// The amount of time stalled retrieval deals will remain open before being canceled.
-	StalledRetrievalDuration Duration
+	StalledRetrievalTimeout Duration
 
 	// A command used for fine-grained evaluation of storage deals
 	// see https://boost.filecoin.io/configuration/deal-filters for more details
--- a/retrievalmarket/rtvllog/retrieval_log.go
+++ b/retrievalmarket/rtvllog/retrieval_log.go
@@ -14,23 +14,27 @@ import (
 var log = logging.Logger("rtrvlog")
 
 type RetrievalLog struct {
-	db              *RetrievalLogDB
-	duration        time.Duration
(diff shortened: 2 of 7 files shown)


node/config/types.go
```diff
@@ -176,7 +176,11 @@ type DealmakingConfig struct {
 	// The amount of time to keep deal proposal logs for before cleaning them up.
 	DealProposalLogDuration Duration
 	// The amount of time to keep retrieval deal logs for before cleaning them up.
+	// Note RetrievalLogDuration should exceed the StalledRetrievalDuration as the
+	// logs db is leveraged for pruning stalled retrievals.
 	RetrievalLogDuration Duration
+	// The amount of time stalled retrieval deals will remain open before being canceled.
+	StalledRetrievalDuration Duration
```
dirkmc, line 183: I suggest we call this StalledRetrievalTimeout

retrievalmarket/rtvllog/retrieval_log.go
```diff
@@ -7,32 +7,38 @@ import (
...
+	stalledDuration time.Duration
+	ctx             context.Context
 
 	lastUpdateLk sync.Mutex
 	lastUpdate   map[string]time.Time
 }
 
-func NewRetrievalLog(db *RetrievalLogDB, duration time.Duration) *RetrievalLog {
+func NewRetrievalLog(db *RetrievalLogDB, duration time.Duration, dt lotus_dtypes.ProviderDataTransfer, stalledDuration time.Duration) *RetrievalLog {
 	return &RetrievalLog{
```
dirkmc, line 28: Suggest logging a warning if stalledDuration < duration

dirkmc (approved):
Love the comprehensive description 🤩 

I left a couple of small suggestions, feel free to merge when ready.

benjaminh83:
Just checked out latest master. Its reporting: `boostd version 1.6.0-rc1+git.b0de635`

I wonder if this version of boost isn't a bit too trigger happy on the cancels? Seems like a lot of deals are uploading, men then just getting canceled. 

image

jacobheun:
@benjaminh83 thanks for sharing this, did you update the config at all for the new `StalledRetrievalTimeout`? 

By default this checks deals that have not been updated in 30minutes. I don't think any of these cancels are actually coming from this update as they're happening too soon and we're still seeing valid transfers, but it would be helpful for us to try and set the message on the transfer that the job timed it out so it's easier to see when those are actually happening. (If you haven't changed your config, I suspect the `ResponderPaused` retrievals will time out in 30mins if something else doesn't terminate/complete them before that).

As part of diagnosing this we also discovered that some [Lassie](https://github.com/filecoin-project/lassie) which is being used be some of the larger retrieval systems (such as Autoretrieve) weren't gracefully canceling retrievals they no longer needed, such as during parallel fetches. That was fixed and deployed yesterday, which explains some of what we're seeing here with the "client cancelled retrieval", and requests for the same CID showing cancels followed by a successful retrieval.

I'm not clear why we're not seeing "client cancelled retrieval" on more of these, I suspect it's a timing issue but will dig into that.

jacobheun:
FYI I created https://github.com/filecoin-project/boost/issues/1238 to make it easier to quickly see if the PeerIDs are different. I'd like to know how many of these are coming from different peers.

PR #1236, opened 2023-02-28
Simple graphsync retrieval server

dirkmc:
Implement a retrieval server for graphsync retrievals that intercepts incoming requests and
- handles them if they're for unpaid retrievals
- passes them on to the existing go-fil-markets implementation if they're for paid retrievals

## Todo
- [x] Boost should not attempt to dial back a client when it disconnects - currently it does a backoff retry
- [x] Update data-transfer cancel methods (including retrieval log stalled timeout job) to attempt the cancel on the unpaid retrieval path, and then the legacy path
- [x] Show the list of active retrievals in `boostd data-transfers list`
- [x] Ensure that `boostd data-transfers cancel` cancels retrievals in the new code path
- [x] Update graphsync version to an official tag

### DIFF
--- a/retrievalmarket/server/gsunpaidretrieval.go
+++ b/retrievalmarket/server/gsunpaidretrieval.go
@@ -27,6 +27,11 @@ import (
 
 var log = logging.Logger("boostgs")
 
+var incomingReqExtensions = []graphsync.ExtensionName{
+	extension.ExtensionIncomingRequest1_1,
+	extension.ExtensionDataTransfer1_1,
+}
+
 // Uniquely identify a request (requesting peer + data transfer id)
 type reqId struct {
 	p  peer.ID
@@ -260,9 +265,22 @@ func (g *GraphsyncUnpaidRetrieval) RegisterIncomingRequestHook(hook graphsync.On
 				return fmt.Errorf("creating accept response message: %w", msgErr)
 			}
 
-			// Send an accept message / validation failed message
-			if err := g.dtnet.SendMessage(g.ctx, p, respMsg); err != nil {
-				return fmt.Errorf("failed to send accept message to requestor %s: %w", p, err)
+			// Send an accept message / validation failed message as an extension
+			// (a go-data-transfer protocol message that gets embedded in a
+			// graphsync message)
+			if respMsg != nil {
+				// This hook uses a unique extension name so it can be attached
+				// to a graphsync message with data from a different hook.
+				// incomingReqExtensions also includes the default extension
+				// name to maintain compatibility with previous data-transfer
+				// protocol versions.
+				extensions, extensionErr := extension.ToExtensionData(respMsg, incomingReqExtensions)
+				if extensionErr != nil {
+					return fmt.Errorf("building extension data: %w", extensionErr)
+				}
+				for _, ext := range extensions {
+					hookActions.SendExtensionData(ext)
+				}
 			}
 
 			if validateErr == datatransfer.ErrPause || validateErr == datatransfer.ErrResume {
@@ -273,6 +291,7 @@ func (g *GraphsyncUnpaidRetrieval) RegisterIncomingRequestHook(hook graphsync.On
 		}()
 
 		if err != nil {
+			hookActions.TerminateWithError(err)
 			g.failTransfer(state, err)
 			return
 		}
--- a/retrievalmarket/server/gsunpaidretrieval.go
+++ b/retrievalmarket/server/gsunpaidretrieval.go
@@ -283,6 +283,10 @@ func (g *GraphsyncUnpaidRetrieval) RegisterIncomingRequestHook(hook graphsync.On
 			return
 		}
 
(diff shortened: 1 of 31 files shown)


retrievalmarket/server/gsunpaidretrieval.go
```diff
@@ -0,0 +1,475 @@
...
+		return nil, xerrors.Errorf("unknown voucher type: %s", vtypStr)
+	}
+	encodable, err := request.Voucher(decoder)
+	if err != nil {
+		return nil, err
+	}
+	return encodable.(datatransfer.Registerable), nil
+}
+
+func (g *GraphsyncUnpaidRetrieval) isRequestForActiveUnpaidRetrieval(p peer.ID, request graphsync.RequestData) (datatransfer.Request, *retrievalState, bool) {
```
hannahhoward, line 447: as I read this, this only returns true if a retrieval was previously tracked with trackTransfer? so this is called for all the hooks after IncomingRequestQueued?

retrievalmarket/server/gsunpaidretrieval.go
```diff
@@ -0,0 +1,475 @@
...
+		// if they are in the list (meaning, they aren't actually sent over the
+		// wire). So here we check if the block was actually sent
+		// over the wire before firing the data sent event.
+		if block.BlockSizeOnWire() == 0 {
+			return
+		}
+
+		// Fire block sent event
+		state.cs.sent += block.BlockSizeOnWire()
+		g.publishDTEvent(datatransfer.DataSent, "", state.cs)
```
hannahhoward, line 402: DataSentProgress as well? I have no idea why there are two events -- someone else put that in, but probably best for compatibility.

retrievalmarket/server/gsunpaidretrieval.go
```diff
@@ -0,0 +1,475 @@
...
+		_, state, intercept := g.isRequestForActiveUnpaidRetrieval(p, request)
+		if !intercept {
+			listener(p, request, err)
+			return
+		}
+
+		// Don't set the channel state to errored, as a network error is not
+		// necessarily fatal
+		// TODO: If the client always sends a restart message after a network
+		// disconnect, then maybe a network error *should* be fatal
```
hannahhoward, line 418: I believe this is cause channel state errored is a terminal state you can't get out of. A network error should ideally put you in a "Waiting for restart" type state. Because a client doesn't ALWAYS send a restart. The network error could be cause they closed their laptop and won't be trying again for a day. This is partly why Jacob had to write the cleanup job. And keeping the request around for anything more than maybe a minute is a DOS vector, especially since we're tracking all state in memory now.

hannahhoward (commented):
I have some anxiety about all this working together, but I'd say why not try right?

The one other thing you might want to look into is hacking the datatransfer.Network interface so you can intercept incoming messages first before go-data-transfer.

Also the DOS vector with network errors documented below needs some care.

Finally, a heads up/caution that go-data-tranasfer v2 is going into lotus and brings graphsync updates and data transfer interface changes that will require some code changes here when it lands in Boost. @rvagg and I would be happy to assist with that. (can make the PR myself if that helps, cause I don't want you all to be burdened)

retrievalmarket/server/gsunpaidretrieval.go
dirkmc: The only place these events are being listened for is in boost (so it can display them to the user) so I don't think we need to replicate every event.

retrievalmarket/server/gsunpaidretrieval.go
dirkmc: Thanks, I guess we can just error out then. If the client sends a restart request it will just open a new transfer (for the remaining data)

retrievalmarket/server/gsunpaidretrieval.go
hannahhoward: ah, totally makes sense.

retrievalmarket/server/gsunpaidretrieval.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+	err = typeRegistry.Register(&migrations.DealProposal0{}, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return &GraphsyncUnpaidRetrieval{
+		GraphExchange:    gs,
+		peerID:           peerID,
+		dtnet:            dtnet,
```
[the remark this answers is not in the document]
dirkmc: We tested it out - the client cancels the graphsync retrieval and also sends a cancel message over the go-data-transfer protocol.
On the provider side the OnRequestorCancelledListener gets called. There is a test that verifies this behaviour in gsunpaidretrieval_test.go

jacobheun:
We'll post final performance metrics soon but after testing in local devnet and having this branch running on Sofiaminer overnight we're seeing significant improvements in reliability and performance. We're going to go ahead and merge this, then update the LID branch for final testing on our Sofia and filcollins SP's.

PR #1248, opened 2023-03-02
Remove specs-storage dependency, use Lotus storiface instead

RobQuistNL:
It looks like this specs-storage dependency is a stripped down (possibly no longer synchronized?) version of what already is in Lotus. I'm not sure why this is the case, but to me it seems like this PR is a better solution.

Best to keep it up to date by using the dependency from the Lotus project itself, IMHO.


storage/sectorblocks/sectorblocks.go
```diff
@@ -48,7 +48,7 @@ func DsKeyToDealID(key datastore.Key) (uint64, error) {
 }
 
 type SectorBuilder interface {
```
nonsense, line 50: Actually I think we are not importing this interface, but the one from Lotus.

nonsense:
Good catch -- we should fix this.

RobQuistNL:
These are the only places where the specs-storage are being used

storage/sectorblocks/sectorblocks.go
RobQuistNL: You're right, I've removed the entire file as its not being used.

jacobheun (approved):
Thanks for the PR! I pushed up a commit of the merge to verify CI was passing, this is gtg.

Issue #1238, opened 2023-03-01
Update PeerID displays in the UI so that more than just the common prefix is displayed


jacobheun:
## Summary
This example image shows the retrieval logs page shared by an SP in a separate issue. The PeerID's being displayed are all showing the same common prefix so they look like they could be the same peer, but could very well all be different. We should either remove the common prefix from PeerID's when displayed in the UI, or just show the tail of the peer id.

![image](https://user-images.githubusercontent.com/639834/222128282-2cf7a5e5-491d-4fe5-a0ee-a338091b46fa.png)

PR #1252, opened 2023-03-03
feat(ui): display end of peerid instead of front

jacobheun:
## Summary
Fixes https://github.com/filecoin-project/boost/issues/1238
Instead of displaying the common characters in a Peer ID, this will display the end of the PeerId. This makes it easier to quickly discern requests from different peers.

You can see in the screenshots below of the exact same data that while the PeerIDs all look the same before, showing the last 8 characters instead of the first 8 clearly shows they're all unique peers.

**Before**
Screenshot 2023-03-03 at 2 08 34 PM

**After**
Screenshot 2023-03-03 at 2 16 41 PM

### DIFF
--- a/react/src/Components.js
+++ b/react/src/Components.js
@@ -28,7 +28,7 @@ export function ShortClientAddress(props) {
 }
 
 export function ShortPeerID(props) {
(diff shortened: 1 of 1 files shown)


dirkmc (approved):
Thanks!

PR #1254, opened 2023-03-03
remove defer statement for boost retrieve

nonsense:


retrievalmarket/client/client.go
```diff
@@ -549,14 +549,16 @@ func (c *Client) retrieveContentFromPeerWithProgressCallback(
 	chanid = newchid
 	chanidLk.Unlock()
 
-	defer func() { _ = c.dataTransfer.CloseDataTransferChannel(ctx, chanid) }()
```
nonsense, line 552: We certainly should not have the `defer func() { _ = c.dataTransfer.CloseDataTransferChannel(ctx, chanid) }()` here as it is called even on successful/completed transfers and we don't want that.

dirkmc (approved):

PR #1265, opened 2023-03-07
feat: Online backup

### LINKED ISSUE #1235: Online backup for Boost
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the Boost forum](https://github.com/filecoin-project/boost/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Boost component

- [X] boost daemon - storage providers
- [ ] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

Boost requires an online backup command as offline backup requires shutting down Boost

### Describe the solution you'd like

Boost should have an API which can be called with a destination path string to create backup

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

LexLuthr:
Fixes #1235 

- [x] Add a simple test or figure out another method to test (would like to avoid itest)


node/impl/backupmgr/backup.go
```diff
@@ -0,0 +1,314 @@
...
+		return fmt.Errorf("error creating config directory %s: %w", path.Join(bkpDir, ConfigDirName), err)
+	}
+	if err := CopyFiles(b.src.Path(), dst.Path(), fl); err != nil {
+		return fmt.Errorf("error copying file: %w", err)
+	}
+
+	// Close backup repo before moving it
+	if err := dst.Close(); err != nil {
+		return fmt.Errorf("error closing the locked backup repo: %w", err)
+	}
```
dirkmc, line 129: Probably dst.Close should be in a defer statement right after dst.Lock, to make sure it gets called even if there's an error

node/impl/backupmgr/backup.go
```diff
@@ -0,0 +1,314 @@
...
+
+	// Move directory to the backup location
+	if err := os.Rename(bkpDir, path.Join(dstDir, "boost_backup_"+time.Now().Format("20060102150405"))); err != nil {
+		return fmt.Errorf("error moving backup directory %s to %s: %w", bkpDir, dstDir, err)
+	}
+
+	// TODO: Archive the directory and save it to the backup location
```
dirkmc, line 136: We can remove this TODO right?

node/impl/backupmgr/backup_test.go
```diff
@@ -0,0 +1,9 @@
+package backupmgr
+
+import (
+	"testing"
+)
+
+func TestBackup(t *testing.T) {
```
dirkmc, line 7: Let's remove this test

node/repo/boostrepo.go
```diff
@@ -0,0 +1,31 @@
+package repo
```
dirkmc, line 1: It's not a problem but I'm curious why move the Boost repo implementation from where it was?

db/db.go
```diff
@@ -9,25 +9,42 @@ import (
...
-	db, err := sql.Open("sqlite3", "file:"+dbPath)
+func SqlDB(dbPath, driverName string) (*sql.DB, *sqlite3.SQLiteConn, error) {
+	var sqlite3conn *sqlite3.SQLiteConn
+	sql.Register(driverName, &sqlite3.SQLiteDriver{
+		ConnectHook: func(conn *sqlite3.SQLiteConn) error {
+			sqlite3conn = conn
+			return nil
+		},
+	})
+	db, err := sql.Open(driverName, "file:"+dbPath)
```
dirkmc, line 34: I think it may be easier to add a separate method to get the `*sqlite3.SQLiteConn` connection on demand, rather than adding it as a third return value on the existing SqlDB method

node/repo/boostrepo.go
LexLuthr: Initially I was keeping backupmgr within modules, that was creating a cyclic import. To avoid this I moved it to it's own package. Although, I moved backup out of module to its own package, but this cyclic dependency issue might come again for someone working on module section. Keeping that in mind, I did not move it back.

cmd/boostd/backup.go
```diff
@@ -240,7 +239,7 @@ var restoreCmd = &cli.Command{
 			return err
 		}
 
-		fpathName := path.Join(bpath, metadaFileName)
+		fpathName := path.Join(bpath, backupmgr.MetadataFileName)
```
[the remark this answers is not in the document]
LexLuthr: This is the source file i.e. inside the backup directory. We will use this file to restore at `<boost repo>/datastore/metadata`

node/impl/backupmgr/backup.go
LexLuthr: Yeah, I don't plan to create zip files anymore. At least till folks complaint about it.

db/db.go
LexLuthr: I couldn't figure out a safe way to do this. We could try https://pkg.go.dev/database/sql#Conn.Raw but It doesn't allow driverConn must not be used outside of f. On the other hand, backup() method works on destination not source and requires a source sqliteConn{} to work.

dirkmc: With this change all calls to `SqlDB()` need to pass a driverName, that is only used to create a `SQLiteConn` that most of them don't use. In each case the new driverName is added to go's sql driver registry, which is a bit strange.

I'd suggest that we
- leave the `SqlDB()` function as it is
- move sqlBackup to the db package
- create a new function in the db package to get the SQLiteConn for source and destination db
- call the new function from sqlBackup

LexLuthr: Redesign works as expected. I was able to take a backup and restore.

dirkmc (approved):

LexLuthr:
@dirkmc Can you review again after the new commits? I will be merging after this review.
@TippyFlitsUK Tested the new code and everything is working as expected.

node/modules/storageminer.go
```diff
@@ -683,3 +683,9 @@ func NewOnlineBackupMgr(cfg *config.Boost) func(lc fx.Lifecycle, r lotus_repo.Lo
 		return backupmgr.NewBackupMgr(r, ds, db.DealsDBName, dealsDB)
 	}
 }
+
```
dirkmc, line 686: Do we need this change for backup?

dirkmc (approved):

node/modules/storageminer.go
LexLuthr: It came because of merge conflict in file as both func() were in same line. Just resolved the conflict and 1 empty line between 2 funcs.

dirkmc: Ok thanks, feel free to merge

PR #1268, opened 2023-03-09
Add monitoring for graphsync retrievals

dirkmc:
Screenshot 2023-03-09 at 11 50 31 AM


jacobheun:
I think I want to revisit these graphs after we get some feedback. As I've been doing performance testing I think these would be more useful as cumulative totals for the time period. Errors could potentially stay as is, but the rest I think will be more informative to change as cumulative totals would let us both see rates of change along with totals. Right now you can't easily see how many total bytes, total transfers, etc have happened over the shown time period.

Issue #1273, opened 2023-03-14
Boost deals stuck at "failed to add index and announce deal"


RobQuistNL:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/boost/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/boost/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [ ] I did not make any code changes to boost.

### Boost component

- [ ] boost daemon - storage providers
- [ ] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### Boost Version

```text
1.6.0
```

### Describe the Bug

I occasionally see a bunch of deals in this state;
```
Paused at 'Announcing': failed to add index and announce deal: failed to register deal with dagstore: failed to register shard: failed to read/generate CAR Index: error reading car header: getting backing reader: getting reader at +0: non-200 code: 500
```

when I click the "retry" button it almost instantly goes to "proving"
Can this be automated?

Might be related: https://github.com/filecoin-project/boost/issues/664

### Logging Information

```text
155297-{"level":"debug","ts":"2023-03-01T01:45:24.278Z","logger":"dagstore","caller":"dagstore@v0.5.5/dagstore_control.go:64","msg":"processing task","op":"OpShardFail","shard":"baga6ea4seaqm4xiwb6owue6sfh3buuf2ix5nnmx5vus6hsirldf6euiovjawqja","error":"failed to read/generate CAR Index: error reading car header: getting backing reader: getting reader at +0: non-200 code: 500"}
[...]

PR #1274, opened 2023-03-14
feat: retry register shard

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1273


LexLuthr:
closing this as stale. We will need to check if the issue persists in v2 and rethink about our approach.

PR #1281, opened 2023-03-15
timeout on EthSubscribe

nonsense:
EthSubscribe typically should not block, but at the moment it does.

This PR adds a timeout to it, so that we don't block the rest of Boost if there is an issue with the subscription.


nonsense:
Closing in favour of https://github.com/filecoin-project/boost/pull/1282

Issue #1286, opened 2023-03-16
Convert PSD simulation fail to a fatal error


nonsense:
Until now we've considered PSD fail to be a recoverable error for good reasons.

With the introduction of contract deals, I suggest we convert PSD to a retryable error (or just flat out to fatal error, if it is hard to do a 3-times-retry), as a stop-gap to clear up the staging area/pipeline on SP side.

We could do that for all clients, or only for `f4` clients.

marshyonline:
I think a simple check that if it is a f4 client and it received the PSD fail or PDS Simulation fail it can just be wiped and tagged funds and storage released. 
Standard deals should never get into this state and usually a retry is enough to get them going again - but in the contract case is 99% of the time its fatal.

I woke to over 100 PSD simulation faults and boost was extremely slow, almost unresponsive. All 50FiL in my PSD wallet tagged to Failed Deals. it took 45mins of clicking to clear out the failed deals.

dirkmc:
I think that makes sense, thanks @marshyonline 

By the way you can run a script to clear failed deals:
https://boost.filecoin.io/graphql-api#example-queries

dirkmc:
@nonsense this is the place in the code where we would need to add the check:
https://github.com/filecoin-project/boost/blob/28881a99f6aca91905fd510a480c40577aefc694/storagemarket/deal_execution.go#L434-L458

Do you know if there's a way that we can tell what kind of error it is, so that if it's caused by the deal being published by another SP we can automatically fail the deal?

PR #1294, opened 2023-03-17
fatally fail contract deals on publish errors

nonsense:
Fixes: https://github.com/filecoin-project/boost/issues/1286


dirkmc (commented):
Code looks good - could you please add a simple test to provider_test.go

Issue #1271, opened 2023-03-13
Update boostx stats command to include agent version & retrieval support breakdown


jacobheun:
## Summary
We want to better monitor adoption of Boost as well as the various retrieval protocols (graphsync, bitswap, http) so we can better support SPs.

### Collect libp2p agent version
The libp2p agent versions should give us a better understanding of who is using lotus/venus/boost/etc and what version adoption of Boost looks like. We already run identify to get the protocols, so we just need to track the agent versions.

This could just be a simple map counter we print out.

### Collect supported retrieval protocols
https://github.com/filecoin-project/boost/pull/723 added support for checking supported retrieval protocols over libp2p. We should run this check to see which protocols each SP supports as we crawl the network.

**Example existing output**
```
...
Total Boost nodes: 309
Total Boost raw power: 673768287275843584
Total Boost quality adj power: 2497977391189557248
Total Lotus Markets nodes: 591
Total SPs with minimum power:  3588
Total Indexer nodes: 721
```

**Example updated output**
```
...
Total Boost nodes: 309
Total Boost raw power: 673768287275843584
Total Boost quality adj power: 2497977391189557248
Total Lotus Markets nodes: 591
Total SPs with minimum power:  3588
Total Indexer nodes: 721
Retrieval Protocol Support:
  graphsync: 309
  http: 25
  bitswap: 19
Agent Versions:
  boost/1.6.0: 250
  boost/1.6.1-rc1: 10
  lotus/1.20.3: 489
  otherversionhere: x
  ...
```

PR #1304, opened 2023-03-20
boostx stats: output agent version and retrieval protocols

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/1271

```
Total Boost nodes: 17
Total Boost raw power: 20588664467750912
Total Boost quality adj power: 164240643386998784
Total Lotus Markets nodes: 40
Total SPs with minimum power:  3603
Total Indexer nodes: 43
Agent Versions:
  lotus-1.18.0+mainnet: 1
  lotus-1.18.2+mainnet: 1
  lotus-1.20.0+mainnet: 2
  lotus-1.20.0-rc1+mainnet: 2
  lotus-1.20.0-rc2+debug: 1
  lotus-1.20.0-rc2+mainnet: 8
  lotus-1.20.6+mainnet: 2
Retrieval Protocol Support:
  bitswap: 1
  libp2p: 17
```

### UserAgent
 
Unfortunately it seems that the UserAgent is being set to the lotus version imported by boost (not boost's build version).

This is because we build the Host using lotus modules rather than building it from boost modules:
https://github.com/filecoin-project/boost/blob/f3622c8e803924c5aa905fc1f191cd1a6ea4136a/node/builder.go#L232

The lotus modules code [sets the AgentVersion](https://github.com/filecoin-project/lotus/blob/2a3dcf6cbd1839b7a8242886a289c52457e92699/node/modules/lp2p/host.go#L52) from the lotus build version:
```
	opts := []libp2p.Option{
		libp2p.Identity(pkey),
		libp2p.Peerstore(params.Peerstore),
		libp2p.NoListenAddrs,
		libp2p.Ping(true),
		libp2p.UserAgent("lotus-" + build.UserVersion()),
	}
	for _, o := range params.Opts {
		opts = append(opts, o...)
	}

	h, err := libp2p.New(opts...)
```

We can fix this by writing our own boost UserAgent libp2p option override. See PR https://github.com/filecoin-project/boost/pull/1305

### DIFF
--- a/cmd/boostx/stats_cmd.go
+++ b/cmd/boostx/stats_cmd.go
@@ -159,12 +159,12 @@ var statsCmd = &cli.Command{
 						}
 					}
 
+					lk.Lock()
 					var out string
 					out += "Provider " + maddr.String()
 					if contains(protos, "/fil/storage/mk/1.2.0") {
 						out += " is running boost"
 
-						lk.Lock()
 						boostNodes++
 						boostQualityAdjPower = big.Add(boostQualityAdjPower, minerToMinerPower[maddr].QualityAdjPower)
 						boostRawBytePower = big.Add(boostRawBytePower, minerToMinerPower[maddr].RawBytePower)
@@ -174,10 +174,9 @@ var statsCmd = &cli.Command{
 								transportProtos[p.Name]++
 							}
 						}
-						lk.Unlock()
-
 					} else if contains(protos, "/fil/storage/mk/1.1.0") {
 						out += " is running markets"
+						agentVersions[agentVersion]++
 						marketsNodes++
 					} else {
 						out += " is not running markets or boost"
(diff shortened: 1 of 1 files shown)


cmd/boostx/stats_cmd.go
```diff
@@ -78,68 +94,124 @@ var statsCmd = &cli.Command{
...
+					var out string
+					out += "Provider " + maddr.String()
+					if contains(protos, "/fil/storage/mk/1.2.0") {
+						out += " is running boost"
+
+						lk.Lock()
+						boostNodes++
+						boostQualityAdjPower = big.Add(boostQualityAdjPower, minerToMinerPower[maddr].QualityAdjPower)
+						boostRawBytePower = big.Add(boostRawBytePower, minerToMinerPower[maddr].RawBytePower)
+						agentVersions[agentVersion]++
```
jacobheun, line 171: I think it might be beneficial to count all agent versions, not just boost versions. Right now we only have an indication of Boost versus non boost nodes, but doing a full agent version would give us a better breakdown of the non boost versions. We could just do market nodes for now, as checking the non market nodes isn't too helpful for boost planning purposes.

jacobheun (approved):
Other than the agentversion comment the rest of this looks good.

PR #1305, opened 2023-03-20
Set user agent from boost build version instead of lotus

dirkmc:
TODO:
- [x] Test on devnet

```
$ boost provider libp2p-info t01000
Provider: t01000
Agent: boost-1.6.1-rc1
...
````


jacobheun (approved):
Nice! I wasn't sure how the libp2p agent detection was setup, good to know.

Issue #1277, opened 2023-03-14
Cancel deals which are past start epoch and are in retry state


LexLuthr:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/boost/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.

### Boost component

- [X] boost daemon - storage providers
- [ ] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### Improvement Suggestion

Deals should auto-cancel after start epoch if they are in retry state to clean things up

### References
- [User report in filecoin Slack](https://filecoinproject.slack.com/archives/C03CKDLEWG1/p1678743699162409)

LexLuthr:
Closing https://github.com/filecoin-project/boost/issues/1279 and adding additional action item to cancel offline deals waiting for data past the start epoch as well.

PR #1319, opened 2023-03-22
feat: fail deal if start epoch passed

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1277 partially as we are failing deals during start up and when deal execution process starts.

### DIFF
--- a/storagemarket/deal_execution.go
+++ b/storagemarket/deal_execution.go
@@ -50,6 +50,11 @@ func (p *Provider) runDeal(deal *types.ProviderDealState, dh *dealHandler) {
 		p.saveDealToDB(dh.Publisher, deal)
 	}
 
+	// Fail offline deals or retry early if start epoch has passed
+	if err := p.checkDealProposalStartEpoch(deal); err != nil {
+		p.failDeal(dh.Publisher, deal, err, false)
+	}
+
 	// Execute the deal
 	err := p.execDeal(deal, dh)
 	if err == nil {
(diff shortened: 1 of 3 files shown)


storagemarket/provider.go
```diff
@@ -451,6 +451,11 @@ func (p *Provider) Start() error {
 			continue
 		}
 
+		// Fail deals if start epoch has passed
+		if err := p.checkDealProposalStartEpoch(deal); err != nil {
+			p.failDeal(dh.Publisher, deal, err, false)
+		}
+
```
dirkmc, line 458: I think there's a continue statement missing here.

Also failDeal should be called inside a go-routine. It will send a message to the main event loop, and that will block indefinitely if it's not called in a go-routine (because at this point the event loop hasn't started yet).

If possible we should try to add a test to catch these kinds of bugs.

storagemarket/deal_execution.go
```diff
@@ -50,6 +50,11 @@ func (p *Provider) runDeal(deal *types.ProviderDealState, dh *dealHandler) {
 		p.saveDealToDB(dh.Publisher, deal)
 	}
 
+	// Fail offline deals or retry early if start epoch has passed
+	if err := p.checkDealProposalStartEpoch(deal); err != nil {
+		p.failDeal(dh.Publisher, deal, err, false)
+	}
+
```
dirkmc, line 57: I don't think we need this. IIUC the call to failDeal inside provider.Start makes this one redundant.

LexLuthr: This would be useful when offline deal import is called or retry is called manually. They won't go via the start route.

dirkmc (approved):

PR #1327, opened 2023-03-24
chore:add graphql server listen address config

rickiey:


node/config/def.go
```diff
@@ -63,7 +63,8 @@ func DefaultBoost() *Boost {
 		},
 
 		Graphql: GraphqlConfig{
-			Port: 8080,
+			ListenAddress: "127.0.0.1",
```
jacobheun, line 66:
```suggestion
			ListenAddress: "0.0.0.0",
```

node/config/types.go
```diff
@@ -73,6 +73,8 @@ type WalletsConfig struct {
 }
 
 type GraphqlConfig struct {
+	// ListenAddress: default: 127.0.0.1
```
jacobheun, line 76:
```suggestion
	// The ip address the GraphQL server will bind to. Default: 0.0.0.0
```

jacobheun (commented):
Thanks for the PR! Overall supportive of the change, but to avoid a change in default behavior we should keep the bind address consistent.

PR #1352, opened 2023-04-04
feat: query UI

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1326

Unfortunately, it doesn't look pretty when we have long error strings. But it works as expected. We are now seeing failed queries. 
Screenshot 2023-04-04 at 10 22 24 PM

### DIFF
--- a/retrievalmarket/rtvllog/db.go
+++ b/retrievalmarket/rtvllog/db.go
@@ -227,7 +227,6 @@ func (d *RetrievalLogDB) list(ctx context.Context, offset int, limit int, where 
 
 		dealState.PayloadCID, err = cid.Parse(payloadCid.String)
 		if err != nil {
-			//return nil, fmt.Errorf("parsing payload cid '%s': %w", payloadCid.String, err)
 			dealState.PayloadCID = cid.Undef
 		}
 
--- a/retrievalmarket/rtvllog/db.go
+++ b/retrievalmarket/rtvllog/db.go
@@ -227,7 +227,8 @@ func (d *RetrievalLogDB) list(ctx context.Context, offset int, limit int, where 
 
 		dealState.PayloadCID, err = cid.Parse(payloadCid.String)
 		if err != nil {
-			return nil, fmt.Errorf("parsing payload cid '%s': %w", payloadCid.String, err)
+			//return nil, fmt.Errorf("parsing payload cid '%s': %w", payloadCid.String, err)
+			dealState.PayloadCID = cid.Undef
 		}
 
 		if pieceCid.Valid && pieceCid.String != "" {
--- a/retrievalmarket/rtvllog/retrieval_log.go
+++ b/retrievalmarket/rtvllog/retrieval_log.go
@@ -61,31 +61,27 @@ func (r *RetrievalLog) OnQueryEvent(evt retrievalmarket.ProviderQueryEvent) {
 		"err", evt.Error)
 
 	// Log failures to DB
+	st := &RetrievalDealState{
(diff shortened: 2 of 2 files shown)


retrievalmarket/rtvllog/db.go
```diff
@@ -227,7 +227,8 @@ func (d *RetrievalLogDB) list(ctx context.Context, offset int, limit int, where
 
 		dealState.PayloadCID, err = cid.Parse(payloadCid.String)
 		if err != nil {
-			return nil, fmt.Errorf("parsing payload cid '%s': %w", payloadCid.String, err)
+			//return nil, fmt.Errorf("parsing payload cid '%s': %w", payloadCid.String, err)
+			dealState.PayloadCID = cid.Undef
```
dirkmc, line 231: Could you explain this change?

retrievalmarket/rtvllog/retrieval_log.go
```diff
@@ -58,6 +59,30 @@ func (r *RetrievalLog) OnQueryEvent(evt retrievalmarket.ProviderQueryEvent) {
...
+	if evt.Error != nil {
+		st.Status = fmt.Sprintf("failed to respond to the query: %s", evt.Error.Error())
+	} else {
+		if evt.Response.Status == retrievalmarket.QueryResponseUnavailable {
+			st.Status = "retrieval query offer was unavailable"
+		}
+		if evt.Response.Status == retrievalmarket.QueryResponseError {
+			st.Status = "retrieval query offer errored"
+		}
+	}
```
dirkmc, line 77: The status should just be a short string like "Failed" or "Unavailable". The Message contains the actual error message.

LexLuthr: I will update the strings

retrievalmarket/rtvllog/db.go
LexLuthr: If we handle the error by returning then UI fails. We get an error saying "parsing payload failed" with whole page going blank. Thus an empty value.

dirkmc: Ok makes sense - let's just remove the commented out line then

retrievalmarket/rtvllog/retrieval_log.go
```diff
@@ -58,6 +58,30 @@ func (r *RetrievalLog) OnQueryEvent(evt retrievalmarket.ProviderQueryEvent) {
...
 		"err", evt.Error)
+
+	// Log failures to DB
+	st := &RetrievalDealState{
+		UnsealPrice: evt.Response.UnsealPrice,
+		Message:     evt.Response.Message,
+	}
+	if evt.Error != nil {
+		st.Status = evt.Error.Error()
+	} else {
```
dirkmc, line 69: In the case that evt.Error is not nil, let's
- set st.Status to "Failed"
- check if evt.Response.Message is set
  - if not, set st.Message to evt.Error.Error()

LexLuthr: I think this will cause us to log all queries (even the successful ones as well). If evt.Error is not nil then response itself was successful but we might send "not available" or "error in query" responses along with "available" response. I would have no condition to decide when to log and when to skip. 
From the code it looks like evt.Response.Message is always set for query event and evt.Error.Error() would be empty for successful responses when "not available" or "error in query" message is being sent. Overriding the message value will result in los of the real reason for failure.

dirkmc (approved):

PR #1355, opened 2023-04-06
fix: correct link for download root block

jacobheun:
This fixes the rendered HTML to use the correct root cid block, when the root cid is show on the inspect page. Previously whatever the search query was (piece cid, root cid, or payload cid) would be used.

Screenshot of the correct html, you can see the label and cid url match

Screenshot 2023-04-06 at 12 41 41 PM


dirkmc (approved):
Thanks!

PR #1356, opened 2023-04-06
chore: add support for multiple node.js versions in makefile

jacobheun:
We've supported building react with Node.js 18 for some time but the Makefile never got fixed. Rather than just bumping the major version to 18 and breaking the build for users still on Node 16, this allows us to specify multiple known good versions. Later we can phase 16 out and 20 in when node 16 hits end of life this year.

A supported node version will show:
```
$ make react
Node.js version v18.12.1 is supported.
npm_config_legacy_peer_deps=yes npm ci --no-audit --prefix react
...
```

and unsupported:
```
$ make react
Unsupported Node.js version. Please install one of the following versions: 16 18
exit 1
make: *** [validate-node-version] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,12 +13,13 @@ $(warning Your Golang version is go$(shell expr $(GOVERSION) / 1000000).$(shell 
 $(error Update Golang to version to at least 1.16.0)
 endif
 
-LTS_NODE_VER=16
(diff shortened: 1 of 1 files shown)


dirkmc (approved):
Nice 🙌

Issue #1366, opened 2023-04-12
Running boost fails with missing deal-staging dir


dirkmc:
After initializing boost, running boost fails with `<boost repo>/deal-staging: no such file or directory`

PR #1368, opened 2023-04-12
fix boost run missing staging-area dir

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/1366

TODO:
- [x] Test on local devnet

### DIFF
--- a/node/modules/legacy_markets.go
+++ b/node/modules/legacy_markets.go
@@ -2,6 +2,8 @@ package modules
 
 import (
 	"context"
+	"fmt"
+	"os"
 	"path/filepath"
 
 	piecefilestore "github.com/filecoin-project/boost-gfm/filestore"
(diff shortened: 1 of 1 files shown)


jacobheun (approved):
There is a patch for this that got added temporarily in the LID work, https://github.com/filecoin-project/boost/blob/224a149/docker/devnet/boost/entrypoint.sh#L48-L49. We likely hit this issue when we ported some code over. 

Did your test with the docker devnet, or just a local devnet? Those lines above shouldn't be needed anymore.

nonsense (approved):

dirkmc:
I tested with the local devnet - but a user reported this in production, so I think we do need it

jacobheun:
Okay, just wanted to make sure the test wasn't polluted by that patch in the dockerfile. We can remove that separately.

PR #1370, opened 2023-04-12
merge: main to lid

jacobheun:
Main introduced metrics on the remoteblockstore. In LID, since graphsync also now uses the remoteblockstore, but it already handles its own metrics as stats behave differently for graphsync due to sessions. We might want to improve overall tracking for this in the future, but for now we just need metrics in the remoteblockstore to be optional.

Key files being modified for this merge:
- cmd/lib/remoteblockstore/remoteblockstore.go
- node/modules/graphsync.go

TODO: 
- [x] Update remoteblockstore and graphsync.go to handle metrics appropriately
- [x] Get tests passing
- [x] Smoke tests on devnet


dirkmc (approved):

jacobheun:
I needed to make updates to the docker devnet to validate this locally, as well as patch the listen address on the embedded LID service for the bitswap and http containers to be able to talk to it, but after that I was able to successfully retrieve new deals across via bitswap, graphsync and http. I'm going to merge this as is, and do a follow up PR into the LID branch that includes the devnet fixes (which are only required for bitswap and http services to work correctly)

PR #1375, opened 2023-04-13
fix: make devnet work for lid

jacobheun:
## Summary
This PR updates docker devnet in the LID (local index directory) branch so that booster-http and booster-bitswap are once again able to run.

_Note_: No changes have been made to the boostd api itself, the only change is in the boostd-data service mentioned below.

* BREAKING CHANGE: The boostd-data service now takes an `--addr` flag instead of `--port`, so that it can be bound to non localhost IP addresses. This is needed for cross container communication, and may be useful in larger scale deployments. - [919cb57](https://github.com/filecoin-project/boost/pull/1375/commits/919cb578b752017dd16f941e7c127367af6d22f8)
* Fixed linting in extern/boostd-data. We currently aren't linting the "sub project" in CI so linting was failing here. I separate this as it's own commit [6294ae0](https://github.com/filecoin-project/boost/pull/1375/commits/6294ae0d4bf7decd4a04e1297ad48f37597d5fe9) to make it easier to see the functional changes.
* [10fbdf4](https://github.com/filecoin-project/boost/pull/1375/commits/10fbdf4e2abdee1ad6c7aee509f9d091bcbadb2c) includes the updates to the docker devnet, as well as the needed invocation changes to Boostd for the piecedirectory.

## Remaining Items
- [X] Verify graphsync retrieval in devnet
- [X] Verify http piece and gateway retrieval in devnet
- [x] Verify devnet with bitswap

### DIFF
--- a/extern/boostd-data/couchbase/service.go
+++ b/extern/boostd-data/couchbase/service.go
@@ -46,11 +46,11 @@ func (s *Store) Start(ctx context.Context) error {
 func (s *Store) AddDealForPiece(ctx context.Context, pieceCid cid.Cid, dealInfo model.DealInfo) error {
 	log.Debugw("handle.add-deal-for-piece", "piece-cid", pieceCid)
 
-	ctx, span := tracing.Tracer.Start(context.Background(), "store.add_deal_for_piece")
+	ctx, span := tracing.Tracer.Start(ctx, "store.add_deal_for_piece")
 	defer span.End()
 
 	defer func(now time.Time) {
-		log.Debugw("handled.add-deal-for-piece", "took", fmt.Sprintf("%s", time.Since(now)))
+		log.Debugw("handled.add-deal-for-piece", "took", time.Since(now))
 	}(time.Now())
 
 	return s.db.AddDealForPiece(ctx, pieceCid, dealInfo)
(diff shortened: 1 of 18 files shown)


extern/boostd-data/ldb/db.go
```diff
@@ -322,7 +322,7 @@ func (db *DB) MarkIndexErrored(ctx context.Context, pieceCid cid.Cid, err error)
 	}
 
 	md.Error = err.Error()
-	md.ErrorType = fmt.Sprintf("%T", err)
+	md.ErrorType = fmt.Sprintf("%T", sourceErr)
```
dirkmc, line 325: Nice catch 👍 
I think we also need to change the previous line:
md.Error = sourceErr.Error()

extern/boostd-data/couchbase/service.go
```diff
@@ -46,11 +46,11 @@ func (s *Store) Start(ctx context.Context) error {
 func (s *Store) AddDealForPiece(ctx context.Context, pieceCid cid.Cid, dealInfo model.DealInfo) error {
 	log.Debugw("handle.add-deal-for-piece", "piece-cid", pieceCid)
 
-	ctx, span := tracing.Tracer.Start(context.Background(), "store.add_deal_for_piece")
+	ctx, span := tracing.Tracer.Start(ctx, "store.add_deal_for_piece")
 	defer span.End()
 
 	defer func(now time.Time) {
-		log.Debugw("handled.add-deal-for-piece", "took", fmt.Sprintf("%s", time.Since(now)))
```
dirkmc, line 53: I'm not sure we want to replace `fmt.Sprintf("%s", time.Since(now))` with time.Since(now)
If we don't do the sprintf it will always output the number of nanoseconds, whereas if it converts it to a string it will output a humanized string like "100ms"

dirkmc (commented):
A couple of small suggestions, but overall looks great 👍

extern/boostd-data/ldb/db.go
jacobheun: You're right, good catch.

extern/boostd-data/couchbase/service.go
jacobheun: Ah, good to know, I'll take a look at this. These were all changes to appease the linter, will find a better way to resolve this.

dirkmc: I think we can probably just do time.Since(now).String()

jacobheun: It does indeed, sample output - `{"took": "2.840023375s"},{"took": "365.096458ms"}` _(not actual operation times)_. The previous one displayed a float of seconds regardless of duration.

jacobheun:
Verified full car and block retrieval in booster-bitswap. This is all set.

PR #1377, opened 2023-04-15
feat: redeclare storage when unsealed state is checked

jacobheun:
## Summary
Adds to https://github.com/filecoin-project/boost/pull/1191 to call storage redeclare on the lotus-miner. Lotus doesn't currently automatically detect fs changes, so this is needed to ensure we're able to automatically keep indexes and their metadata up to date.

Speaking with the Lotus team, this call should be cheap so running it regularly shouldn't be an issue.

### Screenshot from devnet
This is a screenshot of the before and after `lotus-miner storage list`. On the left the list shows 2 of 4 sectors as being unsealed, but we can see by the screenshot that there is only 1 unsealed sector. After the logs of the job execute (on the right), the storage list then correctly shows 1 of 4 sectors as unsealed.

Screenshot 2023-04-15 at 1 48 27 PM

## Note on Frequency of this check
It may be useful to allow configuring how often the unseal manager is run. Currently it runs every hour, but in reality it doesn't need to run that frequently. 12 or even 24h is likely a more reasonable default as we ultimately just care about eventual consistency and hour to hour changes aren't likely to be very frequent.

Edit: I set this to 12 hours for now, instead of the current 1 hour.

## Remaining Work
- [X] Add configuration so that SP's can disable this in case they want to control when redeclares occur (default to enabled)

### DIFF
--- a/indexprovider/unsealedstatemanager.go
+++ b/indexprovider/unsealedstatemanager.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/filecoin-project/boost-gfm/storagemarket"
 	"github.com/filecoin-project/boost/db"
+	"github.com/filecoin-project/boost/node/config"
 	"github.com/filecoin-project/go-address"
 	cborutil "github.com/filecoin-project/go-cbor-util"
 	"github.com/filecoin-project/go-state-types/abi"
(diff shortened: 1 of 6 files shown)


nonsense (approved):

dirkmc (approved):
Looks great 👐 

If it's cheap to run we might as well run it frequently, I'd say once per hour is a good cadence.

jacobheun:
Moved this back to the original hour as default.

PR #1410, opened 2023-05-02
local index directory: recover tool

nonsense:
This tool is a helper to rebuild the LID based only on on-chain data and unsealed sector files.

Fixes https://github.com/filecoin-project/boost/issues/1394

Usage:
```
boostd recover lid --api-fullnode=$FULLNODE_API_INFO  --api-storage=$MINER_API_INFO --ignore-commp
```

If you remove `--ignore-commp` the tool would locally re-calculate CommP from the local data and confirm that it matches the deals on-chains.

-----

TODO:
- [ ] run on `sofiaminer` and inspect report and results
- [ ] run on `filcollins` and inspect report and results
- [ ] write docs on how to use the tool, how long it takes to run, how to interpret report/errors, etc.

### DIFF
--- a/cmd/boostd/recover.go
+++ b/cmd/boostd/recover.go
@@ -340,8 +340,7 @@ func NewDisasterRecovery(ctx context.Context, dir, repodir string) (*DisasterRec
 		Sectors: make(map[uint64]*SectorStatus),
 	}
 
-	// Create a logger for the migration that outputs to a file in the
-	// current working directory
+	// logger for the recovery that outputs to a file and stdout
 	logger, err = createLogger(fmt.Sprintf("%s/output-%d.log", drr.Dir, time.Now().UnixNano()))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


dirkmc (commented):
Looks really good, nice work! 🙌 

I'd suggest we put it under boostd somewhere. Maybe `boostd recover`.

It would also be nice to refactor so that code shared between the migrate commands and this command lives in a common place.

cmd/boostd/recover.go
```diff
@@ -0,0 +1,791 @@
...
+		},
+		&cli.StringFlag{
+			Name:     "api-storage",
+			Usage:    "the endpoint for the storage node API",
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:  "api-lid",
+			Usage: "the endpoint for the LID API",
+			//Required: true,
```
dirkmc, line 84: Is this required in the end?

dirkmc (approved):

dirkmc:
Nice work! LGTM 👍

Issue #1398, opened 2023-04-27
Boost reports incorrect state for failed snap-deals


LexLuthr:
A sectors was being snapped with 2 pieces (filcolins)

Deal 59824cbd-2686-47d2-b285-8b820f99db57
Deal dc2d03f3-17df-4263-af88-e5bfbfd09898

Miner lost access to storage and failed windowPost. This put the sector status as "Not Active"

ID    State          OnChain  Active  Expiration                    Deals  DealWeight  VerifiedPower 
155   Proving        YES      NO      4359832 (in 1 year 24 weeks)  CC                          

When sector was doing FRU (Finalize Replica Update), it noted that sector is no longer active and thus should not be upgraded.

```
890.    2023-04-27 05:31:57 -0400 EDT:  [event;sealing.SectorRetryFinalize]     {"User":{}}
891.    2023-04-27 05:33:35 -0400 EDT:  [event;sealing.SectorFinalized] {"User":{}}
892.    2023-04-27 05:33:35 -0400 EDT:  [event;sealing.SectorSubmitReplicaUpdateFailed] {"User":{}}
893.    2023-04-27 05:34:35 -0400 EDT:  [event;sealing.SectorAbortUpgrade]      {"User":{}}
        sector marked for upgrade 155 no longer active, aborting upgrade
```

Then lotus-miner reverted the snap-deal and set the sector state to proving again.

Since, we only check sector states, our pipeline incorrectly assumed that deal has been successfully sealed by the miner.

```

2023-04-27 15:03:37.490 | 1m | current sealing statestate: ReplicaUpdateFailed
-- | -- | --
2023-04-27 15:04:37.491 | 1m | current sealing statestate: AbortUpgrade
2023-04-27 15:05:37.483 | 1m | current sealing statestate: Proving
2023-04-27 15:05:37.485 | 2ms | deal sealing reached termination state
2023-04-27 15:05:37.486 | 1ms | deal go-routine finished execution
```

We should have failed this deal in Boost. SP is going to end up paying penalty for failure to seal the deal successfully but Boost reports it as success.

This can have multiple impacts:

1. Incorrect QAP for SP (Chain vs Boost)
2. Unseen penalties (Deal is reported as success)
3. Incorrect Index announcement as lotus-miner cleans up the unsealed copy of the sector
4. Client will not be able to get their data back even though they will see that deal was successful

PR #1419, opened 2023-05-04
fix: fail deal if snap-deal sectors fails sealing

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1398
- [x] Update UI
- [x] Update deal flow - fail the deal
- [x] Add Test

### DIFF
--- a/gql/resolver.go
+++ b/gql/resolver.go
@@ -601,15 +601,12 @@ func (dr *dealResolver) sealingState(ctx context.Context) string {
 		log.Warnw("error getting sealing status for sector", "sector", dr.SectorID, "error", err)
 		return "Sealer: Sealing"
 	}
-	if len(si.Deals) > 0 {
-		for _, d := range si.Deals {
-			if d == dr.ProviderDealState.ChainDealID {
-				return "Sealer: " + string(si.State)
-			}
+	for _, d := range si.Deals {
+		if d == dr.ProviderDealState.ChainDealID {
+			return "Sealer: " + string(si.State)
 		}
-		return "Sealer: failed - deal not found in sector"
 	}
-	return "Sealer: failed - deal not found in sector"
+	return fmt.Sprintf("Sealer: failed - deal not found in sector %d", si.SectorID)
 }
 
 func (dr *dealResolver) Logs(ctx context.Context) ([]*logsResolver, error) {
--- a/gql/resolver.go
+++ b/gql/resolver.go
@@ -601,8 +601,15 @@ func (dr *dealResolver) sealingState(ctx context.Context) string {
 		log.Warnw("error getting sealing status for sector", "sector", dr.SectorID, "error", err)
 		return "Sealer: Sealing"
 	}
-
-	return "Sealer: " + string(si.State)
+	if len(si.Deals) > 0 {
+		for _, d := range si.Deals {
+			if d == dr.ProviderDealState.ChainDealID {
+				return "Sealer: " + string(si.State)
+			}
+		}
+		return "Sealer: failed - deal not found in sector"
+	}
+	return "Sealer: failed - deal not found in sector"
 }
 
 func (dr *dealResolver) Logs(ctx context.Context) ([]*logsResolver, error) {
--- a/storagemarket/deal_execution.go
+++ b/storagemarket/deal_execution.go
@@ -129,7 +129,9 @@ func (p *Provider) execDeal(deal *smtypes.ProviderDealState, dh *dealHandler) (d
 
 	// Watch the sealing status of the deal and fire events for each change
 	p.dealLogger.Infow(deal.DealUuid, "watching deal sealing state changes")
-	p.fireSealingUpdateEvents(dh, deal.DealUuid, deal.SectorID)
+	if derr := p.fireSealingUpdateEvents(dh, deal.DealUuid, deal.SectorID); derr != nil {
+		return derr
+	}
 	p.cleanupDealHandler(deal.DealUuid)
 	p.dealLogger.Infow(deal.DealUuid, "deal sealing reached termination state")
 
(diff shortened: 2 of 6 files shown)


gql/resolver.go
```diff
@@ -601,8 +601,15 @@ func (dr *dealResolver) sealingState(ctx context.Context) string {
...
 	}
-
-	return "Sealer: " + string(si.State)
+	if len(si.Deals) > 0 {
+		for _, d := range si.Deals {
+			if d == dr.ProviderDealState.ChainDealID {
+				return "Sealer: " + string(si.State)
+			}
+		}
+		return "Sealer: failed - deal not found in sector"
```
dirkmc, line 610: I'd suggest we include the sector number in the error message here to make it easier for the the user to go and check on the sector status
```suggestion
		return fmt.Sprintf("Sealer: failed - deal not found in sector %d", si.SectorID)
```

gql/resolver.go
```diff
@@ -601,8 +601,15 @@ func (dr *dealResolver) sealingState(ctx context.Context) string {
 		log.Warnw("error getting sealing status for sector", "sector", dr.SectorID, "error", err)
 		return "Sealer: Sealing"
 	}
-
-	return "Sealer: " + string(si.State)
+	if len(si.Deals) > 0 {
```
dirkmc, line 604: I don't think you actually need the if statement.
I believe `for _, d := range si.Deals` will just run zero times if si.Deals is nil

storagemarket/deal_execution.go
```diff
@@ -685,66 +687,88 @@ func (p *Provider) registerShardSync(ctx context.Context, pc cid.Cid) error {
...
-			return ""
+			return lapi.SectorInfo{}
 		}
 
 		// Get the sector status
 		si, err := p.sps.SectorsStatus(p.ctx, sectorNum, false)
-		if err == nil && si.State != lastSealingState {
-			lastSealingState = si.State
+		if err == nil && si.State != sectorInfo.State {
+			sectorInfo.State = si.State
```
dirkmc, line 703: This is a little tricky to understand - I'd suggest keeping the lastSealingState variable and leaving these lines as they were

storagemarket/deal_execution.go
```diff
@@ -685,66 +687,88 @@ func (p *Provider) registerShardSync(ctx context.Context, pc cid.Cid) error {
...
+	if IsFinalSealingState(info.State) {
+		if len(info.Deals) > 0 {
+			for _, d := range info.Deals {
+				if d == deal.ChainDealID {
+					return nil
+				}
+			}
+			return retErr
+		}
+		return retErr
```
dirkmc, line 734: This code is repeated several times in this PR. I'd suggest refactoring the code into a function that can be called from all those places.
Or maybe IsFinalSealingState should return `(bool, error)`? I guess try it out and see which makes the code simpler.

storagemarket/deal_execution.go
```diff
@@ -685,66 +687,88 @@ func (p *Provider) registerShardSync(ctx context.Context, pc cid.Cid) error {
...
 	}
 
-	// Check status every second
-	ticker := time.NewTicker(time.Second)
+	// Check status every 5 second. There is no advantage of checking it every second
+	ticker := time.NewTicker(5 * time.Second)
```
dirkmc, line 738: I'd suggest we change this to 10 seconds

storagemarket/deal_execution.go
```diff
@@ -685,66 +687,88 @@ func (p *Provider) registerShardSync(ctx context.Context, pc cid.Cid) error {
...
 		case <-p.ctx.Done():
-			return
+			return nil
 		case <-ticker.C:
 			count++
 			// Force a status check every forceCount seconds, even if there
 			// are no subscribers (so that we can stop checking altogether
 			// if the sector reaches a final sealing state)
-			state := checkStatus(count >= forceCount)
+			info = checkStatus(count >= forceCount)
```
dirkmc, line 751: The comment here needs to change:
if the length of the tick changes, the frequency of doing the force check will also change. eg if the tick length is 10 seconds, the frequency of the force check will be every 10 minutes.

storagemarket/provider.go
```diff
@@ -45,6 +45,7 @@ import (
 var (
 	ErrDealNotFound        = fmt.Errorf("deal not found")
 	ErrDealHandlerNotFound = errors.New("deal handler not found")
+	ErrDealNotSealed       = errors.New("deal could not be found is the sector")
```
dirkmc, line 48:
```suggestion
	ErrDealNotSealed       = errors.New("deal storage failed - deal not found in sector")
```

storagemarket/deal_execution.go
LexLuthr: I have removed the initial if condition and shorten this code to 3 lines. Inserting this check in IsFinalSealingState() will make things more complicated** as we will need to supply the ProvideDealState to the function as well. Check the new code once and let me know if it is cleaner. Otherwise, I will make the changes.

**IsFinalSealingState() is also being used outside the deal_execution now

storagemarket/deal_execution.go
LexLuthr: I am updating forceCount to every 2 minutes and updating the comment how to calculate it.

storagemarket/provider_test.go
```diff
@@ -1648,15 +1669,24 @@ func NewHarness(t *testing.T, opts ...harnessOpt) *ProviderHarness {
...
+//	if err != nil {
+//		return lapi.SectorInfo{}
+//	}
+//	return lapi.SectorInfo{
+//		SectorID: deal.SectorID,
+//		State:    lapi.SectorState(sealing.Proving),
+//		Deals:    []abi.DealID{deal.ChainDealID},
+//	}
+//}
+
```
dirkmc, line 1689: Let's remove this unused code

storagemarket/lp2pimpl/net.go
```diff
@@ -325,12 +325,30 @@ func (p *DealProvider) getDealStatus(req types.DealStatusRequest) types.DealStat
 		return errResp("getting sector status from sealer")
 	}
 
+	sealingStatus := string(si.State)
+	var found bool
+
+	if storagemarket.IsFinalSealingState(si.State) {
+		if len(si.Deals) > 0 {
```
dirkmc, line 332: This looks similar to the code in deal_execution that checks if the list of deals contains a particular deal id.
Maybe we can refactor this code into a function that can be called from all three places like
```storagemarket.HasDeal(si.Deals, pds.ChangeDealID)```

File: storagemarket/lp2pimpl/net.go [hunk 10]
```diff
@@ -325,12 +325,23 @@ func (p *DealProvider) getDealStatus(req types.DealStatusRequest) types.DealStat
...
+	var found bool
+
+	if storagemarket.IsFinalSealingState(si.State) {
+		found = storagemarket.HasDeal(si.Deals, pds.ChainDealID)
+	}
+
+	if !found {
+		sealingStatus = storagemarket.ErrDealNotFound.Error()
+	}
+
```
dirkmc, line 338: If the deal has not reached the proving state, it looks like this code will always set the SealingStatus to ErrDealNotFound

dirkmc (approved):

PR #1421, opened 2023-05-04
car size required only for online deals

anjor:
I was confused about something here for “offline” deals.

[car-size is a required parameter](https://github.com/filecoin-project/boost/blob/main/cmd/boost/deal_cmd.go#L50)
[which is not allowed to be zero](https://github.com/filecoin-project/boost/blob/main/cmd/boost/deal_cmd.go#L196-L199)
but [transfer](https://github.com/filecoin-project/boost/blob/main/cmd/boost/deal_cmd.go#L200) is zeroed out for offline deals.

This PR means car-size is no longer checked unless it's an online deal.

### DIFF
--- a/cmd/boost/deal_cmd.go
+++ b/cmd/boost/deal_cmd.go
@@ -47,7 +47,7 @@ var dealFlags = []cli.Flag{
 	&cli.Uint64Flag{
 		Name:     "car-size",
 		Usage:    "size of the CAR file",
-		Required: true,
+		Required: false,
 	},
 	&cli.StringFlag{
 		Name:     "payload-cid",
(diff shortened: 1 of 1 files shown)


dirkmc (commented):
I think we'd also need to change the parameter definition so that the parameter is no longer marked as "required"

dirkmc:
Sorry looking at this again I think it would make more sense to
- have a required "car-size" parameter for the "deal" command
- remove the "car-size" parameter for the "offline-deal" command

anjor:
Yeah that's cleaner. Done.

dirkmc (approved):

dirkmc:
Thanks, could you confirm that you were able to test the two code flows (online vs offline deal)?

dirkmc:
@anjor confirmed offline that this new code has been tested

PR #1422, opened 2023-05-04
deps: Update go-libp2p to v0.27.1

MarcoPolo:
I simply ran `go get github.com/libp2p/go-libp2p` to update to the latest v0.27.1. Lots of updates here (https://github.com/libp2p/go-libp2p/releases) including fixing a nasty busy loop bug in the autorelay module (see https://github.com/libp2p/go-libp2p/pull/2208).

PR #1428, opened 2023-05-05
Update to use packages in go-libipni

gammazero:
Replaces use of go-legs and storetheindex with go-libipni.

Blocked waiting for PR #1422


gammazero:
no longer needed

PR #1444, opened 2023-05-10
Change service GetIndex / AddIndex to return channel instead of array

dirkmc:


cmd/migrate-lid/couch-to-yuga.go
```diff
@@ -211,16 +212,26 @@ func migrateLidToLidIndex(ctx context.Context, pieceCid cid.Cid, source StoreMig
 	}
 
 	// Load the index from the source store
-	records, err := source.GetIndex(ctx, pieceCid)
+	idx, err := source.GetIndex(ctx, pieceCid)
```
nonsense, line 215: I wonder if it won't be a better idea to change the signature to receive a channel as an argument. This way the caller can control if they want to have a buffered or unbuffered channel as a receiver and also decide how to control blocking.

Not sure if this is super relevant here though.

nonsense: Then again I am not sure if the reflection magic would work to convert the interface correctly to the JSON streaming API...

dirkmc: I had a look at the jsonrpc docs, I don't believe channel as a parameter is supported:
https://github.com/filecoin-project/go-jsonrpc#supported-function-signatures

PR #1452, opened 2023-05-17
Add option to serve index provider ads over http

dirkmc:


node/config/def.go
```diff
@@ -171,14 +171,19 @@ func DefaultBoost() *Boost {
...
 			Enable:               true,
 			EntriesCacheCapacity: 1024,
 			EntriesChunkSize:     16384,
 			// The default empty TopicName means it is inferred from network name, in the following
 			// format: "/indexer/ingest/<network-name>"
 			TopicName:         "",
 			PurgeCacheOnStart: false,
+
+			HttpEndpoint: IndexProviderHttpEndpointConfig{
+				PublicIP: "",
```
masih, line 184: Rename to something that is "IP" agnostic? Maybe `ExternalHostName` or something like that? not sure.

node/config/types.go
```diff
@@ -279,6 +279,50 @@ type ContractDealsConfig struct {
...
+	// in following format: '/indexer/ingest/<network-name>'
+	// Defaults to empty, which implies the topic name is inferred from network name.
+	TopicName string
+
+	// PurgeCacheOnStart sets whether to clear any cached entries chunks when the provider engine
+	// starts. By default, the cache is rehydrated from previously cached entries stored in
+	// datastore if any is present.
+	PurgeCacheOnStart bool
+
+	HttpEndpoint IndexProviderHttpEndpointConfig
```
masih, line 311: minor: Use the term `Publisher` instead of `Endpoint`.

at some point: replace `IndexProvider` across the repo with `IPNI` for consistency with deal flags and terminology used elsewhere?

node/config/types.go
```diff
@@ -279,6 +279,50 @@ type ContractDealsConfig struct {
...
+}
+
+type IndexProviderHttpEndpointConfig struct {
+	// Set the public IPv4 endpoint for the index provider listener.
+	// eg "82.129.73.111"
+	// This is usually the same as the public IP for the boost node.
+	// If PublicIP is the empty string, requests are served over graphsync
+	// instead.
+	PublicIP string
```
masih, line 320: To be clear this ends up being included in the HTTP announce messages as _multiaddr_, which can be dns name, etc. as well as an IPv4. We already see SPs that are using dns names in their setup so forcing this to be an IP might come back to haunt us for SPs that have name servers setup and use DNS for routing.

[This](https://github.com/ipni/go-libipni/blob/afe2d8ea45b86c2a22f756ee521741c8f99675e5/maurl/convert.go#L106) utility function might be handy to work around this.

node/modules/storageminer_idxprov.go
```diff
@@ -93,13 +93,25 @@ func IndexProvider(cfg config.IndexProviderConfig) func(params IdxProv, marketHo
...
 				engine.WithTopic(t),
 			)
+
+			// Advertisements can be served over the data transfer protocol
+			// (on graphsync) or over HTTP
+			if cfg.HttpEndpoint.PublicIP != "" {
+				opts = append(opts,
+					engine.WithPublisherKind(engine.HttpPublisher),
+					engine.WithHttpPublisherListenAddr(fmt.Sprintf("0.0.0.0:%d", cfg.HttpEndpoint.Port)),
+					engine.WithHttpPublisherAnnounceAddr(fmt.Sprintf("/ip4/%s/tcp/%d/http", cfg.HttpEndpoint.PublicIP, cfg.HttpEndpoint.Port)),
```
masih, line 109: Ditto re hardcoding `ip4`.

node/modules/storageminer_idxprov.go
```diff
@@ -93,13 +93,25 @@ func IndexProvider(cfg config.IndexProviderConfig) func(params IdxProv, marketHo
 			// The extra data is required by the lotus-specific index-provider gossip message validators.
 			ma := address.Address(maddr)
 			opts = append(opts,
-				engine.WithPublisherKind(engine.DataTransferPublisher),
 				engine.WithDataTransfer(dtV1ToIndexerDT(dt, func() ipld.LinkSystem {
```
masih, line 96: I think we won't need DT config nor gossipsub config if HTTP publisher kind is configured. @LexLuthr @dirkmc can I ask you to give this a quick test just to make sure please?

And if all is good, we can move those two options into the if clause below.

node/modules/storageminer_idxprov.go
```diff
@@ -93,13 +93,25 @@ func IndexProvider(cfg config.IndexProviderConfig) func(params IdxProv, marketHo
...
 					return *e.LinkSystem()
 				})),
 				engine.WithExtraGossipData(ma.Bytes()),
 				engine.WithTopic(t),
 			)
+
+			// Advertisements can be served over the data transfer protocol
+			// (on graphsync) or over HTTP
+			if cfg.HttpEndpoint.PublicIP != "" {
```
masih, line 105: minor: Worth defining a dedicated flag for HTTP publishing? I am trying to think how using the presence of HTTP endpoint as a flag can be error prone.

node/modules/storageminer_idxprov.go
```diff
@@ -77,30 +78,47 @@ func IndexProvider(cfg config.IndexProviderConfig) func(params IdxProv, marketHo
...
-			// no validators by default.
-			t, err := ps.Join(topicName)
-			if err != nil {
-				llog.Errorw("Failed to join indexer topic", "err", err)
-				return nil, xerrors.Errorf("joining indexer topic %s: %w", topicName, err)
-			}
+			// Advertisements can be served over the data transfer protocol
+			// (on graphsync) or over HTTP
+			if cfg.HttpPublisher.Enabled {
+				announceAddr, err := util.ToHttpMultiaddr(cfg.HttpPublisher.PublicHostname, cfg.HttpPublisher.Port)
```
masih, line 88: Would be great if we can log `announceAddr` for debugging purposes.

masih (approved):

node/modules/storageminer_idxprov.go
```diff
@@ -82,6 +82,26 @@ func IndexProvider(cfg config.IndexProviderConfig) func(params IdxProv, marketHo
...
+				engine.WithTopic(t),
+				engine.WithExtraGossipData(ma.Bytes()),
+			)
+			if cfg.Announce.AnnounceOverHttp {
+				opts = append(opts, engine.WithDirectAnnounce(cfg.Announce.DirectAnnounceURLs...))
+			}
+
 			// Advertisements can be served over the data transfer protocol
 			// (on graphsync) or over HTTP
 			if cfg.HttpPublisher.Enabled {
```
masih, line 107: We probably want to check that `PublicHostName` is not empty.

masih: Great; add a test case to `ToHttpMultiaddr`?

dirkmc: Good call, turns out `multiaddr.NewMultiaddr()` accepts an empty hostname 🤔 

So I added an explicit check in [test: add test case for empty announce address hostname](https://github.com/filecoin-project/boost/pull/1452/commits/331879f300a07156ee71d4da1c75d02a31b0791e)

PR #1456, opened 2023-05-18
Add `boostd index announce-latest` command

dirkmc:
```
$ boostd index announce-latest
Announced advertisement with cid baguqeeramxvzsv6txws3mtmahyijaf5trwbxjnkur3k44x7e4mgz57px4d6a

$ boostd index announce-latest-http
Announced advertisement to indexers over http with cid baguqeeras24lritlycfsvd3idryxamitoxh3bm6nbhvbj7xm3o6wv2dhfpmq
```


indexprovider/wrapper.go
```diff
@@ -262,6 +263,14 @@ func (w *Wrapper) IndexerAnnounceAllDeals(ctx context.Context) error {
 	return merr
 }
 
+func (w *Wrapper) IndexerAnnounceLatest(ctx context.Context) (cid.Cid, error) {
+	e, ok := w.prov.(*engine.Engine)
```
masih, line 267: Double checking `w.prov` will never be `nil`?

indexprovider/wrapper.go
```diff
@@ -262,6 +263,14 @@ func (w *Wrapper) IndexerAnnounceAllDeals(ctx context.Context) error {
...
+	if !ok {
+		return cid.Undef, fmt.Errorf("index provider is disabled")
+	}
+	return e.PublishLatest(ctx)
```
masih, line 271: There is a separate function for direct announce over HTTP. Should we check if direct announce is enabled and call that as well as this?

dirkmc: Oh interesting.. I guess that could be a separate cli command so that we can debug both explicitly

masih: or `--direct` flag?

indexprovider/wrapper.go
dirkmc: It will never be nil 👍

indexprovider/wrapper.go
dirkmc: I think I will add a separate command because you also have the option to explicitly specify the url

PR #1462, opened 2023-05-23
piecedirectory: Cache piece readers in BlockstoreGet

magik6k:
On top of the LID branch, as that's what my setup is running.

Depends on https://github.com/filecoin-project/lotus/pull/10913

This PR:
* Adds new piecereader metrics
* Rewrites PieceDirectory.BlockstoreGet to reuse piecereader instances
  * Spawning those is quite expensive (some lotus API calls, a bunch of allocations)
  * New piecereader allows parallel ReadAt calls, this code takes advantage of that.


nonsense:
@magik6k I think we can close this, in favour of https://github.com/filecoin-project/boost/pull/1584

We are now using a shared PieceReader, which increased throughput with about 3x when downloading CAR files via booster-http.

Issue #981, opened 2022-11-21
PieceDirectory deal cleanup


LexLuthr:
1. Expired deals should be cleaned up
2. Slashed deals should be cleaned up
3. Deal details should be removed from piece metadata
4. Index should be removed for pieces where no more deals are left after deal cleanup
5. Announcement should be made to indexer that these MultiHashes are no longer available.

LexLuthr:
- [ ] Implement Basic Removal APIs https://github.com/filecoin-project/boost/pull/953
- [ ] Take a decision on how to track legacy deals in Metadata as it only has UUID
- [ ] Add a cleanup job in Legacy Provider and Boost Provider
- [ ] Update Indexer about removed deals if required

PR #1463, opened 2023-05-23
piece doctor and sector state manager refactor

nonsense:
1. Currently after the first iteration, Piece Doctor never loops to check more pieces. This PR is fixing that bug.

2. Additionally on `filcollins` a lot of the flagged pieces are false-negatives for sectors that are already expired or removed. Because of this, I refactored the `UnsealedStateManager` out of the `indexprovider` package, into `SectorStateMgr` that keeps track of the state of sectors and added a check against active sectors into the Piece Doctor.

---

TODO:
- [x] synchronise access to `SectorStateMgr`
- [x] add publisher/subscriber for SectorStateMgr and Index Provider
- [ ] get LID to subscribe to updates to SectorStateMgr too and do a cleanup when unsealed copy is removed
- [ ] figure out why `TestMultipleDealsConcurrent` fails on CI but passes consistently locally


piecedirectory/doctor.go
```diff
@@ -75,7 +109,8 @@ func (d *Doctor) Run(ctx context.Context) {
 		// Sleep for a few seconds between ticks.
 		// The time to sleep is randomized, so that if there are multiple doctor
 		// processes they will each process some pieces some of the time.
-		sleepTime := avgCheckInterval/2 + time.Duration(rand.Intn(int(avgCheckInterval)))*time.Millisecond
```
nonsense, line 78: This is a bug.

piecedirectory/doctor.go
```diff
@@ -119,7 +163,7 @@ func (d *Doctor) checkPiece(ctx context.Context, pieceCid cid.Cid) error {
 		}
 	}
 
-	if !hasUnsealedDeal {
+	if !hasUnsealedDeal && !lacksActiveSector {
```
nonsense, line 166: Arguably this doesn't belong here, but the fact that we don't handle sector lifecycle properly, I'd rather we have this check somewhere than a ton of false-positives with respect to "unindexed" pieces.

piecedirectory/doctor.go
dirkmc: Nice catch 🎣 

I think we still want the randomization behaviour - it's for the case where there are multiple boost processes all running piece doctor. The bug is the multiplication by milliseconds.

dirkmc:
Thanks, I agree about the expired sectors.
We have an open ticket to clean up expired sector data from the piece directory:
https://github.com/filecoin-project/boost/issues/981#top

node/builder.go
```diff
@@ -555,7 +557,7 @@ func ConfigBoost(cfg *config.Boost) Option {
 		Override(new(*dagstore.DAGStore), func() *dagstore.DAGStore { return nil }),
 		Override(new(*mdagstore.Wrapper), func() *mdagstore.Wrapper { return nil }),
 
-		Override(new(pdtypes.Store), modules.NewPieceDirectoryStore(cfg)),
```
nonsense, line 558: I don't think we need the `pdtypes.Store` interface at all - we can just use the concrete type. As far as I can tell, the interface to `boostd-data/client` is not adding anything and is redundant.

piecedirectory/doctor.go
```diff
@@ -7,36 +7,37 @@ import (
...
+	bdclient "github.com/filecoin-project/boostd-data/client"
+	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
 )
 
 var doclog = logging.Logger("piecedoc")
 
-type SealingApi interface {
```
nonsense, line 18: `Piece Doctor` no longer accesses directly APIs from Lotus, and relies on information from the `Sector State Manager`.

indexprovider/wrapper.go
```diff
@@ -77,29 +85,180 @@ func NewWrapper(cfg *config.Boost) func(lc fx.Lifecycle, h host.Host, r repo.Loc
...
-	go w.usm.Run(runCtx)
-
 	// Announce all deals on startup in case of a config change
 	go func() {
 		err := w.AnnounceExtendedProviders(runCtx)
 		if err != nil {
 			log.Warnf("announcing extended providers: %w", err)
 		}
 	}()
+
```
nonsense, line 107: The changeset below is not modified and belonged to the `Unsealed State Manager` but it is really specific to the Index Provider. So I split it away from the Sector State Manager, and kept it under the Index Provider.

cmd/boostd/recover.go
```diff
@@ -190,14 +191,14 @@ func action(cctx *cli.Context) error {
 	if ignoreLID {
 		pd = nil
 	} else {
-		pdClient := piecedirectory.NewStore()
```
nonsense, line 193: This `store` is really the `boostd-data` service and NOT the piece directory.

indexprovider/wrapper_test.go
```diff
@@ -338,26 +326,15 @@ func TestUnsealedStateManagerStateChange(t *testing.T) {
...
-			// Set the first response from MinerAPI.StorageList()
-			resp1 := tc.storageListResponse1(abi.SectorID{Miner: abi.ActorID(minerID), Number: deals[0].SectorID})
+			// Set the current state from db -- response from MinerAPI.StorageList()
+			resp1 := tc.initialState(abi.SectorID{Miner: abi.ActorID(minerID), Number: deals[0].SectorID})
 			storageMiner.storageList = map[storiface.ID][]storiface.Decl{}
 			if resp1 != nil {
 				storageMiner.storageList["uuid"] = []storiface.Decl{*resp1}
 			}
 
-			// Trigger check for updates
```
nonsense, line 348: Looks like previously each test case was doing 2 updates, and not 1 update. At the moment I have it refactored to do 1 update per test, let's decide if we want another 10 tests for the other cases.

indexprovider/wrapper_test.go
```diff
@@ -368,89 +345,109 @@ func TestUnsealedStateManagerStorageList(t *testing.T) {
...
+					{
+						SectorID:       sectorID,
+						SectorFileType: storiface.FTUnsealed,
+					},
+					{
+						SectorID:       sectorID,
+						SectorFileType: storiface.FTSealed,
+					},
+				}
+			},
```
nonsense, line 372: These fixtures need more work, they are terrible right now. A lot if indirection for setting up relatively basic tests.

indexprovider/wrapper_test.go
```diff
@@ -22,62 +21,53 @@ import (
...
 	ctx := context.Background()
 
-	runTest := func(t *testing.T, usm *UnsealedStateManager, storageMiner *mockApiStorageMiner, prov *mock_provider.MockInterface, provAddr address.Address, sectorNum abi.SectorNumber) {
+	runTest := func(t *testing.T, wrapper *Wrapper, storageMiner *mockApiStorageMiner, prov *mock_provider.MockInterface, provAddr address.Address, sectorNum abi.SectorNumber) {
 		// Set the response from MinerAPI.StorageList() to be two unsealed sectors
 		minerID, err := address.IDFromAddress(provAddr)
 		require.NoError(t, err)
-		storageMiner.storageList = map[storiface.ID][]storiface.Decl{
-			// This sector matches the deal in the database
-			"uuid-existing-deal": {{
```
nonsense, line 44: Not sure why this is referring to deal uuid, as this is a sector id.

storagemarket/deal_execution.go
```diff
@@ -83,6 +83,8 @@ func (p *Provider) execDeal(deal *smtypes.ProviderDealState, dh *dealHandler) (d
 	// Capture any panic as a manually retryable error
 	defer func() {
 		if err := recover(); err != nil {
+			// TODO: it'd be nice to expose and log these errors in tests
+			fmt.Println("panic: ", err, string(debug.Stack()))
```
dirkmc, line 87: Let's remove this line

storagemarket/provider_test.go
```diff
@@ -138,7 +138,9 @@ func TestSimpleDealHappy(t *testing.T) {
 	})
 }
 
-func TestMultipleDealsConcurrent(t *testing.T) {
+func XTestMultipleDealsConcurrent(t *testing.T) {
```
dirkmc, line 141: Let's reenable the test

sectorstatemgr/pubsub.go
```diff
@@ -0,0 +1,60 @@
+package sectorstatemgr
+
+import "sync"
+
+type PubSub struct {
```
dirkmc, line 5: Is it possible to use an existing pubsub library rather than creating a new one?
In other places in the code we use "github.com/hannahhoward/go-pubsub" and EventEmitter from "github.com/libp2p/go-libp2p/core/event"

piecedirectory/doctor.go
```diff
@@ -5,38 +5,65 @@ import (
...
+				if !ok {
+					log.Debugw("state updates subscription closed")
+					return
+				}
+				log.Debugw("got state updates from SectorStateMgr", "len(u.updates)", len(u.Updates), "len(u.active)", len(u.ActiveSectors), "u.updatedAt", u.UpdatedAt)
+
+				d.latestUpdateMu.Lock()
+				d.latestUpdate = u
+				d.latestUpdateMu.Unlock()
+
```
dirkmc, line 60: Does it make sense to trigger a check in response to getting a new update from SectorStateMgr or are they on very different schedules?

sectorstatemgr/pubsub.go
nonsense: I don't see references to `EventEmitter` in Boost. `github.com/hannahhoward/go-pubsub` is certainly going to introduce more complexity and indirection if we decide to use it.

storagemarket/deal_execution.go
nonsense: Why? At the moment if a panic happens in tests it is not visible, because we catch it and recover from it, which is super confusing.

nonsense: I am removing the TODO as the println is pretty much doing what it says.

nonsense: A panic happening here is really a problem with Boost that we should fix, and generally should never happen. I agree it is probably a good idea to catch and recover from panics if we somehow introduce a bug that a single deal manages to trigger, so that we don't bring `boostd` down.

dirkmc (approved):
Nice work! 🙌

PR #1471, opened 2023-05-26
feat: display msg params in the mpool UI

LexLuthr:
- [x] Test
Possible nil pointer error from incorrect type

Screenshot 2023-05-30 at 2 31 07 PM


dirkmc (approved):
Nice work 🙌

Issue #1472, opened 2023-05-29
boostd import-data should allow delete after import with proposal cid param


dirkmc:
The user can import data for an offline deal with `boostd import-data <identifier> <file>`. The first parameter to the command is either a deal proposal cid or a uuid.

If the parameter is a deal proposal cid, and the `delete-after-import` flag is set, the command outputs `legacy deal data cannot be automatically deleted after import (only new deals)`.

This message should only be output for legacy deals, but it is currently also output for new deals (deal proposal protocol `v1.2.x`)

PR #1473, opened 2023-05-29
fix: incorrect check when import offline deal data using proposal CID

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1472

### DIFF
--- a/cmd/boostd/import_data.go
+++ b/cmd/boostd/import_data.go
@@ -67,9 +67,6 @@ var importDataCmd = &cli.Command{
 		// If the user has supplied a signed proposal cid
 		deleteAfterImport := cctx.Bool("delete-after-import")
 		if proposalCid != nil {
-			if deleteAfterImport {
-				return fmt.Errorf("legacy deal data cannot be automatically deleted after import (only new deals)")
-			}
 
 			// Look up the deal in the boost database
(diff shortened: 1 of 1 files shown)


dirkmc (approved):
Please test on a local devnet before merging

cmd/boostd/import_data.go
```diff
@@ -80,6 +77,10 @@ var importDataCmd = &cli.Command{
 					return err
 				}
 
+				if deleteAfterImport {
+					return fmt.Errorf("legacy deal data cannot be automatically deleted after import (only new deals)")
```
dirkmc, line 81: We could get to this point in the code if the user entered a proposal cid that doesn't exist. So maybe this error message should be
```
fmt.Errorf("cannot find boost deal with proposal cid %s and legacy deal data cannot be automatically deleted after import (only new deals)", proposalCid)
```

LexLuthr:
On devnet:

```
root@a52a4463538c:/app/sample# boostd import-data --delete-after-import bafyreidz2b4or74w2dhvghslhefdcwojwrqdlm5wnf42mdpo2nypfvnufi ../public/sample.car 
Offline deal import for v1.2.0 deal 876d3517-befe-41cc-9385-eb4237b7fd4c scheduled for execution
root@a52a4463538c:/app/sample# 
```

Screenshot 2023-05-29 at 1 43 23 PM

Issue #1453, opened 2023-05-17
Documentation for Boost v2 (LID information)


nonsense:
1. - [x] Document LID architecture- https://boost.filecoin.io/experimental-features/local-index-directory
2. - [x] Document LID requirements - https://boost.filecoin.io/experimental-features/local-index-directory/requirements
3. - [ ] Document LID migration - https://boost.filecoin.io/experimental-features/local-index-directory/initialisation
        - [x] Complete sections about migration
        - [ ] Guidance (if we want to provide any) on deals to potentially skip
4. - [x] Add variants to Documentation
        - [x] We want 2 variants for http://boost.filecoin.io/ – one for `stable` release, and one for `unreleased / master` version. Coordinate with Mayank and fix DNS / Gitbook for it.
5. - [ ] Changes to `unrelease` variant of Documentation
        - [x] Tutorial for upgrading from Boost v1 to Boost v2
        - [x] Explanation of `boostd-data` and how it interacts with YugabyteDB and LevelDB
        - [x] Maybe add `Deployment` page that explains a full deployment and all processes of Boost (boostd, boostd-data, http, bitswap, lotus, lotus-miner, etc.) (currently this is Getting Started -> As a SP)
        - [x] Add more docs on YugabyteDB – not a full copy-paste of YugabyteDB docs, but some explanation how to get started with it.
        - [x] Downgrade Lotus -> Boost upgrade into a not-so-central page
        - [x] Add documentation page on Storage- and Retrieval- in general why it is important for Retrieval- to be up and running (i.e. incentives, FIL+ data, Spade reputation, etc.)
        - [x] Document that the connection to the LID service must be a websocket connection (not http)
        - [ ] Document what “unsealed”, “sealed”, “indexing” means

LexLuthr:
Please add the following as well:

- [ ] Boost component overview - All binaries and what they do along with how they connect to each other
- [ ] How Boost talks to lotus - boundary between Boost, Miner and daemon
- [ ] All exposed ports and what they are used for - all security considerations
- [ ] Getting familiar with Boost UI - What each page contains and how to look up things
- [ ] How retrievals work - all 3 protocols showing where data comes from

brendalee:
An additional item to consider (mostly for LID migration documentation) - after talking to Rob, could start compiling a list of FAQs on migration tooling. 

For example: 
- Do I need to run the migration in one go, or do I have an option to do it incrementally?
- Can I still make storage deals while doing the migration to LID?

[...]

PR #1476, opened 2023-05-30
fix: set UI default listen address to localhost

LexLuthr:
We should avoid setting it to 0.0.0.0 by default as SPs might end up exposing their UI to the internet if firewall is not configured properly.
I have just come across one such instance. Better safe than sorry.
We also need to update our V2 docs with much more detail about networking and security.

### DIFF
--- a/node/config/def.go
+++ b/node/config/def.go
@@ -65,7 +65,7 @@ func DefaultBoost() *Boost {
 		},
 
 		Graphql: GraphqlConfig{
-			ListenAddress: "0.0.0.0",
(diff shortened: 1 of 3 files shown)


dirkmc (approved):

dirkmc:
LGTM, thanks 👍

dirkmc:
> We also need to update our V2 docs with much more detail about networking and security.

Could you please add this to https://github.com/filecoin-project/boost/issues/1453

dirkmc:
Oh nevermind I see you did already 👍

Issue #1454, opened 2023-05-17
improve visibility of LID within Boost UI


nonsense:
## Information that should be surfaced

1. How much TiB/PiB of deal data an SP has
2. How many unsealed sectors we have
3. How many sealed sectors, without unsealed copies we have
4. Piece directory stats per sector? — how much was indexed… what sectors/deals are missing indices?

---

5. Split unsealed errors / sectors vs other issues
6. Display stats around sectors with no unsealed copies vs sectors with unsealed copies (% of number of sectors)
7. Flagged pieces — number of flagged pieces ; % of data that is flagged, vs total data held by miner
8. Fix lifecycle of deals — i.e. false positives of flagged pieces
9. Logs / progress around re-indexing
10. Block count & Average block size (Inspect Piece Page)

### The Inspect page today

![image](https://github.com/filecoin-project/boost/assets/169124/85553bdf-a633-477a-9027-91ebcb93e753)

### Proposal

Change the name of the "Inspect" menu item to "Local Index Directory"

This menu item links to a LID landing page.
The landing page links to a page with a list of sectors (like `lotus-miner sectors list) and a page with a list of flagged pieces (similar to what we have on the Inspect page today):

![image](https://github.com/filecoin-project/boost/assets/169124/04e7d58e-c0c5-46b0-9b7f-cd0942de69a4)

nonsense:
@LexLuthr yes, I think this is a good idea. We should bring in a bit more details from lotus-miner into Boost as you suggest.

nonsense:
In terms of implementation, I think we don't have to store any of this data in Boost, as it can change from underneath us (i.e. if a SP removes unsealed copies, or they get corrupted, or smth like that). We could have a periodic background worker that populates in-memory information about:
- number of sectors that contain deals, how many deals, and how much storage in total
- how many unsealed sectors with deals, vs sectors with deals that lack unsealed copies (try to get a CAR reader on each of them to make sure they aren't corrupted)
- number of index files, cross-referenced with the pieces in the sectors

That would already give intuition on the health of the unsealed copies + indices on a given SPs system, prior to initialising LID.

brendalee:
A few additional considerations (copying over from slack): 

To make it more clear to SPs where to get more info on if they need to unseal things, or index things, we can consider adding labels - such as data index state (for SPs to see the health of their data indexes), and sector state (to see the status of how many sealed/unsealed sectors they have, etc.). 

[...]

PR #1477, opened 2023-05-30
update local index directory ui

dirkmc:
Solution to https://github.com/filecoin-project/boost/issues/1454

Screenshot 2023-05-30 at 9 48 59 AM


gql/resolver_sectors.go
```diff
@@ -0,0 +1,77 @@
...
+		Sectors:    resolvers,
+		More:       snapshot.More,
+		At:         graphql.Time{Time: snapshot.At},
+	}, nil
+}
+
+// mutation: sectorsListRefresh()
+func (r *resolver) SectorsListRefresh(ctx context.Context) (bool, error) {
+	return true, r.sectorsList.Refresh(ctx)
+}
```
nonsense, line 77: Do we want to be able to trigger a refresh from the UI? It'd be simpler to just leave the SectorStateMgr update it, depending on its period.

react/src/App.js
```diff
@@ -53,13 +53,15 @@ function App(props) {
...
                                         <Route path="/mpool" element={<MpoolPage />} />
                                         <Route path="/settings" element={<SettingsPage />} />
-                                        <Route path="/piece-directory" element={<PieceDirectoryPage />} />
                                         <Route path="/deals/:dealID" element={<DealDetail />} />
                                         <Route path="/legacy-deals/:dealID" element={<LegacyDealDetail />} />
-                                        <Route path="/inspect" element={<InspectPage />} />
+                                        <Route path="/lid" element={<LIDPage />} />
+                                        <Route path="/sectors-list" element={<SectorsListPage />} />
+                                        <Route path="/sectors-list/from/:cursor/page/:pageNum" element={<SectorsListPage />} />
+                                        <Route path="/piece-doctor" element={<PieceDoctorPage />} />
```
nonsense, line 61: Both`/piece-doctor` and `/inspect` route to the PieceDoctor, I suggest we unify them somehow, as this is a bit confusing.

storagemarket/sealingpipeline/sectorslist.go
```diff
@@ -0,0 +1,240 @@
...
+	sort.Slice(apiList, func(i, j int) bool {
+		return apiList[i] < apiList[j]
+	})
+
+	sectorsList := make([]SectorState, 0, len(apiList))
+	for _, sectorNum := range apiList {
+		st, err := l.spApi.SectorsStatus(ctx, sectorNum, true)
+		if err != nil {
+			return err
+		}
```
nonsense, line 161: I don't think this will work for the large SPs we want to support -- currently we have SPs with 30-50PiB of data, and we want to go up to 100PiB... this means we should think how we use these APIs that would return 1.5M up to 2M records.

I don't think we should synchronously make 2M API calls.

davidd8:
A few thoughts on the current design:
1. Change the "Repair" button to have a clearer CTA, i.e. "View Pieces to Repair", otherwise a user may assume a "bulk repair" action.
2. What is the purpose of the "Sector Unsealed Copies" graph? It seems SPs will have unique distributions of the percentage of deals/sectors storing unsealed copies, so it's not clear what the graph is for, i.e. should it all be blue? Should there be a flag for sectors that "should" be unsealed, or this just nice-to-have information?
3. [nit] Make the "flagged" color more red, rather than a light blue to encourage repair.
4. "Sector Proving State" graph seems out of place, wouldn't this be useful information outside of LID?

dirkmc:
> Change the "Repair" button to have a clearer CTA, i.e. "View Pieces to Repair", otherwise a user may assume a "bulk repair" action.

Changed to "View Flagged Pieces" so as to match the language "flagged pieces" in other parts of the page

> What is the purpose of the "Sector Unsealed Copies" graph? It seems SPs will have unique distributions of the percentage of deals/sectors storing unsealed copies, so it's not clear what the graph is for, i.e. should it all be blue? Should there be a flag for sectors that "should" be unsealed, or is this just nice-to-have information?

SPs often ask why retrieval isn't working, and the answer is frequently that there is no unsealed copy of the piece that the client wants to retrieve. We're hoping the "Sector Unsealed Copies" chart will help surface this problem.

> [nit] Make the "flagged" color more red, rather than a light blue to encourage repair.

Agreed 👍 
I'll do this in a follow-up PR.

> "Sector Proving State" graph seems out of place, wouldn't this be useful information outside of LID?

I think we will want to surface this to the main boost landing page when we redesign the UI. I'd suggest we keep it where it is for now until we take on the larger redesign effort.

PR #1487, opened 2023-06-02
LID UI and backend

nonsense:


sectorstatemgr/sectorstatemgr.go
```diff
@@ -214,13 +231,99 @@ func (m *SectorStateMgr) refreshState(ctx context.Context) (*SectorStateUpdates,
...
+	})
+
+	sectorsTable := make([]Sector, 0, len(sectorNumbers))
+	for _, sectorNumber := range sectorNumbers {
+		sid := abi.SectorID{
+			Miner:  abi.ActorID(mid),
+			Number: sectorNumber,
+		}
+
+		//st, err := m.minerApi.SectorsStatus(ctx, sectorNumber, true)
```
nonsense, line 272: @dirkmc I think we should limit ourselves to calling only APIs for all sectors that we have high confidence will respond within up to 30sec.

If we want to individually call `SectorsStatus` for all sectors, it should happen in a background worker/subsystem, and we need to cache the state and introduce concurrency.

`2M sectors * 10ms per response / 8 concurrency factor` is still 40min.

PR #1492, opened 2023-06-05
update boxo

dirkmc:
TODO:
- [x] Update to release version of go-fil-markets
- [x] Update to release version of boost-gfm
- [x] Update to lotus commit


itests/framework/framework.go
```diff
@@ -715,7 +716,12 @@ func (f *TestFramework) ExtractFileFromCAR(ctx context.Context, t *testing.T, fi
 		require.NoError(t, err)
 	}
 
-	nd, err := ipld.Decode(b)
+	reg := ipld.Registry{}
+	reg.Register(cid.DagProtobuf, dag.DecodeProtobufBlock)
+	reg.Register(cid.DagCBOR, ipldcbor.DecodeBlock)
+	reg.Register(cid.Raw, dag.DecodeRawBlock)
+
+	nd, err := reg.Decode(b)
```
aschmahmann, line 724: These are by far the most common codecs registered in the go-ipld-format global map which is being removed. Given that you're only doing UnixFS here we could drop the dag-cbor code too and things should be fine.

itests/framework/framework.go
```diff
@@ -128,7 +129,7 @@ func FullNodeAndMiner(t *testing.T) (*kit.TestFullNode, *kit.TestMiner) {
 	fnOpts := []kit.NodeOpt{
 		kit.ConstructorOpts(
 			lnode.Override(new(lp2p.RawHost), func() (host.Host, error) {
-				return libp2p.New()
+				return libp2p.New(libp2p.DefaultTransports)
```
aschmahmann, line 132: This is a hack to deal with the following issue:
- You cannot start a listener on the same address multiple times (e.g. starting two quic-v1 listeners on `/ip4/1.2.3.4/udp/1234` will [panic](https://github.com/libp2p/go-libp2p/blob/686a32aae741aeb0ebe825553ec066d669a744ae/p2p/transport/quic/virtuallistener.go#L60))
- `libp2p.New()` listens on all interfaces by default
- There's a lotus hook that starts listening https://github.com/filecoin-project/lotus/blob/9e4f1a4d23ad72ab191754d4f432e4dc754fce1b/node/modules/lp2p/addrs.go#L92

We're getting around this by passing the same default transports we would have but no addresses which is making sure the libp2p host constructor doesn't do any listening, only the lotus code that explicitly listens does the listening. Setting any transports or listen addresses would do here though.

Note: The code in lotus that calls the listener was removed in kubo here https://github.com/ipfs/kubo/pull/9857#discussion_r1186109527 (and I tagged the lotus go-libp2p PR). Probably that code can be removed as well given the same caveats as the kubo removal.

This should be fine for this PR, but it seems like probably this should get cleaned up. Maybe just porting the fix from kubo to lotus is sufficient, but perhaps changes in other modules (e.g. go-libp2p, boost, or the lotus DI setup used for testing) would also be helpful.

Heads up @marten-seemann @marcopolo @magik6k

hannahhoward:
Retrieving from booster-http with Lassie, piping, and extracting a file at a path.

https://github.com/filecoin-project/boost/assets/1450680/aea23956-0064-4dc6-a407-a897f4e3a67b

PR #1494, opened 2023-06-06
GraphQL resolvers for LID

nonsense:


gql/resolver_lid.go
```diff
@@ -0,0 +1,63 @@
...
+type lidState struct {
+	DealData             dealData
+	Pieces               pieces
+	SectorUnsealedCopies sectorUnsealedCopies
+	SectorProvingState   sectorProvingState
+	FlaggedPieces        int32
+}
+
+// query: lid: [LID]
+func (r *resolver) LID(ctx context.Context) (*lidState, error) {
```
dirkmc, line 40: I'd suggest creating a different resolver method for each of these subsections (one for DealData, one for Pieces etc).

That way each one is loaded separately, and if one of them fails or is slow it doesn't affect the others.

nonsense: These should not fail, as all the data is coming from Boost and should be cached by that point - I'd rather we make 1 API call for a few int.

PR #1495, opened 2023-06-06
Parse indexed deals for data segment index

willscott:
this makes boost aware of the format in
https://github.com/filecoin-project/go-data-segment/ for new deals.

this is the same logic as in https://github.com/filecoin-project/lotus/pull/10674


go.mod
```diff
@@ -335,10 +335,11 @@ require (
 require (
 	github.com/filecoin-project/boost-gfm v1.26.7
 	github.com/filecoin-project/boost-graphsync v0.13.6
+	github.com/filecoin-project/go-data-segment v0.0.0-20230605095649-5d01fdd3e4a1
```
dirkmc, line 338: Can we point at a release version of go-data-segment?

piecedirectory/piecedirectory.go
```diff
@@ -222,17 +248,44 @@ func (ps *PieceDirectory) addIndexForPiece(ctx context.Context, pieceCid cid.Cid
...
 	}
+	return recs, nil
+}
 
-	// Add mh => piece index to store: "which piece contains the multihash?"
-	// Add mh => offset index to store: "what is the offset of the multihash within the piece?"
-	log.Debugw("add index: store index in local index directory", "pieceCid", pieceCid)
-	if err := ps.store.AddIndex(ctx, pieceCid, recs, true); err != nil {
-		return fmt.Errorf("adding CAR index for piece %s: %w", pieceCid, err)
+func parseShardWithDataSegmentIndex(ctx context.Context, pieceCid cid.Cid, unpaddedSize int64, r types.SectionReader) ([]model.Record, error) {
```
dirkmc, line 256: nit: I suggest we use the word "piece" instead of shard. Shard only has meaning in the dagstore context

piecedirectory/piecedirectory.go
```diff
@@ -222,17 +248,44 @@ func (ps *PieceDirectory) addIndexForPiece(ctx context.Context, pieceCid cid.Cid
...
+	recs := make([]model.Record, 0)
+	for _, s := range segments {
+		segOffset := s.UnpaddedOffest()
+		segSize := s.UnpaddedLength()
+
+		lr := io.NewSectionReader(r, int64(segOffset), int64(segSize))
+		subRecs, err := parseRecordsFromCar(lr)
+		if err != nil {
+			log.Debugw("Unexpected index format on generation in shard", "piece", pieceCid, "offset", segOffset)
+			continue
```
dirkmc, line 283: Shouldn't we return an error if the piece data is corrupted?

piecedirectory/piecedirectory_test.go
```diff
@@ -248,6 +253,61 @@ func testImportedIndex(ctx context.Context, t *testing.T, cl *client.Store) {
...
+			OffsetSize: model.OffsetSize{
+				Offset: r.Offset,
+				Size:   0,
+			},
+		}
+	}
+
+	// Add the index to the local index directory, marked as incomplete
+	err = cl.AddIndex(ctx, pieceCid, recs, false)
+	require.NoError(t, err)
```
dirkmc, line 299: I don't think we need to test importing from the dagstore here. Just the data segment index functionality.

I'd suggest we also skip lines 266 - 269. Parsing the piece data and adding the index should all happen as part of AddDealForPiece.

go.mod
willscott: this is the release version i think:
https://github.com/filecoin-project/go-data-segment/releases/tag/0.0.1

because the tag is '0.0.1' and not 'v0.0.1' go.mod uses the commit hash like this to describe it.

cc @Kubuxu

piecedirectory/piecedirectory.go
willscott: added a comment explaining why

go.mod
Kubuxu: Hrm, I will fix it.

Kubuxu: Released: https://github.com/filecoin-project/go-data-segment/releases/tag/v0.0.1

LaurenSpiegel:
@CodiumAI-Agent

LexLuthr:
Quick checklist for this PR.
- [ ] Add an itest with a non-car file for deal flow and HTTP retrieval - @willscott
- [ ] Rebase this branch on https://github.com/filecoin-project/boost/tree/feat/noncar-files

willscott:
@rvagg  it looks like there's some nit with car offset propagation that didn't make it through the merge here properly - graphsync / retrievals look like they're not always getting expected readers

rvagg:
yeah, a bunch missing.

2 commits pushed, first one (e3b5841) fixes all that stuff; the second one (5932ce7) just adds some fixes to your data_segment_index_retrieval_test.go to make it pass — I know it's not what you're aiming for, hence the second commit you can unwind if you want, but aside from changing CIDs to make it pass, there's a couple of proper fixes in there, notably the `false` argument in `RetrieveDirect` telling it to not unpack the CAR UnixFS files, you just want to look at the CAR, and the second retrieval root check that was comparing against root 1.

PR #1502, opened 2023-06-08
check that sector has deals for unsealed sectors

nonsense:


gql/resolver_lid.go
```diff
@@ -56,9 +56,11 @@ func (r *resolver) LID(ctx context.Context) (*lidState, error) {
 
 	var sealed, unsealed int32
 	for id, s := range lu.SectorStates { // TODO: consider adding this data directly in SSM
+		_, sectorHasDeals := lu.SectorWithDeals[id]
+
 		if s == db.SealStateUnsealed {
 			unsealed++
-		} else if s == db.SealStateSealed {
+		} else if s == db.SealStateSealed && sectorHasDeals == true {
```
nonsense, line 63: At the moment we are including `CC` sectors in the `unsealed` count, which is wrong. We care only about sectors that has only sealed copy and contain deals.

dirkmc (approved):
Looks like there's a linting error but otherwise LGTM 👍

PR #1503, opened 2023-06-08
piece doctor to ignore expired/slashed deals

nonsense:
The algorithm in `checkPiece` currently flags piececids that the SP never stored -- they submitted a PSD, but the deal was never added to a sector. Even worse, another deal later on got added to that particular sectorID, which means that based on the metadata in LID, we would "think" that the deal's sector is `active`.

Therefore I am changing the algorithm to:
1. Check if piece belongs to an active sector
2. For the active sectors / deals from 1, check that at least one deal is actually found on-chain
3. At this stage we know there is an active sector and a real ChainDealID; check if the piececid is indexed
4. If the piece is indexed and has active sector and active deal on chain, check for unsealed copy

We `Flag` a piece on step 3 and 4 if a piece is not indexed or if there is no unsealed copy.
Otherwise we `Unflag` a piece.

### DIFF
--- a/piecedirectory/doctor.go
+++ b/piecedirectory/doctor.go
@@ -13,6 +13,7 @@ import (
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
 )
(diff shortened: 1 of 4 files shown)


piecedirectory/doctor.go
```diff
@@ -98,12 +100,77 @@ func (d *Doctor) Run(ctx context.Context) {
...
+	// Check that Deal is actually on-chain for the active sectors
+	/////////////////////////////////////////////////////////////////
+	head, err := d.fullnodeApi.ChainHead(ctx)
+	if err != nil {
+		return err
+	}
+
+	found := false
+	for _, dealId := range chainDealIds {
+		_, err := d.fullnodeApi.StateMarketStorageDeal(ctx, dealId, head.Key())
```
dirkmc, line 154: We might also want to consider passing the StateMarketStorageDeal function as a parameter to checkPiece, so that it's easier to test (without having to use mocks) - take a look at testCheckPieces in doctor_test.go

PR #1508, opened 2023-06-09
LID landing page: add stats around Flagged and non-Flagged pieces

nonsense:


dirkmc (approved):

extern/boostd-data/ldb/db.go
```diff
@@ -448,6 +448,32 @@ func (db *DB) NextPiecesToCheck(ctx context.Context) ([]cid.Cid, error) {
 	return pieceCids, nil
 }
 
+func (db *DB) PiecesCount(ctx context.Context) (int, error) {
```
nonsense, line 451: This is crazy inefficient... I guess we might want to optimise at some point and store the count in a key/value pair.

dirkmc: ha yeah we should - hopefully people won't use leveldb for large datasets anyway

PR #1512, opened 2023-06-12
feat: mpool monitor

LexLuthr:
- [ ] Test on filcolins
- [ ] Test on devnet

### DIFF
--- a/gql/mpool_monitor.go
+++ b/gql/mpool_monitor.go
@@ -1,0 +1,229 @@
+package gql
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"sync"
+	"time"
+
+	gqltypes "github.com/filecoin-project/boost/gql/types"
+	"github.com/filecoin-project/boost/node/config"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/lotus/api/v1api"
+	"github.com/filecoin-project/lotus/chain/consensus"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/ipfs/go-cid"
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+type msg struct {
+	To         string
+	From       string
+	Nonce      gqltypes.Uint64
+	Value      gqltypes.BigInt
+	GasFeeCap  gqltypes.BigInt
+	GasLimit   gqltypes.Uint64
+	GasPremium gqltypes.BigInt
+	Method     string
+	Params     string
+	BaseFee    gqltypes.BigInt
+}
+
+// timeStampedMsg wraps the pending msg with a chainEpoch
+type timeStampedMsg struct {
+	m     msg
+	added abi.ChainEpoch // Epoch when message was first noticed in mpool
+}
+
+type mpoolMonitor struct {
+	ctx                context.Context
+	cancel             context.CancelFunc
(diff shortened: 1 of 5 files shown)


lib/mpoolMonitor/mpool_monitor.go
```diff
@@ -0,0 +1,180 @@
...
+		msgs:               make(map[cid.Cid]*timeStampedMsg),
+	}
+}
+
+func (mm *MpoolMonitor) Start(ctx context.Context) error {
+	log.Infow("Mpool monitor: starting")
+	mmctx, cancel := context.WithCancel(ctx)
+	mm.ctx = mmctx
+	mm.cancel = cancel
+	err := mm.update(ctx)
```
dirkmc, line 47: I'd suggest we move mm.update inside of startMonitoring so it doesn't block boost from starting up other components (dependency injection calls the Start method of each component in series)

lib/mpoolMonitor/mpool_monitor.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to start mpool monitor: %w", err)
+	}
+	go mm.startMonitoring(mmctx)
+	return nil
+}
+
+func (mm *MpoolMonitor) startMonitoring(ctx context.Context) {
+	ticker := time.NewTicker(10 * time.Second)
```
dirkmc, line 56: I'd suggest running update once per epoch

```suggestion
	ticker := time.NewTicker(10 * time.Second)
```

node/config/def.go
```diff
@@ -65,8 +65,9 @@ func DefaultBoost() *Boost {
 		},
 
 		Graphql: GraphqlConfig{
-			ListenAddress: "127.0.0.1",
-			Port:          8080,
+			ListenAddress:      "127.0.0.1",
+			Port:               8080,
+			PendingAlertEpochs: 10,
```
dirkmc, line 70: I'd suggest
- we move the config somewhere else (it's not directly to do with graphql)
  - maybe at the top level? not sure where exactly would make sense
- we change the name to `MpoolAlertEpochs`
- we start with a value of 30 (I believe lotus tries to estimate gas fees such that messages will get sent within 20 epochs)

lib/mpoolMonitor/mpool_monitor.go
```diff
@@ -0,0 +1,180 @@
+package mpoolMonitor
```
dirkmc, line 1: nit: usually directory names and package names are all lowercase (eg remoteblockstore)

dirkmc (commented):
Looks great 👍

node/config/def.go
```diff
@@ -58,16 +58,17 @@ func DefaultBoost() *Boost {
 	cfg := &Boost{
 		Common: defCommon(),
 
+		MpoolAlertEpochs: 30,
```
dirkmc, line 61: Because all the other config fields are under a section, we should probably move this one under a section too.
I suggest we call the section "Monitoring"

dirkmc (approved):
One suggestion to move the config param but otherwise LGTM 👍 
Nice work 🙌

node/config/def.go
LexLuthr: Added the new section.

```
./boostd config default --no-comment | grep -A1 Monitoring
[Monitoring]
  MpoolAlertEpochs = 30
```

PR #1515, opened 2023-06-12
fix unsealed field in flagged piece

nonsense:
This PR is fixing the issue where we flag every piece as if it has no unsealed copy in the FlaggedPiece table:

```
err = d.store.FlagPiece(ctx, pieceCid, false)
```


dirkmc:
LGTM 👍

PR #1521, opened 2023-06-13
Add flag to allow restarting devnet

lalexgap:
This PR modifies the devnet script so you can restart it without removing all your existing state directories and config files. Being able to restart the devnet means we can modify config files in our local ~/.lotus directory and restart the devnet to pick up those changes. It also means you don't have to run through all on-chain setup if you accidentally kill your running devnet.


nonsense (approved):

nonsense:
Thank you @lalexgap , this LGTM. Generally we use the `docker` setup nowadays, so check it out in case it is helpful for your usecase as well.

PR #1523, opened 2023-06-14
feat: add send epoch, time, elapsed epoch and elapsed time for each message in mpool to UI

LexLuthr:
- [ ] Test in devnet

Screenshot 2023-06-14 at 4 44 14 PM

### DIFF
--- a/lib/mpoolmonitor/mpoolmonitor.go
+++ b/lib/mpoolmonitor/mpoolmonitor.go
@@ -16,10 +16,11 @@ import (
 
 var log = logging.Logger("mpoolmonitor")
 
-// timeStampedMsg wraps the pending msg with a chainEpoch
-type timeStampedMsg struct {
-	m     types.SignedMessage
-	added abi.ChainEpoch // Epoch when message was first noticed in mpool
+// TimeStampedMsg wraps the pending msg with a chainEpoch
+type TimeStampedMsg struct {
+	SignedMessage types.SignedMessage
+	Added         abi.ChainEpoch // Epoch when message was first noticed in mpool
+	AddedTime     time.Time
 }
 
 type MpoolMonitor struct {
@@ -28,14 +29,14 @@ type MpoolMonitor struct {
 	fullNode         v1api.FullNode
 	lk               sync.Mutex
 	mpoolAlertEpochs abi.ChainEpoch
-	msgs             map[cid.Cid]*timeStampedMsg
+	msgs             map[cid.Cid]*TimeStampedMsg
 }
 
 func NewMonitor(fullNode v1api.FullNode, mpoolAlertEpochs int64) *MpoolMonitor {
 	return &MpoolMonitor{
 		fullNode:         fullNode,
 		mpoolAlertEpochs: abi.ChainEpoch(mpoolAlertEpochs),
-		msgs:             make(map[cid.Cid]*timeStampedMsg),
+		msgs:             make(map[cid.Cid]*TimeStampedMsg),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


lib/mpoolmonitor/mpoolmonitor.go
```diff
@@ -84,12 +86,13 @@ func (mm *MpoolMonitor) update(ctx context.Context) error {
...
+	newMsgs := make(map[cid.Cid]*TimeStampedMsg)
 
 	for _, pm := range msgs {
-		newMsgs[pm.Cid()] = &timeStampedMsg{
-			m:     *pm,
-			added: ts.Height(),
+		newMsgs[pm.Cid()] = &TimeStampedMsg{
+			SignedMessage: *pm,
+			Added:         ts.Height(),
+			AddedTime:     t,
```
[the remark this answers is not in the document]
LexLuthr: Calculating time from epoch will not work as we don't know when the first epoch is set. We need to take into account that network start epoch will be different for mainnet, calibnet etc. 
Let me know if I missed something here.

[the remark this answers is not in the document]
dirkmc: Fair enough - in that case I'd say let's not keep track of the wall clock time at all - just display relative times on the front end (I suggested a relative time format [in another comment](https://github.com/filecoin-project/boost/pull/1523#discussion_r1229548682))

react/src/Mpool.js
```diff
@@ -44,32 +46,28 @@ function MpoolMessage(props) {
...
     }
 
-    let t = msg.ElapsedSeconds+''
     let hh = Math.floor(t / 3600)
     t %= 3600
     let mm = Math.floor(t / 60)
-    let etime = `${hh} hours and ${mm} minutes`
+    t %= 60
+    let etime = `${hh} hours and ${mm} minutes ${t} seconds`
+    let epochInfo = `${sentEpoch} (${currEpoch - sentEpoch} epochs / ${etime} ago)`
```
dirkmc: You can use the moment library to do most of this work:
```
let elapsed = data.epoch.Epoch - msg.SentEpoch
moment().add(-elapsed*30, 'seconds').fromNow()
```

LexLuthr: I have updated the code to use moment library. I was thinking about `Show All Messages` and it does not seems to be adding any value. It is not helpful for debugging either. Plus, it is slow to load.
I would vote to remove the section all together and use that section to show alert messages. We call the button `show alerts` and render full messages there. In a full size SP, a lot of messages are pending at a time and this would be helpful. Or get rid of the button completely.

nonsense: Agree, we don't care about `Show All Messages` and we should remove it.

react/src/Mpool.js
```diff
@@ -13,7 +15,7 @@ export function MpoolPage(props) {
 
 function MpoolContent(props) {
     const [local, setLocal] = useState(true)
-    const {loading, error, data} = useQuery(MpoolQuery, { variables: { local } })
+    const {loading, error, data} = useQuery(MpoolQuery, { variables: { local } , pollInterval: 4000, fetchPolicy: "network-only"})
```
nonsense, line 18: Why pool every 4 seconds? This should be at least 15sec. given that block time is 30sec.

nonsense (approved):

react/src/Mpool.js
LexLuthr: Actually, this accounts for devnet as well. Ideally, I wanted to use build.BlockDelaySecs but that comes from MpoolQuery so I went with shortest available number. Even with 4 seconds, it should not be too aggressive as we are not polling directly to lotus node.

nonsense: Why do we care about this feature at all on a devnet? There should be no alerts or backlogged messages on a devnet. I'd rather we don't spam Lotus with unnecessary API calls -- all the pages start to add up at some point.

LexLuthr: I am pushing one more commit to remove all messages page and everything related to that. I will also update the polling time.

This page does not make a call to lotus API directly. It uses cached entries in mpoolManager with Boost.

LexLuthr: @nonsense I have adjusted the polling time and moving the other changes (Removing all messages) to a new PR.

nonsense (approved):

Issue #1526, opened 2023-06-15
Remove mpool show all messages


LexLuthr:
We should replace the button with a new one that shows all local messages. The default behaviour of mpool page should be changes to show alert messages.

PR #1530, opened 2023-06-19
feat: refactor mpool page in UI

LexLuthr:
This PR fixes https://github.com/filecoin-project/boost/issues/1526  and introduces the following changes to the Mpool page in UI.

1. It removed the all messages
2. Only local messages are captured by mpoolMonitor
3. Mpool resolver now works for alert messages or all messages
4. By default mpool UI will show all alert messages in details. This would allow for better debugging than just saying how many messages are stuck.

Screenshot 2023-06-19 at 7 51 47 PM

5. The button had been updated to show all local messages. User will have less use of local messages as well if there are not alerts. Thus the decision to make alerts primary and all messages a secondary query. 

Screenshot 2023-06-19 at 7 51 51 PM

### DIFF
--- a/lib/mpoolmonitor/mpoolmonitor.go
+++ b/lib/mpoolmonitor/mpoolmonitor.go
@@ -3,6 +3,7 @@ package mpoolmonitor
 import (
 	"context"
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
@@ -20,7 +21,6 @@ var log = logging.Logger("mpoolmonitor")
 type TimeStampedMsg struct {
 	SignedMessage types.SignedMessage
 	Added         abi.ChainEpoch // Epoch when message was first noticed in mpool
-	AddedTime     time.Time
 }
 
 type MpoolMonitor struct {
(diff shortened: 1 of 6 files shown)


lib/mpoolmonitor/mpoolmonitor.go
```diff
@@ -76,7 +77,17 @@ func (mm *MpoolMonitor) startMonitoring(ctx context.Context) {
...
+	localAddr := make(map[string]struct{})
+
+	addrs, err := mm.fullNode.WalletList(ctx)
+	if err != nil {
+		return fmt.Errorf("getting local addresses: %w", err)
+	}
+
+	for _, a := range addrs {
+		localAddr[a.String()] = struct{}{}
+	}
```
nonsense, line 89: I suggest we run this once on startup and that's it. No reason to call this API 3 times every epoch.

LexLuthr: Wallets are dynamic in Lotus and updating local wallets (dedicated wallets for miner actions) is not new. I would argue to reduce the frequency of this check but running it once on start up might not be the best solution. This would force folks to restart Boost nodes if they make any changes to daemon or miner in terms of wallets.

LexLuthr: Resolving based on https://filecoinproject.slack.com/archives/C037UKADRB2/p1688731687211039

dirkmc (approved):

PR #1539, opened 2023-06-27
Build indexes for legacy deals

dirkmc:
The list of flagged pieces is quite slow to load. This is because it scans through all the legacy deals to find matches against flagged pieces.

This PR adds an index for legacy deals by piece cid, root payload cid and publish cid. The index is refreshed every 10 minutes.

Test:
- [x] Manual test on sofiaminer


LexLuthr (approved):
LGTM!

Issue #1529, opened 2023-06-19
Clarify flagged piece reason


dirkmc:
Currently when it's not possible to access the data in an unsealed sector (eg because the data is corrupted), LID shows the sector's pieces in the list of flagged pieces with an unsealed copy:
<img width="600" alt="Screenshot 2023-06-19 at 3 02 02 PM" src="https://github.com/filecoin-project/boost/assets/169124/f7004edc-697c-4d11-8d03-b69c99849130">

However when the user clicks into the detail for the flagged piece, LID shows the piece Unsealed status as "No":
<img width="600" alt="Screenshot 2023-06-19 at 3 04 30 PM" src="https://github.com/filecoin-project/boost/assets/169124/7199dc3c-0b52-4ba7-8a11-c1c8b6b05c47">

Instead, boost should show the status of the piece as "Corrupted"

PR #1548, opened 2023-06-30
Check unseal status of piece through both apis

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/1529

Check the unsealed status through both StorageList and SectorsStatus APIs.
If they disagree, it likely means the piece is corrupted.

Screenshot 2023-06-30 at 12 56 10 PM


LexLuthr (approved):
Looks like we are covering all cases by giving priority to IsUnsealed. I think we are good to merge.

PR #1560, opened 2023-07-10
migrate config to v5

LexLuthr:
- [x] Test in devnet
- [x] Test on filcollins


LexLuthr:
Devnet:

```
cat /var/lib/boost/config.toml 
# The version of the config file (used for migrations)
#
# type: int
ConfigVersion = 5
```

filcollins

```
# The version of the config file (used for migrations)
#
# type: int
ConfigVersion = 5
```

dirkmc (approved):

PR #1562, opened 2023-07-11
IPNI UX

dirkmc:
Screenshot 2023-07-11 at 11 53 40 AM
Screenshot 2023-07-11 at 11 54 00 AM

### DIFF
--- a/react/src/Ipni.js
+++ b/react/src/Ipni.js
@@ -349,7 +349,7 @@ function IpniAdDetailFields({adCid}) {
             <td><IpniAdDetailEntryCount adCid={adCid} /></td>
         </tr>
         <tr>
-            <th>Previous Entry</th>
+            <th>Previous Advertisement</th>
             <td><Link to={'/ipni/ad/'+ad.PreviousEntry}>{ad.PreviousEntry}</Link></td>
         </tr>
         <tr>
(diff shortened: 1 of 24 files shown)


react/src/Ipni.js
```diff
@@ -0,0 +1,562 @@
...
+import './Ipni.css'
+import {Pagination} from "./Pagination";
+import {humanFileSize} from "./util";
+import {addClassFor} from "./util-ui";
+import bezier2Img from "./bootstrap-icons/icons/bezier2.svg";
+import {getDealStatus} from "./RetrievalLogDetail";
+import closeImg from "./bootstrap-icons/icons/x-circle.svg";
+
+const basePath = '/ipni'
+const indexerHost = 'cid.contact'
```
LexLuthr, line 25: It would be better we don't hardcode this value. A config variable would be best. For I think we can start with just one indexer with defaulting to cid.contact.

react/src/Ipni.js
```diff
@@ -0,0 +1,562 @@
...
+        <tr>
+            <th>Metadata</th>
+            <td><IpniAdMetadata metadata={ad.Metadata} /></td>
+        </tr>
+        <tr>
+            <th>Entries</th>
+            <td><IpniAdDetailEntryCount adCid={adCid} /></td>
+        </tr>
+        <tr>
+            <th>Previous Entry</th>
```
LexLuthr, line 352:
```suggestion
            <th>Previous Advertisement</th>
```

PR #1563, opened 2023-07-12
feat: add IPNI itest

LexLuthr:


itests/ipni_publish_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// Make a deal
+	res, err := f.MakeDummyDeal(dealUuid, carFilepath, rootCid, server.URL+"/"+filepath.Base(carFilepath), false)
+	require.NoError(t, err)
+	require.True(t, res.Result.Accepted)
+	log.Debugw("got response from MarketDummyDeal", "res", spew.Sdump(res))
+
+	time.Sleep(2 * time.Second)
+
+	// Wait for the first deal to be added to a sector and cleaned up so space is made
```
dirkmc, line 78:
```suggestion
	// Wait for the first deal to be added to a sector
```

dirkmc (commented):
Looks great, nice work 🙌 

I think you may need to add the new file to .circleci/config.yml in order for it to be picked up by the CI test runner (see test-itest-dummydeal)

itests/ipni_publish_test.go
LexLuthr: Yeah, I will add those now.

itests/ipni_publish_test.go
```diff
@@ -0,0 +1,109 @@
...
+	found := false
+
+	for _, mh := range mhs {
+		if mh.String() == rootCid.Hash().String() {
+			found = true
+			break
+		}
+	}
+
+	require.True(t, found)
```
masih, line 108: Use `require.Contains`?

itests/ipni_publish_test.go
```diff
@@ -0,0 +1,109 @@
...
+	require.NoError(t, err)
+	require.True(t, res.Result.Accepted)
+	log.Debugw("got response from MarketDummyDeal", "res", spew.Sdump(res))
+
+	time.Sleep(2 * time.Second)
+
+	// Wait for the deal to be added to a sector
+	err = f.WaitForDealAddedToSector(dealUuid)
+	require.NoError(t, err)
+	time.Sleep(100 * time.Millisecond)
```
masih, line 81: Use `require.Eventually`. Ditto re other `time.Sleep`s.

itests/ipni_publish_test.go
```diff
@@ -0,0 +1,109 @@
...
+	fileSize := 200000
+	randomFilepath, err := testutil.CreateRandomFile(tempdir, 5, fileSize)
+	require.NoError(t, err)
+
+	// create a dense carv2 for deal making
+	rootCid, carFilepath, err := testutil.CreateDenseCARv2(tempdir, randomFilepath)
+	require.NoError(t, err)
+
+	// Start a web server to serve the car files
+	log.Debug("starting webserver")
```
masih, line 61: Use `t.log`? Ditto in other places.

masih (commented):
I would say this is a Unit Test more than an integration test. Valuable nevertheless and thank you for adding this 👍

itests/ipni_publish_test.go
LexLuthr: Logging is separate for whole framework in itests

itests/ipni_publish_test.go
LexLuthr: I am removing the sleep as we don't really need them.

dirkmc (approved):
Great work 🚀

PR #1572, opened 2023-07-19
feat: manual publish storage deal message

LexLuthr:
## Current Scenario
Currently, Boost controls when the PSD message is sent for a deal. This decision is governed by a couple of factors.
1. A timeout value forcing to publish all deals (1 hour)
2. Maximum number of deal to be included in each message (8)
Another special case is to Publish all deals forcefully using UI button and Graphql mutation.

## Feature request background
SPs have expressed interest in being able to control how the deal are published. The reason this control ranges from being able to control deal flow to miner to control the cost of PSD message

## Feature
The new feature allows SPs to turn on manual PSD. Once this feature is turned on, Boost will no longer send any PSD messages unless explicitly prompted by the user. The feature can be turned on with the below config variable:

```
[Dealmaking]
        // When ExternalDealPublishControl is set to true, user is responsible to publish the deal message.
	// A list of pending deals can be queried and published using the GraphQL endpoint.
	// If this value is set to true, the values of MaxDealsPerPublishMsg and PublishMsgPeriod will be
	// ignored
	ExternalDealPublishControl bool

```

### Querying the deals pending to be published
The deals which have not been published can be queried using the Graphql endpoint of the Boost. The curl for the query:

`curl -X POST -H "Content-Type: application/json" -d '{"query":"query { dealPublish{ ManualPSD Deals {ID IsLegacy ClientAddress ProviderAddress CreatedAt PieceCid PieceSize ProviderCollateral StartEpoch EndEpoch ClientPeerID PublishCid Transfer { Type Size Params ClientID} Message }}}"}' http://localhost:8080/graphql/query | jq`

### Decision
SPs can create a decision script which queries the Graphql endpoint using the above curl and take a decision on which deal to be published.

### Publishing deals
Once the decision has been taken on which deals to be published, SPs can use the Grapqhql endpoint to Publish the deals.
The mutation is:

`publishPendingDeals(ids: [ID!]!): Boolean!`

which takes an array of deal UUIDs to be published and an error message is returned. The following curl can be used to publish the deal.

`curl -X POST -H "Content-Type: application/json" -d '{"query":"mutation {publishPendingDeals(ids: [\"cc78d877-f69c-410b-80a7-b23f2fab8951\", \"test1\"])}"}' http://localhost:8080/graphql/query | jq`

#### Successful output
```
{
  "data": {
    "publishPendingDeals": true
  }
}
```

#### Failure output

```
{
  "errors": [
    {
      "message": "failed to get deal details from DB cc78d877-f69c-410b-80a7-b23f2fab8951: scanning deal row: sql: no rows in result set",
      "path": [
        "publishPendingDeals"
      ]
    }
  ],
  "data": null
}
```

### Publish all deals
The publish all button in UI would have the same functionality as before. Same would be true for the Graphql mutation `dealPublishNow`

### UI screenshots
Screenshot 2023-07-19 at 11 09 14 AM


markets/storageadapter/dealpublisher.go
```diff
@@ -172,10 +187,17 @@ func (p *DealPublisher) ForcePublishPendingDeals() {
 }
 
 func (p *DealPublisher) Publish(ctx context.Context, deal market.ClientDealProposal) (cid.Cid, error) {
-	pdeal := newPendingDeal(ctx, deal)
+	pdeal, err := newPendingDeal(ctx, deal)
+	if err != nil {
+		return cid.Undef, fmt.Errorf("failed to generate cid for the deal proposal: %w", err)
```
dirkmc, line 192:
```suggestion
		return cid.Undef, fmt.Errorf("failed create pending deal: %w", err)
```

markets/storageadapter/dealpublisher.go
```diff
@@ -172,10 +187,17 @@ func (p *DealPublisher) ForcePublishPendingDeals() {
...
+	}
 
 	// Add the deal to the queue
-	p.processNewDeal(pdeal)
+	if p.manualPSD {
+		p.queueDeals(pdeal)
+	} else {
+		p.processNewDeal(pdeal)
+	}
```
dirkmc, line 200: I think we can simplify by just making some modifications to processNewDeal, rather than adding a new queueDeals method: in processNewDeal we should return after adding the deal to the queue if p.manualPSD is set

markets/storageadapter/dealpublisher.go
```diff
@@ -257,11 +279,13 @@ func (p *DealPublisher) waitForMoreDeals() {
 }
 
 func (p *DealPublisher) publishAllDeals() {
-	// If the timeout hasn't yet been cancelled, cancel it
-	if p.cancelWaitForMoreDeals != nil {
-		p.cancelWaitForMoreDeals()
-		p.cancelWaitForMoreDeals = nil
-		p.publishPeriodStart = time.Time{}
+	if !p.manualPSD {
```
dirkmc, line 282: I don't think you need to check against p.manualPSD here - as long as we never call waitForMoreDeals, cancelWaitForMoreDeals will always be nil

markets/storageadapter/dealpublisher.go
```diff
@@ -443,3 +467,83 @@ func (p *DealPublisher) filterCancelledDeals() {
 	}
 	p.pending = filtered
 }
+
+func (p *DealPublisher) queueDeals(pdeal *pendingDeal) {
```
dirkmc, line 471: I don't think we need the queueDeals method, let's just add some checks to processNewDeal

markets/storageadapter/dealpublisher.go
```diff
@@ -443,3 +467,83 @@ func (p *DealPublisher) filterCancelledDeals() {
...
+		return
+	}
+
+	// Add the new deal to the queue
+	p.pending = append(p.pending, pdeal)
+	log.Infof("add deal with piece CID %s to publish deals queue - %d deals in queue",
+		pdeal.deal.Proposal.PieceCID, len(p.pending))
+}
+
+func (p *DealPublisher) PublishQueuedDeals(deals []cid.Cid) (error, []cid.Cid) {
```
dirkmc, line 489: I'd suggest we combine PublishQueuedDeals with the existing publishAllDeals method. Instead of calling publishAddDeals, the existing code should call PublishQueuedDeals but pass all deals to it.

gql/resolver_dealpublish.go
```diff
@@ -163,3 +172,41 @@ func (r *resolver) DealPublishNow(ctx context.Context) (bool, error) {
 	r.publisher.ForcePublishPendingDeals()
 	return true, nil
 }
+
+func (r *resolver) PublishPendingDeals(ctx context.Context, args dealsToPublish) (bool, error) {
+	if !r.publisher.ManualPSD() {
+		return false, fmt.Errorf("manual deal publishing is disabled")
+	}
+
```
dirkmc, line 180: I think it's fine if people want to publish specific pending deals, even when manual psd is not enabled.
I'd just remove this check here.

markets/storageadapter/dealpublisher.go
```diff
@@ -443,3 +467,83 @@ func (p *DealPublisher) filterCancelledDeals() {
...
+		if !found {
+			notFound = append(notFound, c)
+		}
+	}
+
+	p.lk.Unlock()
+
+	if len(notFound) > 0 {
+		return fmt.Errorf("deals not found in the publisher queue"), notFound
+	}
```
dirkmc, line 514: I don't think we need to return an error here.
I'd suggest that the resolver should return a list of the deals that were published. For example if
- the user wants to publish deals [1, 2, 3]
- deal 2 was not found (maybe it was already published)
- the resolver should return [1, 3]

node/config/types.go
```diff
@@ -272,6 +272,12 @@ type DealmakingConfig struct {
 	// Whether to enable legacy deals on the Boost node or not. We recommend keeping
 	// them disabled. These will be completely deprecated soon.
 	EnableLegacyStorageDeals bool
+
+	// When set to true, user is responsible to publish the deal message.
+	// A list of pending deals can be queried and published using the GraphQL endpoint.
+	// If this value is set to true, the values of MaxDealsPerPublishMsg and PublishMsgPeriod will be
+	// ignored
+	ExternalDealPublishControl bool
```
dirkmc, line 280:
```suggestion
	// When set to true, the user is responsible for publishing deals manually.
	// The values of MaxDealsPerPublishMsg and PublishMsgPeriod will be
	// ignored, and deals will remain in the pending state until manually published.
	ManualDealPublish bool
```

react/src/DealPublish.js
```diff
@@ -40,6 +40,29 @@ function DealPublishContent() {
...
+                        {deals.length} deal{deals.length === 1 ? '' : 's'} pending to be published
+                    </p>
+
+                    <div className="buttons">
+                        <div className="button" onClick={doPublish}>Publish Now</div>
+                    </div>
+                </>
+            ) : null}
+
+            <h5>Note: Users must publish deals manually as external deal publish control is enabled</h5>
```
dirkmc, line 58:
```suggestion
            <h5>Note: Manual deal publishing is enabled in config: deals will not be published automatically, the Storage Provider must publish deals manually</h5>
```

nonsense:
It'd be nice to have a test for this, at least for the very happy case, probably an itest.

dirkmc:
Yes that makes sense 👍 
We already have some tests in dealpublisher_test.go, we should add more there.

LexLuthr:
@dirkmc I have modified the existing test to account for the new method. I need help with UI. I am not able to use the new mutation to Publish all deals. Can you please help me code the JS part?

gql/schema.graphql
```diff
@@ -598,9 +599,6 @@ type RootMutation {
   """Fail a Deal that was paused because of an error"""
   dealFailPaused(id: ID!): ID!
 
-  """Publish all pending deals now"""
-  dealPublishNow: Boolean!
-
```
dirkmc: Sorry Mayank I think I may have confused you with earlier comments.
I don't think we should remove the dealPublishNow endpoint. For SPs who are still following the previous flow it's easier for them to use curl invoke dealPublishNow and publish all deal IDs (otherwise they have to first make a request to get all the IDs and then call publishPendingDeals).
My suggestion to remove the publish all option was only for the internals of boost, ie within the DealPublisher class.

markets/storageadapter/dealpublisher.go
```diff
@@ -94,6 +105,9 @@ type PublishMsgConfig struct {
 	MaxDealsPerMsg uint64
 	// Minimum start epoch buffer to give time for sealing of sector with deal
 	StartEpochSealingBuffer uint64
+	// Enable ExternalDealPublishControl to allow users to choose when to send Publish messages
+	// for each deal
+	ExternalDealPublishControl bool
```
dirkmc, line 110: Let's also change this from ExternalDealPublishControl to ManualPSD for consistency

markets/storageadapter/dealpublisher.go
```diff
@@ -161,18 +176,11 @@ func (p *DealPublisher) PendingDeals() api.PendingDealInfo {
...
-// ForcePublishPendingDeals publishes all pending deals without waiting for
-// the publish period to elapse
-func (p *DealPublisher) ForcePublishPendingDeals() {
-	p.lk.Lock()
-	defer p.lk.Unlock()
-
-	log.Infof("force publishing deals")
-	p.publishAllDeals()
-}
-
```
dirkmc, line 173: As I mentioned above I'd suggest we keep ForcePublishPendingDeals()

markets/storageadapter/dealpublisher_test.go
```diff
@@ -186,7 +187,7 @@ func TestDealPublisher(t *testing.T) {
 	}
 }
 
-func TestForcePublish(t *testing.T) {
+func TestPublishPendingDeals(t *testing.T) {
```
dirkmc, line 190: Let's also add a test for publishing less than all deals.
eg publish only 1 of 2 pending deals.

LexLuthr:
Closing this in favour of https://github.com/filecoin-project/boost/pull/1585

PR #1584, opened 2023-07-20
Use shared piece reader for blockstore get (fixed conflicts)

nonsense:


piecedirectory/piecedirectory.go
```diff
@@ -347,61 +349,49 @@ type cachedSectionReader struct {
...
 }
 
 func (r *cachedSectionReader) Close() error {
-	r.ps.prCacheLk.Lock()
-	defer r.ps.prCacheLk.Unlock()
-
-	r.refs--
-	if r.refs == 0 {
-		r.ps.pieceReaderCache.Remove(r.pieceCid)
-		return r.SectionReader.Close()
```
dirkmc: I think we want to close the underlying section reader here.

piecedirectory/piecedirectory.go
```diff
@@ -347,61 +349,49 @@ type cachedSectionReader struct {
...
 		r = &cachedSectionReader{
 			ps:       ps,
 			pieceCid: pieceCid,
 			ready:    make(chan struct{}),
-			refs:     1,
 		}
-		ps.pieceReaderCache.Add(pieceCid, r)
-	}
-	ps.prCacheLk.Unlock()
+		ps.pieceReaderCache.Set(pieceCid.String(), r)
```
dirkmc: I think we need to do the Get then Set under a lock

piecedirectory/piecedirectory.go
```diff
@@ -347,61 +349,49 @@ type cachedSectionReader struct {
...
 
-		ps.prCacheLk.Lock()
 		r.SectionReader = sr
 		r.err = err
-		ps.prCacheLk.Unlock()
 
 		// Inform any waiting threads that the cached reader is ready
 		close(r.ready)
 	} else {
+		r = rr.(*cachedSectionReader)
```
dirkmc, line 393: Can we use templating to avoid needing to cast here? I'm not sure if the new lib supports templating.

piecedirectory/piecedirectory.go
nonsense: Closing it now in the expire callback.

piecedirectory/piecedirectory.go
nonsense: Reverted the lock.

piecedirectory/piecedirectory.go
nonsense: Unfortunately this `ttlcache` doesn't include generics.

nonsense:
Ready for review again.

PR #1585, opened 2023-07-20
feat: manual publish storage deal message

LexLuthr:
## Current Scenario
Currently, Boost controls when the PSD message is sent for a deal. This decision is governed by a couple of factors.
1. A timeout value forcing to publish all deals (1 hour)
2. Maximum number of deal to be included in each message (8)
Another special case is to Publish all deals forcefully using UI button and Graphql mutation.

## Feature request background
SPs have expressed interest in being able to control how the deal are published. The reason this control ranges from being able to control deal flow to miner to control the cost of PSD message

## Feature
The new feature allows SPs to turn on manual PSD. Once this feature is turned on, Boost will no longer send any PSD messages unless explicitly prompted by the user. The feature can be turned on with the below config variable:

```
[Dealmaking]
        // When set to true, the user is responsible for publishing deals manually.
	// The values of MaxDealsPerPublishMsg and PublishMsgPeriod will be
	// ignored, and deals will remain in the pending state until manually published.
	ManualDealPublish bool

```

### Querying the deals pending to be published
The deals which have not been published can be queried using the Graphql endpoint of the Boost. The curl for the query:

`curl -X POST -H "Content-Type: application/json" -d '{"query":"query { dealPublish{ ManualPSD Deals {ID IsLegacy ClientAddress ProviderAddress CreatedAt PieceCid PieceSize ProviderCollateral StartEpoch EndEpoch ClientPeerID PublishCid Transfer { Type Size Params ClientID} Message }}}"}' http://localhost:8080/graphql/query | jq`

### Decision
SPs can create a decision script which queries the Graphql endpoint using the above curl and take a decision on which deal to be published.

### Publishing deals
Once the decision has been taken on which deals to be published, SPs can use the Grapqhql endpoint to Publish the deals.
The mutation is:

`publishPendingDeals(ids: [ID!]!): [ID!]!`

which takes an array of deal UUIDs to be published and an error message is returned. The following curl can be used to publish the deal.

`curl -X POST -H "Content-Type: application/json" -d '{"query":"mutation {publishPendingDeals(ids: [\"d9f849f1-d5d8-4bfc-b034-2866bddfc8cb\", \"a8eb58ef-7381-4251-ae7a-1227c032c0b9\"])}"}' http://localhost:8080/graphql/query | jq`

#### Successful output
```
{
  "data": {
    "publishPendingDeals": [
      "d9f849f1-d5d8-4bfc-b034-2866bddfc8cb",
      "a8eb58ef-7381-4251-ae7a-1227c032c0b9"
    ]
  }
}
```

#### Failure output

```
{
  "errors": [
    {
      "message": "failed to get deal details from DB cc78d877-f69c-410b-80a7-b23f2fab8951: scanning deal row: sql: no rows in result set",
      "path": [
        "publishPendingDeals"
      ]
    }
  ],
  "data": null
}
```

### Publish all deals
The publish all button in UI would have the same functionality as before. Same would be true for the Graphql mutation `dealPublishNow`

### UI screenshots
Screenshot 2023-07-19 at 11 09 14 AM

### DIFF
--- a/markets/storageadapter/dealpublisher.go
+++ b/markets/storageadapter/dealpublisher.go
@@ -495,7 +495,7 @@ func (p *DealPublisher) PublishQueuedDeals(deals []cid.Cid) (error, []cid.Cid) {
 	for _, c := range deals {
 		found := false
 		for _, pd := range p.pending {
-			if c == pd.proposalCid {
+			if c.Equals(pd.proposalCid) {
 				found = true
 				toPublish = append(toPublish, pd)
 				break
--- a/markets/storageadapter/dealpublisher.go
+++ b/markets/storageadapter/dealpublisher.go
@@ -176,28 +176,14 @@ func (p *DealPublisher) PendingDeals() api.PendingDealInfo {
 	}
 }
 
-// ForcePublishPendingDeals publishes all pending deals without waiting for
-// the publish period to elapse
-func (p *DealPublisher) ForcePublishPendingDeals() {
-	p.lk.Lock()
-	defer p.lk.Unlock()
-
-	log.Infof("force publishing deals")
-	p.publishAllDeals()
-}
-
 func (p *DealPublisher) Publish(ctx context.Context, deal market.ClientDealProposal) (cid.Cid, error) {
 	pdeal, err := newPendingDeal(ctx, deal)
 	if err != nil {
-		return cid.Undef, fmt.Errorf("failed to generate cid for the deal proposal: %w", err)
+		return cid.Undef, fmt.Errorf("failed create pending deal: %w", err)
 	}
 
 	// Add the deal to the queue
-	if p.manualPSD {
-		p.queueDeals(pdeal)
-	} else {
-		p.processNewDeal(pdeal)
-	}
+	p.processNewDeal(pdeal)
 
 	// Wait for the deal to be submitted
 	select {
@@ -230,6 +216,11 @@ func (p *DealPublisher) processNewDeal(pdeal *pendingDeal) {
 	p.pending = append(p.pending, pdeal)
 	log.Infof("add deal with piece CID %s to publish deals queue - %d deals in queue (max queue size %d)",
 		pdeal.deal.Proposal.PieceCID, len(p.pending), p.maxDealsPerPublishMsg)
+
+	// Return from here if manual PSD is enabled
+	if p.manualPSD {
+		return
+	}
 
 	// If the maximum number of deals per message has been reached or we're not batching, send a
 	// publish message
@@ -279,13 +270,12 @@ func (p *DealPublisher) waitForMoreDeals() {
 }
 
 func (p *DealPublisher) publishAllDeals() {
-	if !p.manualPSD {
-		// If the timeout hasn't yet been cancelled, cancel it
-		if p.cancelWaitForMoreDeals != nil {
-			p.cancelWaitForMoreDeals()
-			p.cancelWaitForMoreDeals = nil
-			p.publishPeriodStart = time.Time{}
-		}
+
+	// If the timeout hasn't yet been cancelled, cancel it
+	if p.cancelWaitForMoreDeals != nil {
+		p.cancelWaitForMoreDeals()
+		p.cancelWaitForMoreDeals = nil
+		p.publishPeriodStart = time.Time{}
 	}
 
 	// Filter out any deals that have been cancelled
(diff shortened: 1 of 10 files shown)


markets/storageadapter/dealpublisher.go
```diff
@@ -94,6 +105,9 @@ type PublishMsgConfig struct {
 	MaxDealsPerMsg uint64
 	// Minimum start epoch buffer to give time for sealing of sector with deal
 	StartEpochSealingBuffer uint64
+	// Enable ExternalDealPublishControl to allow users to choose when to send Publish messages
+	// for each deal
+	ExternalDealPublishControl bool
```
dirkmc, line 110: I'd suggest for consistency we also rename ExternalDealPublishControl to ManualPSD

markets/storageadapter/dealpublisher.go
```diff
@@ -443,3 +467,63 @@ func (p *DealPublisher) filterCancelledDeals() {
...
+			if c.Equals(pd.proposalCid) {
+				found = true
+				toPublish = append(toPublish, pd)
+				break
+			}
+		}
+		if found {
+			ret = append(ret, c)
+		} else {
+			log.Debugf("failed to find the proposal %S in pending deals", c)
```
dirkmc, line 489:
```suggestion
			log.Debugf("failed to find the proposal %s in pending deals", c)
```

markets/storageadapter/dealpublisher.go
```diff
@@ -443,3 +467,63 @@ func (p *DealPublisher) filterCancelledDeals() {
...
+		}
+	}
+
```
dirkmc, line 492: Do we need to cancel deals whose context has expired?

markets/storageadapter/dealpublisher.go
```diff
@@ -443,3 +467,63 @@ func (p *DealPublisher) filterCancelledDeals() {
...
+	p.lk.Unlock()
+
+	log.Infof("publishing deal proposals: %s", ret)
+
+	// Remove the deal from Pending
+	p.cleanupPending(deals)
```
dirkmc, line 498: I'd suggest we clean up pending deals under the lock also, so that the whole operation is atomic.
Otherwise two threads could interfere with each other.
Maybe you can even refactor so that it splits the list into the items to be published and the items that will remain (you're doing this in separate for loops at the moment, probably you can do both in the same for loop).

markets/storageadapter/dealpublisher_test.go
```diff
@@ -232,6 +233,61 @@ func TestForcePublish(t *testing.T) {
 	checkPublishedDeals(t, dpapi, dealsToPublish, []int{2})
 }
 
+func TestPublishPendingDeals(t *testing.T) {
```
dirkmc, line 236: Can we also add a test that only publishes some of the deals instead of all of them.
eg there are three pending deals, only publish two of them and verify that there is one remaining pending deal.

LexLuthr: I have refactored the test to send 4 deals. 1 with ctx.Error and other 3 normal. It will try to Publish 3 deals total (2 correct ones and 1 random CID which is not in Publisher). Test will result in 2 deal Publish, 1 debug log with error about random CID (not found in publisher) and returning 1 pending deal.

markets/storageadapter/dealpublisher.go
LexLuthr: I have refactored the code a little. Unique identifier don't go well with slice.
The new code looks more clean. It resulted in some minor changes in other functions as well.

markets/storageadapter/dealpublisher.go
```diff
@@ -266,10 +289,13 @@ func (p *DealPublisher) publishAllDeals() {
...
 	p.filterCancelledDeals()
-	deals := p.pending
-	p.pending = nil
 
 	// Send the publish message
+	var deals []*pendingDeal
+	for _, deal := range p.pending {
+		deals = append(deals, deal)
+	}
+	p.pending = nil
```
dirkmc, line 298:
```suggestion
	p.pending = make(map[cid.Cid]*pendingDeal)
```

markets/storageadapter/dealpublisher.go
```diff
@@ -434,12 +460,41 @@ func pieceCids(deals []market.ClientDealProposal) string {
...
-	for _, pd := range p.pending {
+	filtered := make(map[cid.Cid]*pendingDeal)
+	for c, pd := range p.pending {
 		if pd.ctx.Err() != nil {
 			continue
 		}
-		filtered = append(filtered, pd)
+		filtered[c] = pd
 	}
 	p.pending = filtered
```
dirkmc, line 470: Rather than creating a new map, I think you can just delete from the existing map

markets/storageadapter/dealpublisher_test.go
```diff
@@ -232,6 +233,69 @@ func TestForcePublish(t *testing.T) {
...
+	publishDeal(t, dp, 0, true, false)
+	// 3. Regular deal
+	publishDeal(t, dp, 0, false, false)
+	// 4. Regular deal
+	publishDeal(t, dp, 0, false, false)
+
+	// Allow a moment for them to be queued
+	build.Clock.Sleep(10 * time.Millisecond)
+
+	// Should be two deals in the pending deals list
```
dirkmc, line 260:
```suggestion
	// Should be three deals in the pending deals list
```

markets/storageadapter/dealpublisher_test.go
```diff
@@ -232,6 +233,69 @@ func TestForcePublish(t *testing.T) {
...
+	}
+
+	toPublish := pcids[1:]
+	pending := []cid.Cid{pcids[0]}
+
+	// Send an additional CID not present in publisher
+	c, err := cid.Decode("bafy2bzacea3wsdh6y3a36tb3skempjoxqpuyompjbmfeyf34fi3uy6uue42v4")
+	require.NoError(t, err)
+
+	// Publish all pending deals and verify all have been published
```
dirkmc, line 280:
```suggestion
	// Publish three pending deals and verify all deals whose context has not expired have been published
```

markets/storageadapter/dealpublisher_test.go
```diff
@@ -232,6 +233,69 @@ func TestForcePublish(t *testing.T) {
...
+	publishedDeals := dp.PublishQueuedDeals(append(toPublish, c))
+	require.Equal(t, toPublish, publishedDeals)
+
+	// Should be no pending deals
```
dirkmc, line 284:
```suggestion
	// Should be one remaining pending deal
```

dirkmc (approved):

PR #1589, opened 2023-07-21
feat: add corshandler to IPFS gateway

### LINKED ISSUE #1583: Implement CORS support in booster-http
The feature request is to allow CORS handling for IPFS gateway within booster-http. It will allow Magmo team to easily integrate go-nitro. It can also be utilised by other builders within the ecosystem.

LexLuthr:
fixes #1583


cmd/booster-http/http_test.go
```diff
@@ -27,10 +27,19 @@ func TestNewHttpServer(t *testing.T) {
 	require.NoError(t, err)
 	waitServerUp(t, 7777)
 
-	// Check that server is responding with 200 status code
-	resp, err := http.Get("http://localhost:7777/")
+	// Create a request with Cors header
```
dirkmc, line 30: The new test looks good - I'd say let's also keep the old test. So first check a regular request returns 200, then check that the CORS request returns the expected headers.

gql/server.go
```diff
@@ -49,6 +51,13 @@ func (s *Server) Start(ctx context.Context) error {
 
 	// Serve React app
 	mux := http.NewServeMux()
+
+	// Get new cors
+	options := cors.Options{
+		AllowedHeaders: []string{"*"},
+	}
+	c := cors.New(options)
+
```
dirkmc, line 60: I think we can add CORS request handling in the Start method or in NewServer but probably don't need both.

dirkmc (approved):

PR #1590, opened 2023-07-24
fix version flag for booster-http and booster-bitswap

nonsense:


cmd/booster-bitswap/init.go
```diff
@@ -115,13 +114,3 @@ var initCmd = &cli.Command{
 		return err
 	},
 }
-
-var versionCmd = &cli.Command{
```
nonsense, line 119: No need for this, `--version` flag is supported out of the box with `urfave/cli`

cmd/booster-bitswap/main.go
```diff
@@ -22,16 +23,15 @@ func main() {
 		Name:                 "booster-bitswap",
 		Usage:                "Bitswap endpoint for retrieval from Filecoin",
 		EnableBashCompletion: true,
+		Version:              build.UserVersion(),
 		Flags: []cli.Flag{
-			cliutil.FlagVerbose,
```
nonsense, line 26: We shouldn't use `-v` for `verbose`, it is used for `--version` typically.

dirkmc (approved):

LexLuthr (approved):

nonsense:
> I agree with the flag naming convention. We should stick to traditional naming. We also need to update other binaries and docs as well.
> 
> -v, --version = version -V, --verbose = Verbose

I think `-v` for `--version` and `--vv` for `--very-verbose` is probably good enough. Then we have 2 levels of logging, and anything else is really custom and probably belongs to `set-log-level` command.

PR #1592, opened 2023-07-26
feat: show ad chain head in UI

LexLuthr:
This PR adds an extra row to the Provider info in Indexer page of Boost UI.
Users can compare the head to last sync ad and figure out if their sync is working or not.

Screenshot 2023-07-26 at 12 51 35 PM

### DIFF
--- a/react/src/Ipni.js
+++ b/react/src/Ipni.js
@@ -5,6 +5,7 @@ import {
     IpniAdEntriesQuery,
     IpniAdQuery,
     IpniProviderInfoQuery,
+    IpniLatestAdQuery,
     RetrievalLogsCountQuery, RetrievalLogsListQuery,
 } from "./gql";
 import moment from "moment";
@@ -52,6 +53,7 @@ function ProviderInfo(props) {
 }
 
 function ProviderIpniInfo({peerId}) {
+    const head = useQuery(IpniLatestAdQuery)
     const [{loading, error, data}, setResp] = useState({ loading: true })
     const idxHost = indexerHost()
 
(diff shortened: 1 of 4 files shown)


react/src/Ipni.js
```diff
@@ -75,6 +77,11 @@ function ProviderIpniInfo({peerId}) {
     if (loading) return <div>Loading...</div>
     if (!data) return null
 
+    let defined = false
+    if (head.data !== undefined ) {
+        defined = true
+    }
+
```
dirkmc, line 84:
```suggestion
    let defined = false
    if (head.data !== undefined ) {
        defined = true
    }

```

react/src/Ipni.js
```diff
@@ -98,6 +105,12 @@ function ProviderIpniInfo({peerId}) {
                     <span className="aux">({moment(data.LastAdvertisementTime).fromNow()} ago)</span>
                 </td>
             </tr>
+            <tr>
+                <th>Latest Advertisement on Boost</th>
+                <td>
+                    {defined ? <Link to={'/ipni/ad/'+head.data.ipniLatestAdvertisement}>{head.data.ipniLatestAdvertisement}</Link>: ''}
```
dirkmc, line 111:
```suggestion
                    {head.data ? <Link to={'/ipni/ad/'+head.data.ipniLatestAdvertisement}>{head.data.ipniLatestAdvertisement}</Link>: ''}
```

dirkmc (approved):
I suggested a small refactor to simplify, but otherwise LGTM 👍

PR #1600, opened 2023-08-01
multi-miner / single LID - piece doctor

dirkmc:
TODO:
- [x] Test migration on local devnet
- [ ] Test in production


extern/boostd-data/svc/setup_yugabyte_test_util.go
```diff
@@ -15,37 +18,128 @@ var tlog = logging.Logger("ybtest")
...
-	ConnectString: "postgresql://postgres:postgres@localhost:5433",
+	ConnectString: "postgresql://postgres:postgres@localhost:5433?sslmode=disable",
 }
 
-func SetupYugabyte(t *testing.T) {
-	ctx := context.Background()
+func init() {
+	// Uncomment when testing against a local yugabyte instance.
+	//TestYugabyteSettings = TestYugabyteSettingsLocal
+}
```
nonsense, line 33: We could probably just move this to `setup_yugabyte_test_util_test.go` which should be compiled only if you run `go test`, no?

piecedirectory/doctor.go
```diff
@@ -135,10 +142,15 @@ func (d *Doctor) checkPiece(ctx context.Context, pieceCid cid.Cid, lu *sectorsta
 		}
 	}
 
+	if !hasDealsOnThisMiner {
+		doclog.Debugw("ignoring piece as it is not present in any deals on this miner", "piece", pieceCid.String(), "miner", d.maddr.String())
+		return nil
+	}
```
nonsense, line 148: I'm confused -- we go over all pieces stored in LID on all boostd instances?

extern/boostd-data/svc/setup_yugabyte_test_util.go
dirkmc: We only want it when running `go test` on a local machine though (not in CI)

piecedirectory/doctor.go
dirkmc: This is just a sanity check. There are two different processes that populate the list of pieces to check, and that add / remove deals. So it's possible that
- the process that adds to the list of pieces to check adds the piece from deal A
- another process deletes deal A
- GetNextPiecesToCheck returns the piece cid from deal A
- there are no deals for that piece

extern/boostd-data/yugabyte/create.cql
```diff
@@ -1,6 +1,4 @@
-create keyspace if not exists idx with replication = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };
```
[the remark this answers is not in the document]
dirkmc: That command is now run as a preparatory step in the CreateKeyspace function. I moved it there so we can create a different keyspace for tests.

piecedirectory/doctor.go
dirkmc: You're right that this should happen very rarely. I updated the log level to a warning.

nonsense (approved):

PR #1607, opened 2023-08-04
Make boost types importable by other go modules

masih:
Boost go module included a `replace` directive to `boostd-data` with a module name for which no repository exists. This prohibited boost types, specifically transfer query and markets/deal making types, to be imported by other go modules without also forking the boost repo internal directories. As a result a number of projects ended up forking code.

The changes here introduce a go workspace that allows the local development to continue as it was before, while keeping the boost go module importable.

Upon releasing a new version however the development should explicitly `go get` the version of `boostd-data` they want to include in the release.


go.mod
```diff
@@ -326,7 +324,7 @@ require (
 
 require (
 	github.com/NYTimes/gziphandler v1.1.1
-	github.com/filecoin-project/boostd-data v0.0.0-00010101000000-000000000000
+	github.com/filecoin-project/boost/extern/boostd-data v0.0.0-20230804101708-7565454b8600
```
masih, line 327: This is set to a commit from this branch since commits to main do not contain this code yet, nor any specific tag.

We should update this to a commit from main or a tag once one such thing exists. This is the dependency version that will be pulled by other go mods depending on boost.

masih:
The advice from boost dev seems to be that it's expected that the users fork necessary types. 

Closing.

Issue #1550, opened 2023-07-03
Dockerfile for boost client


mrtechit:
### Checklist

- [X] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [X] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the Boost forum](https://github.com/filecoin-project/boost/discussions/categories/ideas) and select the category as `Ideas`.
- [X] I **have** a specific, actionable, and well motivated feature request to propose.

### Boost component

- [ ] boost daemon - storage providers
- [X] boost client
- [ ] boost UI
- [ ] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

I want to use boost client on amazon linux, however some of the packages are not available (at least during my initial investigation). So its better to have a dockerfile

### Describe the solution you'd like

Dockerfile for all boost binaries

### Describe alternatives you've considered

Use a different AWS EC2 image instead of amazon linux

LexLuthr:
@mrtechit We have a dockerfile for devnet at https://github.com/filecoin-project/boost/blob/main/docker/devnet/Dockerfile.source

Please feel free to hack through it to remove lotus related parts and build a mainnet binary of Boost instead of devnet.

LexLuthr:
Task list:
1. Use  https://github.com/filecoin-project/boost/blob/main/docker/devnet/Dockerfile.source for reference
2. We can get rid of Node related stuff as client doesn't have UI
3. We can drop rust related stuff as we don't need to compile `boostx`
4. Create under docker/boost-client/
5. Use Golang 1.20 image
6. Drop lotus related binaries

alvin-reyes:
Do we just want a containerized `boost` or do we need to run specific commands on the `entrypoint?`

alvin-reyes:
@mrtechit do you just need a containerized `boost` (boost-client) without any of the dependencies or do you also need boostd,boostx,boost-data?

alvin-reyes:
Draft: https://github.com/filecoin-project/boost/compare/main...alvin-reyes:boost:main

PR #1616, opened 2023-08-14
feat: add docker to containerize a standalone boost-client

alvin-reyes:
Fixes https://github.com/filecoin-project/boost/issues/1550
# Changes
- Added a dockerfile for `boost` client. 
- Add some entrypoint.sh and a few commands to run when the container runs.


docker/boost-client/Dockerfile.source
```diff
@@ -0,0 +1,73 @@
...
+      description="This image is used to host the boost-dev storage provider"
+
+ENV BOOST_PATH /var/lib/boost
+VOLUME /var/lib/boost
+EXPOSE 8080
+
+COPY --from=builder /go/src/boost /usr/local/bin/
+COPY docker/boost-client/entrypoint.sh /app/
+COPY docker/boost-client/sample/* /app/sample/
+RUN boost -v
```
LexLuthr, line 71: We can probably get rid of this line. This command is not a deamon or long running process. I think container will exit after this. Did we test this Dockerfile?

docker/boost-client/Dockerfile.source
```diff
@@ -0,0 +1,73 @@
...
+COPY .git/modules/extern/filecoin-ffi /go/src/.git/modules/extern/filecoin-ffi
+
+RUN make build/.filecoin-install
+##############################################
+COPY . /go/src
+##############################################
+
+RUN --mount=type=cache,target=/root/.cache/go-build \
+    --mount=type=cache,target=/go/pkg/mod \
+      make debug
```
LexLuthr, line 35: We should use `make build`

docker/boost-client/Dockerfile.source
```diff
@@ -0,0 +1,73 @@
...
+WORKDIR /app
+
+#########################################################################################
+FROM runner as boost-main
+
+ARG BUILD_VERSION=0.1
+
+LABEL org.opencontainers.image.version=$BUILD_VERSION \
+      org.opencontainers.image.authors="Boost Dev Team" \
+      name="boost-main" \
```
LexLuthr, line 56:
```suggestion
      name="boost-client" \
```

docker/boost-client/Dockerfile.source
```diff
@@ -0,0 +1,73 @@
...
+      release=$BUILD_VERSION \
+      summary="This image is used to host the boost-dev storage provider" \
+      description="This image is used to host the boost-dev storage provider"
+
+ENV BOOST_PATH /var/lib/boost
+VOLUME /var/lib/boost
+EXPOSE 8080
+
+COPY --from=builder /go/src/boost /usr/local/bin/
+COPY docker/boost-client/entrypoint.sh /app/
```
LexLuthr, line 69: We probably don't need this.

docker/boost-client/sample/make-a-deal.sh
LexLuthr, line 1: This file will not work outside of the devnet. We have no lotus binaries anymore.

LexLuthr (changes requested):
We need to clean up sample deal making as we don't have all binaries for it. The init script is also not required.

docker/boost-client/entrypoint.sh
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
alvin-reyes: We can just still add it but only put a boost -v command so that the user has confirmation that boost is installed.

```
#!/usr/bin/env bash
set -e
boost -v
```

docker/boost-client/Dockerfile.source
alvin-reyes: For line 68 - we can't remove this line cause this is the actual boost binary.
For line 69 -  It will exit but I think it's better to have the entrypoint still and just put `boost -v` to confirm that the binary is there and it's on the env path. 

```
#!/usr/bin/env bash
set -e
boost -v
```

IMO, we can remove 70 and 71.

The container will exit but the point is to create the image that can be used to build an containerized environment rather than a running process.

Yes this was tested, here's the output (verification) after running it on my local dockerhub.
image

docker/boost-client/Dockerfile.source
alvin-reyes: ack.

docker/boost-client/Dockerfile.source
alvin-reyes: For line 68 - we can't remove this line cause this is the actual boost binary.
For line 69 - It will exit but I think it's better to have the entrypoint still and just put boost -v to confirm that the binary is there and it's on the env path.

```
#!/usr/bin/env bash
set -e
boost -v
```

docker/boost-client/sample/make-a-deal.sh
alvin-reyes: I'm thinking we can just connect to a calibration network and use a test network address with Datacap. So on the sample folder, we'll add a "test net folder with `wallet`, COPY it over to environment.

We then cp the wallet to the `~/.boost-client/wallet` folder before running a dummy verified deal.

alvin-reyes: If not, we can just remove the file/sh script.

docker/boost-client/Dockerfile.source
alvin-reyes: resolved on the latest commit.

LexLuthr (commented):
Changes look good now. Can you build the image for both Linux and MacOS once and try to run them alongside an already running devnet? If everything looks okay, then please make a deal and retrieve from this new container.

docker/boost-client/Dockerfile.source
alvin-reyes: @LexLuthr, should we use `make boost` instead?

alvin-reyes:
## Built and ran on Ubuntu
- Build successful
- I was able to run devnet and made a deal.


image

I had to run `boostx market-add` to put some FIL in.

```
boost deal --verified=false --provider=t01000 --http-url=http://demo-http-server/sample.car --commp=baga6ea4seaqpfonfro6j3gcw4uxkxbivlqn2fghx5dpulc6sbi5noz7bcvzmuiq --car-size=1175 --piece-size=2048 --payload-cid=bafk2bzacec3jst4tkh424chatp273o6rxvipfg54kphd56gaxobpcdtr2sgco --storage-price 20000000000
```


## Build and ran on MacOs
- Build successful
- I was able to run devnet and made a deal.

image


I had to run `boostx market-add` to put some FIL in.

```
boost deal --verified=false --provider=t01000 --http-url=http://demo-http-server/sample.car --commp=baga6ea4seaqpfonfro6j3gcw4uxkxbivlqn2fghx5dpulc6sbi5noz7bcvzmuiq --car-size=1175 --piece-size=2048 --payload-cid=bafk2bzacec3jst4tkh424chatp273o6rxvipfg54kphd56gaxobpcdtr2sgco --storage-price 20000000000
```

LexLuthr (approved):

PR #1617, opened 2023-08-14
fix: add the padding to the carv2 size

alvin-reyes:
# Changes
- Add the data padding to the CarV2 size.

### DIFF
--- a/storagemarket/deal_commp.go
+++ b/storagemarket/deal_commp.go
@@ -193,14 +193,13 @@ func GenerateCommP(filepath string) (*abi.PieceInfo, error) {
 }
 
 func getCarSize(filepath string, rd *carv2.Reader) (int64, error) {
+	var size int64
 	if rd.Version == 1 || rd.Version == 2 {
 		st, err := os.Stat(filepath)
 		if err != nil {
 			return 0, fmt.Errorf("failed to get CAR file size: %w", err)
 		}
-		return st.Size(), nil
+		size = st.Size()
 	}
-
-	return 0, fmt.Errorf("unsupported CAR version: %d", rd.Version)
+	return size, nil
 }
-
--- a/storagemarket/deal_commp.go
+++ b/storagemarket/deal_commp.go
@@ -196,7 +196,8 @@ func getCarSize(filepath string, rd *carv2.Reader) (int64, error) {
 	var size int64
 	switch rd.Version {
 	case 2:
-		size = int64(rd.Header.DataSize)
+		padding := rd.Header.DataOffset - carv2.PragmaSize - carv2.HeaderSize
+		size = int64(rd.Header.DataSize) + int64(padding)
 	case 1:
 		st, err := os.Stat(filepath)
 		if err != nil {
--- a/storagemarket/deal_commp.go
+++ b/storagemarket/deal_commp.go
@@ -197,6 +197,9 @@ func getCarSize(filepath string, rd *carv2.Reader) (int64, error) {
 	switch rd.Version {
 	case 2:
 		padding := rd.Header.DataOffset - carv2.PragmaSize - carv2.HeaderSize
+		if padding < 0 {
+			return 0, fmt.Errorf("CARv2 file is invalid or malformed: padding is negative")
+		}
 		size = int64(rd.Header.DataSize) + int64(padding)
 	case 1:
 		st, err := os.Stat(filepath)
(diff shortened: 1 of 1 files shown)


storagemarket/deal_commp.go
```diff
@@ -196,7 +196,8 @@ func getCarSize(filepath string, rd *carv2.Reader) (int64, error) {
 	var size int64
 	switch rd.Version {
 	case 2:
-		size = int64(rd.Header.DataSize)
+		padding := rd.Header.DataOffset - carv2.PragmaSize - carv2.HeaderSize
+		size = int64(rd.Header.DataSize) + int64(padding)
```
nonsense, line 200: This is certainly changing the `size`, which is then used in `storagemarket/deal_commp.go:131`. I am a bit confused how that is not changing the behavior and breaking existing functionality...

alvin-reyes: My guess is we haven't had any V2 CAR files that had any padding.

nonsense:
Could you re-push the commit, so that it triggers CI ?

storagemarket/deal_commp.go
```diff
@@ -196,7 +196,8 @@ func getCarSize(filepath string, rd *carv2.Reader) (int64, error) {
 	var size int64
 	switch rd.Version {
 	case 2:
-		size = int64(rd.Header.DataSize)
+		padding := rd.Header.DataOffset - carv2.PragmaSize - carv2.HeaderSize
```
nonsense, line 199: I think we should probably check that `padding` > 0, and log an error if this is not the case, including the `carv2.PragmaSize` and `carv2.HeaderSize` if they are malformed.

alvin-reyes: I think having a padding is normal but it is optional so having a 0 padding means that the size on the header is correct. 

We can log info if the padding > 0, but i don’t think its an error.

nonsense: I am worried about the malformed case, where padding ends up being negative, and therefore we calculate `size` to be smaller than it actually is.

alvin-reyes: ah got it. If it's negative value then we can add a error log to it i.e if `padding < 0`, we log error.

alvin-reyes:
```
padding := rd.Header.DataOffset - carv2.PragmaSize - carv2.HeaderSize
if padding < 0 {
	return 0, fmt.Errorf("CARv2 file is invalid: padding is negative")
}
```

storagemarket/deal_commp.go
```diff
@@ -196,13 +196,15 @@ func getCarSize(filepath string, rd *carv2.Reader) (int64, error) {
 	var size int64
 	switch rd.Version {
 	case 2:
-		size = int64(rd.Header.DataSize)
```
willscott, line 199: We should be trusting this in the context where it's a carv2 that boost has generated locally.

I think the main issue underlying this from my perspective is that this call is being used on an inbound deal (e.g. from `verifyCommP`). We should be less smart here and just look at the file's size on disk, rather than trying to parse the car at all?

consider the case where the deal is a carv2 file with extra data concatenated onto it - we don't want to trust the supplied data from the carv2 header as representing the full deal.

nonsense: As you say, this is used on inbound deals, and we can't trust these files - they are not generated by Boost.

We check the file size when rd.Version is 1, I guess we should just do the same for rd.Version 2, i.e. for all car files. cc @alvin-reyes

willscott:
There's something else here going on that I don't fully understand. Perhaps @dirkmc  has insight?

The surrounding call to `GenerateCommP` (https://github.com/filecoin-project/boost/blob/main/storagemarket/deal_commp.go#L150) - is only running over the [carv1 blocks](https://github.com/filecoin-project/boost/blob/main/storagemarket/deal_commp.go#L164-L169) of the file.

Is there a reason we aren't able to directly commp the bytes of the file?

my guess is that we have code with the following flow:
* carv1 gets made as the deal
* boost may have a local carv2 wrapped around that carv1 with index (although i thought we kept indexes separate?)
* we read the carv2 and commp over the inner carv1, which is the deal.

In general though, we want commp to just directly be over the bytes of the deal, we shouldn't be trying to open a carv1 or a carv2 data reader if we're expecting the deal to be able to contain arbitrary bytes or a data segment index.

alvin-reyes:
@dirkmc mentioned that him and @nonsense will check this one. I’ll stay on the sides for now.

PR #1627, opened 2023-08-16
direct data onboarding - disabled

nonsense:
This PR is adding:
- `import-direct` CLI command to trigger direct data onboarding
- `DirectData` Boost API
- FIL+ support for the Boost devnet
- the initial provider for handling direct data onboard calls from CLI
- initial db tables and migrations, since direct data doesn't fit the existing Boost Deal model
- initial direct data model (called `DirectDataEntry`)

So far this PR lacks:
- ~~[ ] tests - to be added only once we are happy with behavior and makes sense to lock it~~ - we skip tests for now, as DDO stays disabled until it gets into a network upgrade.
- [x] UI - Boost UI needs to visualize all direct data entries, and follow their sealing lifecycle, similarly to current Boost deals
  https://github.com/filecoin-project/boost/pull/1665
- [x] logs - we want to record the lifecycle of a direct data import, similarly to current Boost deals
- [x] allocate cmd - add cmdline args to specify "TermMin", "TermMax", "Expiration"
  https://github.com/filecoin-project/boost/pull/1668
- [x] CleanupData - make sure we cleanup original CAR if CleanupData is specified.


cmd/boostd/import_direct_data.go
```diff
@@ -0,0 +1,97 @@
...
+		},
+		&cli.StringFlag{
+			Name:     "client-addr",
+			Usage:    "",
+			Required: true,
+		},
+		&cli.Uint64Flag{
+			Name:     "allocation-id",
+			Usage:    "",
+			Required: true,
```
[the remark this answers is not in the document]
nonsense: Yes, that's correct, I just think for the MVP we are assuming it is required. This will certainly change for non-FIL+ deals though.

nonsense:
Marking this PR as ready for review, with the idea that some DDO code is commented out, until it gets support in Lotus and actors, and until we get clarity on how it affects the deal model.

Within the Boost team we've made the decision to merge into `main` in order to avoid conflicts in the future, and have an easier time adding full support, when DDO gets into the mainnet network.

cmd/boostd/main.go
```diff
@@ -46,6 +46,7 @@ func main() {
 			retrievalDealsCmd,
 			indexProvCmd,
 			importDataCmd,
+			importDirectDataCmd,
```
dirkmc, line 49: Remove?

dirkmc (approved):
LGTM 👍

ischasny:
Thanks Anton. It will take me some time to review this one - will try to wrap my head around it today / tomorrow!

node/impl/boost.go
```diff
@@ -526,6 +529,10 @@ func (sm *BoostAPI) BoostDagstoreDestroyShard(ctx context.Context, key string) e
 	return nil
 }
 
+func (sm *BoostAPI) BoostDirectDeal(ctx context.Context, params types.DirectDealParams) (*api.ProviderDealRejectionInfo, error) {
+	return sm.DirectDealsProvider.Import(ctx, params)
```
LexLuthr, line 533:
```suggestion
        return nil, fmt.Errorf("not implemented")
	// return sm.DirectDealsProvider.Import(ctx, params)
```

LexLuthr (approved):
Should we keep the APIs hidden as well?

Issue #1629, opened 2023-08-17
Add direct data onboarding client allocation implementation


LexLuthr:
Boost client should allow users to create allocation for verified deals. This functionality is independent of mode of the direct deal i.e. online or offline
The implementation will be based on https://github.com/filecoin-project/lotus/blob/00b8b1856cf8e72031b24952c9cfb89806038828/itests/verifreg_test.go#L426-L464

Expected functionality:
1. Allocate for one pieceCID (Data) to one minerID
2. Allocate for multiple pieceCIDs (Data) to one minerID
3. Allocate one pieceCID (Data) to multiple miners

Wait for message to land on chain and then print new allocations (allow a flag to print all allocations)

1. Add a command to query allocations

Expected commands
```
boost allocate [--wallet <walletID>] --miner <minerID1> <minerID2> --pieceCID <pieceCID1> <pieceCID2>
```

LexLuthr:
Additional Task: Allow for JSON output

jennijuju:
How is this blocked?

LexLuthr:
> why/How is this blocked?

We are waiting for the new branch on Lotus side to sync the dependency. The new branch will also have https://github.com/filecoin-project/lotus/pull/11183 so I can uncomment the verification code and use the gateway APIs instead of full node.

jennijuju:
https://filecoinproject.slack.com/archives/CP50PPW2X/p1692852646487879 - then you should be unblocked 

(also please proactively raise anything thats blocking you from lotus side to us, so we can help 🔪 !

PR #1632, opened 2023-08-18
feat: add `boost allocate` and `boost get-allocations` command

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1629

Currently, lotus gateway does not have the following APIs exposed. We need them to be enabled on gateway to allow Boost client to work with public gateways. 

```
StateGetAllocationForPendingDeal(ctx context.Context, dealId abi.DealID, tsk types.TipSetKey) (*verifregtypes.Allocation, error)
StateGetAllocation(ctx context.Context, clientAddr address.Address, allocationId verifregtypes.AllocationId, tsk types.TipSetKey) (*verifregtypes.Allocation, error)
StateGetAllocations(ctx context.Context, clientAddr address.Address, tsk types.TipSetKey) (map[verifregtypes.AllocationId]verifregtypes.Allocation, error)
StateGetClaim(ctx context.Context, providerAddr address.Address, claimId verifregtypes.ClaimId, tsk types.TipSetKey) (*verifregtypes.Claim, error)
StateGetClaims(ctx context.Context, providerAddr address.Address, tsk types.TipSetKey) (map[verifregtypes.ClaimId]verifregtypes.Claim, error)
```

https://github.com/filecoin-project/lotus/issues/11181 resolved by https://github.com/filecoin-project/lotus/pull/11183 to fix gateway APIs.

Currently, all the code pertaining to get the claims is commented out and would require to be uncommented in future once the APIs are available on the Lotus gateway.


cmd/boost/direct_deal.go
```diff
@@ -0,0 +1,440 @@
...
+		pieces := cctx.StringSlice("pieceInfo")
+		for _, p := range pieces {
+			pieceDetail := strings.Split(p, "=")
+			if len(pieceDetail) > 2 {
+				return fmt.Errorf("incorrect pieceInfo format: %s", pieceDetail)
+			}
+
+			n, err := strconv.ParseInt(pieceDetail[1], 10, 64)
+			if err != nil {
+				return fmt.Errorf("failed to parse the piece size for %s: %w", pieceDetail[0], err)
```
dirkmc, line 110:
```suggestion
				return fmt.Errorf("failed to parse the piece size for %s: %w", pieceDetail[1], err)
```

dirkmc (commented):
Looks great so far.

I wonder if we should simplify such that you can only make allocation requests for one miner at a time? Is there a use case where the user would want to make the same piece cid / size allocation against multiple miners at once?

cmd/boost/direct_deal.go
LexLuthr: I have clarified the error a little further. I was printing pieceCid on purpose. I have modified it to print both cid and size now.

LexLuthr:
@dirkmc The use case for multi-miner would be to allow replication. Let's say I have a piece and I want to store it with 6 different miners. I would not want to pay gas fee for 6 messages in that case.

dirkmc:
Ok seems like a good justification 👍

jennijuju:
> . I would not want to pay gas fee for 6 messages in that case.

Are we talking about the create allocation message here (datacap Transfer) or something else?

LexLuthr:
> > . I would not want to pay gas fee for 6 messages in that case.
> 
> Are we talking about the create allocation message here (datacap Transfer) or something else?

Allocation(datacap transfer)

nonsense (approved):

Issue #1621, opened 2023-08-14
mmlid: Boost Web UI should display storage and access miner IDs


dirkmc:
To help SPs switching between different boost instances set up with multiple retrieval miners:
- [x] The boost web UI should show the ID of the miner that it stores data to at the top right (next to the Epoch).
  https://github.com/filecoin-project/boost/pull/1633
- [x] On the LID page, boost web UI should show a list of the miners that it can retrieve from.

PR #1637, opened 2023-08-21
Show list of retrieval miners on retrievals list page

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/1621

Screenshot 2023-08-21 at 11 47 14 AM

### DIFF
--- a/gql/resolver_maddr.go
+++ b/gql/resolver_maddr.go
@@ -5,9 +5,10 @@ func (r *resolver) MinerAddress() string {
 }
 
 func (r *resolver) GraphsyncRetrievalMinerAddresses() []string {
-	addrs := r.cfg.Dealmaking.GraphsyncStorageAccessApiInfo
-	if len(addrs) == 0 {
-		addrs = []string{r.provider.Address.String()}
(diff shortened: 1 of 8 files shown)


gql/resolver_maddr.go
```diff
@@ -3,3 +3,11 @@ package gql
 func (r *resolver) MinerAddress() string {
 	return r.provider.Address.String()
 }
+
+func (r *resolver) GraphsyncRetrievalMinerAddresses() []string {
+	addrs := r.cfg.Dealmaking.GraphsyncStorageAccessApiInfo
```
nonsense, line 8: Isn't the `GraphsyncStorageAccessApiInfo` in the form of `xxx:/ip4/127.0.0.1/tcp/1234/http` and not a miner address?

nonsense (approved):

gql/resolver_maddr.go
dirkmc: Good catch 👍 
I moved the config param so my manual test wasn't doing what I thought it was.

nonsense (approved):

Issue #1640, opened 2023-08-22
mmlid: Show a warning on LID web UI while piece doctor performing initial scan


dirkmc:
When the user first migrates to boost v2.0, they will likely look at the LID page to see the new functionality.
The initial piece doctor scan takes 24 hours by default, so the LID page will show incomplete information.
We should show a warning to the user until the initial scan has completed:
> Piece Doctor is performing an initial scan for pieces in LID, expected completion time: 14 hours

PR #1643, opened 2023-08-23
Show warning on piece doctor page if still doing initial scan

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/1640

Screenshot 2023-08-23 at 2 54 15 PM

TODO:
- [x] Test in production

### DIFF
--- a/extern/boostd-data/client/client.go
+++ b/extern/boostd-data/client/client.go
@@ -27,6 +27,7 @@ type Store struct {
 		GetOffsetSize             func(context.Context, cid.Cid, mh.Multihash) (*model.OffsetSize, error)
 		ListPieces                func(ctx context.Context) ([]cid.Cid, error)
 		PiecesCount               func(ctx context.Context, maddr address.Address) (int, error)
+		ScanProgress              func(ctx context.Context, maddr address.Address) (*types.ScanProgress, error)
 		GetPieceMetadata          func(ctx context.Context, pieceCid cid.Cid) (model.Metadata, error)
 		GetPieceDeals             func(context.Context, cid.Cid) ([]model.DealInfo, error)
 		IndexedAt                 func(context.Context, cid.Cid) (time.Time, error)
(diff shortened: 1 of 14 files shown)


extern/boostd-data/client/client.go
```diff
@@ -27,6 +27,7 @@ type Store struct {
 		GetOffsetSize             func(context.Context, cid.Cid, mh.Multihash) (*model.OffsetSize, error)
 		ListPieces                func(ctx context.Context) ([]cid.Cid, error)
 		PiecesCount               func(ctx context.Context, maddr address.Address) (int, error)
+		ScanProgress              func(ctx context.Context, maddr address.Address) (*types.ScanProgress, error)
```
nonsense, line 30: The API for boostd-data has exploded quite a bit with auxiliary functionality.

I wonder if we can refactor in the future and split into 2 APIs -- one for critical functionality that one needs to implement if they want their own adapter for a new database (such as PostgreSQL with Citus), and one for auxiliary functionality that we build - Piece Doctor, UI feedback (such as ScanProgress), etc.

Just a thought as we keep on adding new methods here, no action required at this time.

extern/boostd-data/yugabyte/piecedoctor.go
```diff
@@ -235,6 +235,10 @@ func (s *Store) getPieceCheckPeriod(ctx context.Context) (time.Duration, error)
 	return period, nil
 }
 
+// Returns the number of rows in the PieceTracker table.
```
nonsense, line 238: Nit but for go-doc it is convention to start the comment with the name of the method (i.e. `PieceCount returns the number...`

If we are adding go-doc to some/all methods, we might as well stick to the convention.

extern/boostd-data/yugabyte/piecedoctor.go
```diff
@@ -249,6 +253,53 @@ func (s *Store) PiecesCount(ctx context.Context, maddr address.Address) (int, er
...
+	// Calculate approximate progress
+	progress := float64(scanned) / float64(total)
+
+	// Given that the denominator may be a little inflated, round up to 100% if we're close
+	if progress > 0.95 {
+		progress = 1
+	}
+	if progress > 1 {
+		progress = 1
+	}
```
nonsense, line 298:
```suggestion
```

nonsense (approved):

extern/boostd-data/client/client.go
dirkmc: That's true, we are adding a bunch of new methods 👍

extern/boostd-data/yugabyte/piecedoctor.go
dirkmc, line 238:
```suggestion
// PiecesCount returns the number of rows in the PieceTracker table.
```

Issue #1646, opened 2023-08-24
mmlid: Simple cassandra migration library


dirkmc:
As part of the solution to https://github.com/filecoin-project/boost/issues/1623 we will need to integrate a cassandra migration into boost.

We need to change the MinerAddr column from `t1234` to `f1234`, however there are no built-in string manipulation functions, and yugabyte does not support user defined functions. So we need a migration library that can run golang migration functions.

We use goose to do this for sqlite and postgres, but it turns out that:
- [goose](https://github.com/pressly/goose) doesn't support cassandra
- to add cassandra support is not straight-forward, because
- goose uses the `sql.DB` construct
  - there are no maintained cassandra drivers that implement sql.DB
  - there is [a sort of sql.DB implementation of a cassandra client](https://github.com/MichaelS11/go-cql-driver) but it doesn't support transactions, which are required by goose

Given all of the above, I think we should write a simple migration library ourselves for cassandra that
- uses hard-coded migration functions with a name and a migrate function
- only does up migrations
- does not have any CLI tools
- creates a table `migration_version` with a `name` columns to keep track of cassandra migrations
- when boost starts up it
  - checks `migration_version` against the array of migration functions
  - if there are any migration functions that have not been run it
    - runs the function
    - adds a row for each one to the `migration_version` table

PR #1648, opened 2023-08-25
Cassandra migrations

dirkmc:
Fixes https://github.com/filecoin-project/boost/issues/1646

TODO:
- [x] Test in production

### DIFF
--- a/extern/boostd-data/yugabyte/cassmigrate/migrate.go
+++ b/extern/boostd-data/yugabyte/cassmigrate/migrate.go
@@ -41,10 +41,10 @@ func Migrate(ctx context.Context, session *gocql.Session) error {
 	// Get a list of migrations that have already completed
 	qry = `SELECT Name from migration_version`
 	iter := session.Query(qry).WithContext(ctx).Iter()
-	var dbnames []string
-	var name string
-	for iter.Scan(&name) {
-		dbnames = append(dbnames, name)
+	var appliedMigrations []string
+	var appliedMigration string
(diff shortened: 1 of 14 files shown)


extern/boostd-data/yugabyte/cassmigrate/migrate.go
```diff
@@ -0,0 +1,111 @@
...
+	sort.Slice(migs, func(i, j int) bool {
+		return migs[i].name < migs[j].name
+	})
+
+	// For each migration
+	for i, m := range migs {
+		// Skip completed migrations
+		if i < len(dbnames) {
+			// Check the migrations order matches between the db and the executable
+			if m.name != dbnames[i] {
```
nonsense, line 77: `dbnames` is reference to `name` that are already in the `db`, but I think we could use a better name, such as `appliedMigrations`.

extern/boostd-data/yugabyte/cassmigrate/migrate.go
```diff
@@ -0,0 +1,111 @@
...
+
+var migrations []migration
+
+func addMigration(fn migrationFn) {
+	_, filename, _, _ := runtime.Caller(1)
+	migrations = append(migrations, migration{
+		name: path.Base(filename),
+		fn:   fn,
+	})
+}
```
nonsense, line 30: I think we can just assume the order in which we add the migrations to a slice to the order they should be applied, and just ignore all the sorting and getting of the filename.

PR #1647, opened 2023-08-25
chore: bump boost-graphsync to v0.13.7

LexLuthr:
This version of boost-graphsync contains the below 2 fixes.

Handle context cancellation properly (https://github.com/ipfs/go-graphsync/pull/428) by @hannahhoward in https://github.com/filecoin-project/boost-graphsync/pull/2
fix: update go-ipld-prime, don't slurp bytes on non-match visit by @rvagg in https://github.com/filecoin-project/boost-graphsync/pull/3

PR #1652, opened 2023-08-29
chore: add selector retrieval tests

rvagg:
On top of #1647

It'd be great to run these against booster-http too; we could just run an http client directly and request the CAR and do the same inspection.


testutil/car.go
```diff
@@ -62,6 +67,29 @@ func CreateRandomFile(dir string, rseed, size int) (string, error) {
 	return file.Name(), nil
 }
 
+func CreateRandomUnixfsFileInCar(t *testing.T, dir string, rseed, size int) (string, unixfs.DirEntry) {
```
rvagg, line 70: `testing.T` is a little out of place in this file, I hope this is OK. The utilities in github.com/ipfs/go-unixfsnode/testutil need one (although I'm relaxing that for an interface [here](https://github.com/ipfs/go-unixfsnode/pull/53)).

testutil/car.go
```diff
@@ -181,3 +209,20 @@ func WriteUnixfsDAGTo(path string, into ipldformat.DAGService, chunksize int64,
...
+	for {
+		next, err := blockReader.Next()
+		if err != nil {
+			break
+		}
+		gotCids = append(gotCids, next.Cid())
+	}
+	require.NoError(t, err)
+	return gotCids
+}
```
LexLuthr, line 228: I think we can remove the `testing.T` from this function and simply return the error from here. It would make for a more uniform util func.

testutil/car.go
LexLuthr: This should be okay.

itests/framework/framework.go
```diff
@@ -758,15 +762,17 @@ func (f *TestFramework) ExtractFileFromCAR(ctx context.Context, t *testing.T, fi
...
+func (f *TestFramework) RetrieveDirect(ctx context.Context, t *testing.T, root cid.Cid, pieceCid *cid.Cid, carExport bool, extractCar bool, selectorNode datamodel.Node) string {
 	offer, err := f.FullNode.ClientMinerQueryOffer(ctx, f.MinerAddr, root, pieceCid)
 	require.NoError(t, err)
 
-	return f.retrieve(ctx, t, offer, carExport)
+	return f.retrieve(ctx, t, offer, carExport, extractCar, selectorNode)
 }
 
-func (f *TestFramework) retrieve(ctx context.Context, t *testing.T, offer lapi.QueryOffer, carExport bool) string {
+func (f *TestFramework) retrieve(ctx context.Context, t *testing.T, offer lapi.QueryOffer, carExport bool, extractCar bool, selectorNode datamodel.Node) string {
```
LexLuthr, line 772: Both variables `carExport` and `extractCar` seems to be doing the same thing.

LexLuthr (changes requested):
Looks good. Just a couple of nits.

itests/framework/framework.go
rvagg: Sadly no, `carExport` gets sent along to the Lotus API to ask for a CAR (although I'm not sure what `IsCar: false` is going to do, perhaps a piece fetch?), while `extractCar` is used to determine whether to unpack the CAR into a unixfs file. I wish they could be a single boolean because I'm not happy extending this function signature even more.

The alternative might be to just expose a `RetrieveCar` method that handles this internally with `retrieve()`, that might make it a bit cleaner.

rvagg:
pushed a fixup (remember to squash merge this if I don't beforehand)

itests/framework/framework.go
rvagg: Actually .. it looks like this argument isn't actually used by anything else, perhaps this was originally copied out of lotus and we only ever ask for CAR exports.

So I've just removed that argument and replaced it with my new one that determines whether to extract the CAR or not; hard-wiring a `true` to the lotus call.

LexLuthr:
@rvagg it looks like we are missing the circleCI config to trigger this new itest. Can you add that as well?

LexLuthr (approved):

PR #1659, opened 2023-09-01
fix: use fixed car BlockReader#SkipNext SourceOffset

rvagg:
PR against `fix/commp` in #1651

Ref: https://github.com/ipld/go-car/pull/491


rvagg:
I've pulled in a cbor-gen update to go-car master which is showing up here with the cbor-check lint, so this also includes cbor-gen regeneration, which just contains a write optimisation: https://github.com/whyrusleeping/cbor-gen/pull/85

willscott (approved):
:+1: 

Merging into commp branch

PR #1660, opened 2023-09-01
fix: offsets to CAR section starts

rvagg:
these are required for carv2 indexes when we pass them back to a Blockstore


willscott (approved):
* the need for both `HeaderSize` and `PragmaSize` is unfortunate
* I'm unconvinced that the pain we force when boost using an un-aligned index is better than https://github.com/ipld/go-car/pull/490 but it's going to be a bit ugly either way.

PR #1664, opened 2023-09-01
Build containers for arm64 and amd64 platforms

masih:
Add multi platform build for container images for platforms:
* `linux/amd64`
* `linux/arm64`


masih:
sigh. this needs lotus to start publishing arm64 arch.

dirkmc:
@masih is it likely that lotus will add arm64 arch any time soon? Otherwise I'd say let's close this PR out for now.

PR #1680, opened 2023-09-07
feat: add InUnsealed utility to boostx

LexLuthr:
Tested on devnet

```
# boostx sector is-unsealed --help
NAME:
   boostx sector is-unsealed - boostx sector is-unsealed <Sector ID>

USAGE:
   boostx sector is-unsealed [command options] [arguments...]

DESCRIPTION:
   Check if a particular sector is unsealed using the miner API. This is a definitive method to check for unsealed copies

OPTIONS:
   --api-fullnode value  the endpoint for the full node API
   --api-storage value   the endpoint for the storage node API
   --help, -h            show help
```

For pre-sealed sector
```
# boostx sector is-unsealed --api-fullnode $FULLNODE_API_INFO --api-storage $MINER_API_INFO 0
Sector 0 is NOT unsealed
```

For a new deal sector
```
# boostx sector is-unsealed --api-fullnode $FULLNODE_API_INFO --api-storage $MINER_API_INFO 1
Sector 1 is unsealed
```

### DIFF
--- a/cmd/boostx/sector.go
+++ b/cmd/boostx/sector.go
@@ -133,11 +133,6 @@ var isUnsealedCmd = &cli.Command{
 		}
 		defer ncloser()
 
-		err = lib.CheckFullNodeApiVersion(ctx, fullnodeApi)
-		if err != nil {
-			return err
-		}
-
 		// Connect to the storage API and create a sector accessor
 		storageApiInfo := cctx.String("api-storage")
 		sa, storageCloser, err := lib.CreateSectorAccessor(ctx, storageApiInfo, fullnodeApi, log)
--- a/cmd/boostx/sector.go
+++ b/cmd/boostx/sector.go
@@ -2,12 +2,13 @@ package main
 
 import (
 	"fmt"
+	"strconv"
+	"time"
+
 	"github.com/filecoin-project/boost/cmd/lib"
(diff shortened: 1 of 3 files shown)


cmd/boostx/sector.go
```diff
@@ -92,3 +94,71 @@ var sectorUnsealCmd = &cli.Command{
...
+		if err != nil {
+			return fmt.Errorf("getting full node API: %w", err)
+		}
+		defer ncloser()
+
+		err = lib.CheckFullNodeApiVersion(ctx, fullnodeApi)
+		if err != nil {
+			return err
+		}
+
```
dirkmc, line 140: For simplicity I'd suggest we don't check the api version - probably this particular API is not going to change, and this way the SP doesn't have to worry about getting the version of the API in boostx to exactly match the API version in lotus-miner

dirkmc (approved):
Looks great! Nice work 🙌

cmd/boostx/sector.go
LexLuthr: Shall I remove it from https://github.com/filecoin-project/boost/blob/a0509fa9dade5ba84340f376bac79399b2a947a8/cmd/boostx/sector.go#L59 as well? The use case as same.

dirkmc: Sure sounds good 👍

PR #1684, opened 2023-09-08
add itest for direct data onboarding

dirkmc:
At the moment the test just checks that we can create a direct deal.
It doesn't wait for any checkpoints, because currently the actors bundle is not ready so lotus can't seal the data.


itests/ddo_test.go
```diff
@@ -0,0 +1,170 @@
...
+		StartEpoch:         35000,
+		EndEpoch:           518400,
+		FilePath:           carFilepath,
+		DeleteAfterImport:  false,
+		RemoveUnsealedCopy: false,
+		SkipIPNIAnnounce:   false,
+	}
+	rej, err := f.Boost.BoostDirectDeal(ctx, ddParams)
+	require.NoError(t, err)
+	if rej != nil && rej.Reason != "" {
```
nonsense, line 162: I think we should be able to check for `rej.Reason == "Accepted"` even though the deal is not successfully sealed in a sector at the moment.

itests/ddo_test.go
```diff
@@ -0,0 +1,170 @@
...
+		Method: verifreg.Methods.AddVerifier,
+		Params: params,
+		Value:  big.Zero(),
+	}
+
+	sm, err := f.FullNode.MpoolPushMessage(ctx, msg, nil)
+	require.NoError(t, err, "AddVerifier failed")
+	res, err := f.FullNode.StateWaitMsg(ctx, sm.Cid(), 1, lapi.LookbackNoLimit, true)
+	require.NoError(t, err)
+	require.EqualValues(t, 0, res.Receipt.ExitCode)
```
nonsense, line 82: I think the datacap setup should go under `framework`, as it is not really important for the DDO test, and we'll probably use it elsewhere in the future too.

itests/ddo_test.go
dirkmc: If `Reason` is "" that means it was accepted - the `Reason` field is only filled in if there's an error (it's used to tell the client the reason why the deal was rejected, eg not enough datacap)

itests/ddo_test.go
dirkmc: Makes sense, I will refactor

PR #1687, opened 2023-09-11
Watch sealing state until finalized

dirkmc:
Once the deal has been handed off to the sealing subsystem, watch the sealing state until it reaches a final state, then mark the deal as completed.

### DIFF
--- a/storagemarket/direct_deals_provider.go
+++ b/storagemarket/direct_deals_provider.go
@@ -13,6 +13,7 @@ import (
 	"github.com/filecoin-project/boost/api"
 	"github.com/filecoin-project/boost/db"
 	"github.com/filecoin-project/boost/storagemarket/logs"
+	"github.com/filecoin-project/boost/storagemarket/sealingpipeline"
 	"github.com/filecoin-project/boost/storagemarket/types"
 	smtypes "github.com/filecoin-project/boost/storagemarket/types"
 	"github.com/filecoin-project/boost/storagemarket/types/dealcheckpoints"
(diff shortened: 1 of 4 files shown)


storagemarket/direct_deals_provider.go
```diff
@@ -404,9 +407,73 @@ func (ddp *DirectDealsProvider) execDeal(ctx context.Context, entry *smtypes.Dir
...
+		// Get the sector status
+		si, err := ddp.sps.SectorsStatus(ddp.ctx, sectorNum, false)
+		if err == nil && si.State != lastSealingState {
+			// Sector status has changed
+			lastSealingState = si.State
+			ddp.dealLogger.Infow(dealUuid, "current sealing state", "state", si.State)
+		}
+
+		if IsFinalSealingState(si.State) {
+			if HasDeal(si.Deals, deal.ChainDealID) {
```
nonsense, line 438: We don't have `deal.ChainDealID` at all in DDO.

storagemarket/direct_deals_provider.go
```diff
@@ -404,9 +407,60 @@ func (ddp *DirectDealsProvider) execDeal(ctx context.Context, entry *smtypes.Dir
...
+	checkSealingFinalized := func() bool {
+		// Get the sector status
+		si, err := ddp.sps.SectorsStatus(ddp.ctx, sectorNum, false)
+		if err == nil && si.State != lastSealingState {
+			// Sector status has changed
+			lastSealingState = si.State
+			ddp.dealLogger.Infow(dealUuid, "current sealing state", "state", si.State)
+		}
+
+		return IsFinalSealingState(si.State)
```
nonsense, line 436: if `err != nil` I think this will panic, as `si` will be nil.

dirkmc: Thanks - still waking up here :)

PR #1696, opened 2023-09-12
feat: index and announce direct deals

LexLuthr:
This PR adds the code for indexing and announcing the direct deals.

- [x] Make a decision on what to use for onChainDealID in pieceStore (add an extra column for allocation ID?)
- [x] Update the code based on the above decision
- [x] Change unique context for direct deals to UUID
- [x] Update multihash iterator to use proposalCIDs as well as UUID

We will reuse the ChainDealID column to save the allocationIDs.
There will be an additional column called IsDirectDeal to distinguish what the ChainDealID column contains.


extern/boostd-data/model/model.go
```diff
@@ -31,6 +31,7 @@ type DealInfo struct {
 
 	// If we don't have CarLength, we have to iterate over all offsets, get
 	// the largest offset and sum it with length.
```
dirkmc, line 33: Let's update this comment

extern/boostd-data/yugabyte/cassmigrate/20230913144459_deals_add_is_direct.go
```diff
@@ -0,0 +1,48 @@
...
+			qry1 := `UPDATE PieceDeal SET IsDirectDeal = ? WHERE DealUuid = ?`
+			err1 := session.Query(qry1, dd, i).WithContext(ctx).Exec()
+			if err1 != nil {
+				return fmt.Errorf("setting deal %s miner address to %t: %w", i, dd, err1)
+			}
+			return nil
+		})
+	}
+
+	return eg.Wait()
```
dirkmc, line 47: I'm not sure that we need to set IsDirectDeal to false explicitly - IIUC the database with automatically set a new boolean column to false by default.

indexprovider/wrapper.go
```diff
@@ -528,18 +532,22 @@ func (w *Wrapper) IndexerAnnounceLatestHttp(ctx context.Context, announceUrls []
 }
 
 func (w *Wrapper) MultihashLister(ctx context.Context, prov peer.ID, contextID []byte) (provider.MultihashIterator, error) {
-	provideF := func(proposalCid cid.Cid, pieceCid cid.Cid) (provider.MultihashIterator, error) {
+	provideF := func(identifier string, isDD bool, pieceCid cid.Cid) (provider.MultihashIterator, error) {
+		idName := "propCid"
```
dirkmc, line 536: I think we can simplify by creating a new log variable, eg `idlog := log.With(idName, identifier)`

indexprovider/wrapper.go
```diff
@@ -551,69 +559,88 @@ func (w *Wrapper) MultihashLister(ctx context.Context, prov peer.ID, contextID [
...
+			e := fmt.Errorf("getting deal with UUID %s from direct deal database: %w", dealUUID, dderr)
+			log.Infow("pausing ingestion", "deal UUID", dealUUID, "err", e)
+			return nil, e
+		}
 	}
 
-	// The deal was not found in the boost or legacy database.
-	// Skip this deal and continue ingestion.
-	err = fmt.Errorf("deal with proposal cid %s not found", proposalCid)
-	log.Infow("skipping ingestion", "proposalCid", proposalCid, "err", err)
```
dirkmc, line 616: I think we still want to preserve this "not found" error message in the case that the context ID is a uuid but it's not found in the boost DB

indexprovider/wrapper.go
```diff
@@ -684,3 +711,38 @@ type basicDealInfo struct {
...
 	DealProposal   storagemarket.ClientDealProposal
 }
+
+func (w *Wrapper) AnnounceBoostDirectDeal(ctx context.Context, entry *types.DirectDeal) (cid.Cid, error) {
+	// Filter out deals that should not be announced
+	if !entry.AnnounceToIPNI {
+		return cid.Undef, nil
+	}
+
+	contextID := []byte(entry.ID.String())
```
dirkmc, line 721: Does this work? As we're parsing the context ID with uuid.FromBytes, I would expect that we would need to product the context ID with MarshallBinary()

indexprovider/wrapper.go
```diff
@@ -684,3 +711,38 @@ type basicDealInfo struct {
...
+	}
+
+	// Ensure we have a connection with the full node host so that the index provider gossip sub announcements make their
+	// way to the filecoin bootstrapper network
+	if err := w.meshCreator.Connect(ctx); err != nil {
+		log.Errorw("failed to connect boost node to full daemon node", "err", err)
+	}
+
+	// Announce deal removal to network Indexer
+	annCid, err := w.prov.NotifyRemove(ctx, "", []byte(dealUUID.String()))
```
dirkmc, line 743: Same question about MarshallBinary here

storagemarket/direct_deals_provider.go
```diff
@@ -71,8 +77,10 @@ func NewDirectDealsProvider(cfg DDPConfig, fullnodeApi v1api.FullNode, pieceAdde
 		//logsSqlDB: logsSqlDB,
 		//logsDB: logsDB,
 
-		dealLogger: dealLogger,
-		running:    make(map[uuid.UUID]struct{}),
+		dealLogger:     dealLogger,
+		running:        make(map[uuid.UUID]struct{}),
+		piecedirectory: piecedirectory,
```
dirkmc, line 82: nit: let's rename the piecedirectory parameter so that it doesn't override the package name

storagemarket/direct_deals_provider.go
```diff
@@ -566,3 +586,50 @@ func (ddp *DirectDealsProvider) FailPausedDeal(ctx context.Context, id uuid.UUID
...
+			annCid, err := ddp.ip.AnnounceBoostDirectDeal(ctx, entry)
+			if err != nil {
+				return &dealMakingError{
+					retry: types.DealRetryAuto,
+					error: fmt.Errorf("failed to announce deal to network indexer: %w", err),
+				}
+			}
+			ddp.dealLogger.Infow(entry.ID, "announced the direct deal to network indexer", "announcement-cid", annCid)
+		} else {
+			ddp.dealLogger.Infow(entry.ID, "didn't announce the direct deal as requested in the deal proposal")
```
dirkmc, line 624:
```suggestion
			ddp.dealLogger.Infow(entry.ID, "didn't announce the direct deal as requested in the deal proposal because announcements are disabled")
```

storagemarket/types/direct_data.go
```diff
@@ -32,6 +32,7 @@ type DirectDeal struct {
 
 	// InboundCARPath is the file-path where the storage provider will persist the CAR file sent by the client.
 	InboundFilePath string
+	InboundFileSize int64
```
dirkmc, line 35: Is InboundFileSize being stored to the db? If so we will need to add it to the db code also

extern/boostd-data/yugabyte/cassmigrate/20230913144459_deals_add_is_direct.go
LexLuthr: I don't think Cassandra does default values. It will set it to NULL as everything is key:value pair in the lower layers. I will  check with Yugabyte team about this.

storagemarket/direct_deals_provider.go
LexLuthr: There is a separate line for when announcements are disabled as index-provider is disabled. This case is for when client requests not to announce

extern/boostd-data/yugabyte/cassmigrate/20230913144459_deals_add_is_direct.go
dirkmc: hmm yeah I think you're right, it will set the column to null.
If that's the case then I think we can still simplify this a bit - we don't need to update all rows one by one, we can just do a single update for all rows:
`UPDATE PieceDeal SET IsDirectDeal = false`

storagemarket/direct_deals_provider.go
dirkmc: Ah ok thanks, I got confused by the language.
Let's change it to
```didn't announce the direct deal because the client disabled announcements for this deal```

File: extern/boostd-data/yugabyte/cassmigrate/20230913144459_deals_add_is_direct.go (hunk 2 above)
Comment (LexLuthr) [reply extern/boostd-data/yugabyte/cassmigrate/20230913144459_deals_add_is_direct.go:47]: Cassandra doesn't allow setting a column value for all rows. I tried different approaches but none of them worked. 

```
  (ql error -12)
go-tests_1  |     svc_test_util.go:348: 
go-tests_1  |           Error Trace:    /go/src/svc/svc_test_util.go:348
go-tests_1  |                                                   /go/src/svc/svc_yugabyte_test.go:52
go-tests_1  |           Error:          Received unexpected error:
go-tests_1  |                           starting local index directory service: running cassandra migrations: running migration ts20230913144459_dealsAddIsDirectDealColumn: setting IsDirectDeal to false: Invalid CQL Statement. Missing partition key
go-tests_1  |                           UPDATE PieceDeal SET IsDirectDeal = FALSE
go-tests_1  |                           ^^^^^^
go-tests_1  |                            (ql error -12)
go-tests_1  |           Test:           TestCleanupYugabyte
```

dirkmc: Interesting, I guess that makes sense as it's not really a relational db.
Ok in that case I'll resolve the comment.

dirkmc (approved):

PR #1697, opened 2023-09-13
fix: update recovery for non-car files

### LINKED ISSUE #1693: nocar: Update LID recovery tool to account for non-car files
This would need be implemented along with changes required to recover without onChain details about the deals

LexLuthr:
Fixes #1693 
This PR uses UnsealSectorAt() to get a reader and directly feeds the reader to the commP writer. This is the same mechanism as getting a full piece from `booster-HTTP` and calculating commP on it.

### DIFF
--- a/cmd/boostd/recover.go
+++ b/cmd/boostd/recover.go
@@ -574,10 +574,10 @@ func processPiece(ctx context.Context, sectorid abi.SectorNumber, chainDealID ab
 	if err != nil {
 		return err
 	}
+	defer reader.Close()
 	if !isUnsealed {
 		return fmt.Errorf("sector %d is not unsealed", sid)
 	}
-	defer reader.Close()
 
 	dr.Sectors[sid].Deals[cdi].IsUnsealed = true
 	dr.Sectors[sid].Deals[cdi].GotDataReader = true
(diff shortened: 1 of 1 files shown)


cmd/boostd/recover.go
```diff
@@ -670,14 +661,13 @@ func processPiece(ctx context.Context, sectorid abi.SectorNumber, chainDealID ab
...
 
-		//fmt.Println("CommP CID: ", encoder.Encode(commp.PieceCID))
-		//fmt.Println("Piece size: ", types.NewInt(uint64(commp.PieceSize.Unpadded().Padded())))
-
 		if !commp.PieceCID.Equals(piececid) {
 			return fmt.Errorf("calculated commp doesnt match on-chain data, expected %s, got %s", piececid, commp.PieceCID)
 		}
 	}
 
+	_ = reader.Close()
```
dirkmc, line 669:
```suggestion
```

cmd/boostd/recover.go
```diff
@@ -580,15 +580,6 @@ func processPiece(ctx context.Context, sectorid abi.SectorNumber, chainDealID ab
```
dirkmc, line 580: I'd suggest putting `defer reader.Close()` right after the error check above

cmd/boostd/recover.go
```diff
@@ -577,22 +577,9 @@ func processPiece(ctx context.Context, sectorid abi.SectorNumber, chainDealID ab
 	if !isUnsealed {
 		return fmt.Errorf("sector %d is not unsealed", sid)
 	}
+	defer reader.Close()
```
dirkmc, line 580: The defer should go before the isUnsealed check above

dirkmc (approved):

PR #1704, opened 2023-09-20
chore(deps): upgrade graphsync for identity CID fix

rvagg:


rvagg:
test failing because the test itself uses an identity CID for "missing" .. working on some additional tests and possible changes to handle identity CIDs through a retrieval cycle

rvagg:
OK, this is a bit of a problem and I'm going to have to take an alternative approach with graphsync because of classic online deal flow and it's reliance on identity CIDs being non-simulated blocks _everywhere_.

LexLuthr (commented):
The new itest is still not in circleCI.

LexLuthr (approved):

LexLuthr (approved):
LGTM!

Issue #1593, opened 2023-07-26
Multi Stream http download for online deals


LexLuthr:
Current Status:
Boost current uses a single stream to download the data. Most of the clients have some kind of bandwidth control on their side to balance the load. This makes a single stream download slower.

Proposal:
1. Limits: Default streams - 5
                Maximum Stream - 16
                
2. User can configure the number of streams allowed per deal download. This number must be between 1 and 16.
3. We download car file for each deal with the configured number of stream. **The total number of deals to deal download in parallel with stay same and would not be in the scope of this feature.**
4. To break the download into chunks, we request the content-Length and divide is by (number of streams -1) and remainder is downloaded by the last stream.

Example: With 10 stream
          Content Length % (10 -1) = Stream 10
          Content Length / (10 -1) = Stream 1 to Stream 9 equally 

5. The total required space should not exceed the car file size
6. Backoff/Retry should work for each stream. Client might have a cap on number of parallel connections from same IP.
7. Resume should work for each stream instead of each deal. This will help reduce the bandwidth and data consumption on SP side.
8. File writing should be handler by a single atomic writer.
9. Each stream will have it's own writer which will keep track of all the range offset and current offset. This writer will in turn call the other atomic writer to write to the disk using os.WriteAt().
10. If writing to the disk is slow, we will need to figure out a way to write to the disk concurrently. Probably a multiple independent writers will work if we use a single *File as long as we can guarantee no overwriting. 

The aria2c utility used by most of our SPs has a similar working process. It allows users to queue multiple downloads. From the queue, it picks up the configured number of downloads to process. For each download it will use configured number of streams to download the data to a single file.
Aria2c utilises a control file resume downloads. It is few KBs for each download and is removed once download complete. We can probably use a similar approach or use DB.

PR #1710, opened 2023-09-22
feat: add multistream http download

ischasny:
* Add a feature to split http download into multiple chunks. Chunks are downloaded into temporary files, which then get merged into one;
* Downnloads over libp2p are always done in one chunk only;
* Add NChunks configuration parameter;
* Fix tests to handle not just open-ended range requests;
* Don't allow private ip addresses and redirects.

### DIFF
--- a/transport/httptransport/http_downloader.go
+++ b/transport/httptransport/http_downloader.go
@@ -62,7 +62,7 @@ func (d *downloader) doHttp() error {
 	// send http request and validate response
 	resp, err := d.transfer.client.Do(req)
 	if err != nil {
-		return &httpError{error: fmt.Errorf("failed to send  http req: %w", err)}
+		return &httpError{error: fmt.Errorf("failed to send http req: %w", err)}
 	}
 	// we should either get back a 200 or a 206 -> anything else means something has gone wrong and we return an error.
 	defer resp.Body.Close() // nolint
--- a/transport/httptransport/http_downloader.go
+++ b/transport/httptransport/http_downloader.go
@@ -7,9 +7,6 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"sync/atomic"
-
-	"github.com/filecoin-project/boost/transport/types"
 )
 
 type downloader struct {
@@ -20,6 +17,7 @@ type downloader struct {
 	rangeStart int64
 	rangeEnd   int64
 	chunkNo    int
+	dealSize   int64
 }
 
 func (d *downloader) doHttp() error {
@@ -37,7 +35,7 @@ func (d *downloader) doHttp() error {
 	// calculate range boundaries taking into account bytes that have been already downloaded
 	fi, err := os.Stat(d.chunkFile)
 	if err != nil && !errors.Is(err, os.ErrNotExist) {
-		return &httpError{error: fmt.Errorf("failed to read file size: %w", err)}
+		return &httpError{error: fmt.Errorf("failed to read size of chunk file %s: %w", d.chunkFile, err)}
 	}
 	var offset int64
 	if fi != nil {
@@ -79,7 +77,7 @@ func (d *downloader) doHttp() error {
 	// create chunk file if it doesn't exist
 	of, err := os.OpenFile(d.chunkFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
 	if err != nil {
-		return &httpError{error: fmt.Errorf("failed to open chunk file for write: %w", err)}
+		return &httpError{error: fmt.Errorf("failed to open chunk file %s for write: %w", d.chunkFile, err)}
 	}
 	defer of.Close()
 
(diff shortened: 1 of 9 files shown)


dirkmc (commented):
Looks great so far 👍 

If I'm reading the code correctly, it will split the file into chunks (5 by default), and make a request for the chunks in parallel. However it will only start reading the response stream for each chunk once the previous chunk has completed. So I'm not sure this will end up being faster than a single stream download.

transport/httptransport/http_transport.go
```diff
@@ -120,6 +140,12 @@ func (h *httpTransport) Execute(ctx context.Context, transportInfo []byte, dealI
 	}
 	h.dl.Infow(duuid, "existing file size", "file size", fileSize, "deal size", dealInfo.DealSize)
 
+	// default to a single stream for libp2p urls as libp2p server doesn't support range requests
```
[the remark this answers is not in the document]
ischasny: yes, looks like it doesn't

transport/httptransport/http_transport.go
```diff
@@ -252,7 +267,75 @@ func (t *transfer) execute(ctx context.Context) error {
 
 		// start the http transfer
 		remaining := t.dealInfo.DealSize - t.nBytesReceived
-		reqErr := t.doHttp(ctx, req, of, remaining)
+
+		// split download into chunks
+		// each chunk should be bigger than minChunkBytes
+		nChunks := int(remaining/minChunkBytes) + 1
```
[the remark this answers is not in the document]
ischasny: Good spot - fixed.

transport/httptransport/http_transport.go
```diff
@@ -50,6 +58,15 @@ func BackOffRetryOpt(minBackoff, maxBackoff time.Duration, factor, maxReconnectA
 	}
 }
 
+func NChunksOpt(nChunks int) Option {
+	return func(h *httpTransport) {
+		if nChunks < 1 || nChunks > 16 {
+			return
+		}
```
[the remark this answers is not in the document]
ischasny: Fixed

transport/httptransport/http_transport.go
```diff
@@ -216,53 +234,114 @@ type transfer struct {
...
 func (t *transfer) execute(ctx context.Context) error {
 	duuid := t.dealInfo.DealUuid
-	for {
-		// construct request
-		req, err := http.NewRequest("GET", t.tInfo.URL, nil)
-		if err != nil {
-			return fmt.Errorf("failed to create http req: %w", err)
+
+	// create downloaders. Each downloader must be initialised with the same byte range across restarts in order to resume previous downloads.
+	// if the output file contains some data in it already, do not create a downloader for it again
```
dirkmc, line 245: > if the output file contains some data in it already, do not create a downloader for it again

I'm not sure I understand this comment.. are you saying that if some chunks have already downloaded, don't download them again?

transport/httptransport/http_downloader.go
```diff
@@ -0,0 +1,207 @@
...
+
+			// if the number of read and written bytes don't match -> something has gone wrong, abort the http req.
+			if nw < 0 || nr != nw {
+				if writeErr != nil {
+					return &httpError{error: fmt.Errorf("failed to the chunk file: %w", writeErr)}
+				}
+				return &httpError{error: fmt.Errorf("read-write mismatch writing to the chunk file, read=%d, written=%d", nr, nw)}
+			}
+			chunkBytesReceived += int64(nw)
+			totalBytesReceived := atomic.AddInt64(&d.transfer.nBytesReceived, int64(nw))
```
dirkmc, line 109: We should check that wherever transfer.nBytesReceived is being accessed, it's accessed with atomic.LoadInt64().
Maybe it would make sense to add a method to transfer (eg `addBytesReceived(delta int64)`) that
- updates the nBytesReceived (with an exclusive lock)
- emits an event with the new total bytes received

dirkmc (commented):
A few suggestions, but overall it looks great, nice work! 🙌

transport/httptransport/http_transport.go
```diff
@@ -216,53 +234,114 @@ type transfer struct {
...
+	outputStats, err := os.Stat(t.dealInfo.OutputFile)
+	if err != nil {
+		return &httpError{error: fmt.Errorf("failed to get stats of the output file: %w", err)}
+	}
+
+	chunkSize := t.dealInfo.DealSize / int64(t.nChunks)
```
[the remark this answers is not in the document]
ischasny: Good point - adding

transport/httptransport/http_transport.go
ischasny: There are a few possible scenarios there: 
1. some chunks might have been fully downloaded, while others just partially. We need to do nothing for the former and finish downloading for the latter. 
2. all chunks have been downloaded. Only some chunks have been merged into the main file and their temp files have been deleted. We need to make sure to not to re-download the merged chunks again. 

Added clarification to the code.

transport/httptransport/http_downloader.go
ischasny: Good point - added

transport/httptransport/http_transport.go
```diff
@@ -119,6 +121,16 @@ func (h *httpTransport) Execute(ctx context.Context, transportInfo []byte, dealI
 	}
 	tInfo.URL = u.Url
 
+	// don't allow private network addresses as per https://en.wikipedia.org/wiki/Private_network
```
nonsense, line 124: I think we want this to be configurable. We don't really know what kind of setups SPs have and what software they've integrated Boost with... I wouldn't be surprised if some deals are made with private IPs for fetching data.

ischasny:
Splitting this PR into multiple small ones. The first one: https://github.com/filecoin-project/boost/pull/1723

Issue #1716, opened 2023-09-26
LID web UI: don't look up cid by root payload cid


dirkmc:
In the LID web UI, the SP can look up a cid in the search box.

It will look up the cid as a payload cid by searching
1. in the LID payload cid -> piece cid index
2. in the boostd sqlite database by root payload cid (deal data root cid).

In practice it is confusing to lookup the cid by root payload cid (the deal's Label field) in the boostd sqlite database. It can give the SP the impression that the deal label field corresponds to a payload cid, when that may not actually be the case.

If the user wants to search by the deal's Label field they can already do this using the search box on the Storage Deals list page.

We should remove the code that looks up the cid by root payload cid in the web UI, including removing it from the graphql endpoint (resolver).

PR #1721, opened 2023-09-26
fix: remove payloadRootCid lookup

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1716

### DIFF
--- a/react/src/LID.js
+++ b/react/src/LID.js
@@ -3,7 +3,7 @@ import {useMutation, useQuery} from "@apollo/react-hooks";
 import {
     LIDQuery,
     FlaggedPiecesQuery, PieceBuildIndexMutation,
-    PieceStatusQuery, PiecesWithPayloadCidQuery, PiecesWithRootPayloadCidQuery, FlaggedPiecesCountQuery
+    PieceStatusQuery, PiecesWithPayloadCidQuery, FlaggedPiecesCountQuery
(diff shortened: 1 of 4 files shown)


react/src/LID.js
```diff
@@ -510,23 +510,12 @@ function SearchResults({searchQuery, setSearchQuery, showSearchPrompt}) {
...
     var pieceCid = null
     var pieceCids = []
-    if ((payloadRes || {}).data && (rootPayloadRes || {}).data) {
+    if ((payloadRes || {}).data) {
         pieceCids = [...new Set([
-            ...payloadRes.data.piecesWithPayloadCid,
-            ...rootPayloadRes.data.piecesWithRootPayloadCid
-        ])]
+            ...payloadRes.data.piecesWithPayloadCid])]
         if (pieceCids.length === 0) {
```
dirkmc, line 519: We don't need the Set any more so we can replace the above 3 lines with
```
if (payloadRes.data.piecesWithPayloadCid.length === 0) {
```

dirkmc: And the `if` below with
```
} else if (payloadRes.data.piecesWithPayloadCid.length === 1) {
```

dirkmc (approved):

dirkmc:
Love these before and after screenshots, nice work! 👐

PR #1723, opened 2023-09-28
Ivan/feat multistream download

ischasny:
* Add a feature to split http download into multiple chunks. Chunks are downloaded into temporary files, which then get merged into one;
* Downloads over libp2p are always done in one chunk only;
* Add NChunks configuration parameter;
* Fix tests to handle not just open-ended range requests;


dirkmc (approved):
Looks great, nice work! 🙌

PR #1724, opened 2023-09-28
Ivan/http security fix

ischasny:
* Don't allow download from private IP addresses by default. Add an option to drive that behaviour. 
* Don't follow http redirects.

### DIFF
--- a/transport/httptransport/http_transport.go
+++ b/transport/httptransport/http_transport.go
@@ -61,7 +61,7 @@ func NChunksOpt(nChunks int) Option {
 	}
 }
 
-func AllowProvateIPsOpt(b bool) Option {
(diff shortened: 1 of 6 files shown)


transport/httptransport/http_transport.go
```diff
@@ -59,6 +61,12 @@ func NChunksOpt(nChunks int) Option {
 	}
 }
 
+func AllowProvateIPsOpt(b bool) Option {
```
nonsense, line 64:
```suggestion
func AllowPrivateIPsOpt(b bool) Option {
```

dirkmc (approved):
LGTM 👍 
Nice work!

PR #1732, opened 2023-10-03
Ivan/http transport upgrades

ischasny:
This branch has a series of reviewed PRs merged into it. The code introduces the following changes:

* **Feature**: Add `NChunks` property that defines the number of concurrent streams for each http download with the minimum of 1, maximum of 15 and the default of 5. That will speed up how fast car files get downloaded up to the factor of the `NChunks` value. Libp2p over http is always defaulted to 1 stream only;
* **Feature**: Http downloads persist parameters that they have been started with into the control file. That will allow boost to consistently resume downloads even if the configuration changes between node restarts;
* **Test**: Add performance test that simulates network latency and ensures that multi-stream http downloads are reasonably faster than single-stream;

Multistream downloads have been tested on the main net. With `NChunks=5` the overall download speed averaged at 448MBps, while with `NChunks=1` at 169MBps (approx 2.6x improvement). It's expected that the speed-up should be proportional to the `NChunks` value if enough bandwidth / IOPs is available as the chunks are downloaded concurrently without blocking each other. 

After download is complete chunks need to be glued up into the main output file. In our testing that process took approx. 4 minutes to complete for a 32GB car file on HDD. We expect that to be much faster for SSDs. However, even with 4 minutes of extra overhead the performance boost of parallelising HTTP download is still significant. 

### DIFF
--- a/transport/httptransport/http_transport_perf_test.go
+++ b/transport/httptransport/http_transport_perf_test.go
@@ -1,0 +1,117 @@
+package httptransport
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/filecoin-project/boost/transport/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHttpTransportMultistreamPerformance(t *testing.T) {
+	// this test runs a tcp proxy in front of the http web server that adds a latency to each tcp packet.
+	// This is done to simulate network latency and ensure that multistream transfer is reasonably faster than singlestream.
+	// rawSize and latency need to be big enough so that data transfer takes a reasonable proportion of time comparing to writing to the disk.
+	ctx := context.Background()
+	latency := 100 * time.Millisecond
+	rawSize := 100 * readBufferSize // 100 Mib
+	st := newServerTest(t, rawSize)
+	carSize := len(st.carBytes)
+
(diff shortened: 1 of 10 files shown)


transport/httptransport/http_transport_perf_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/filecoin-project/boost/transport/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHttpTransportMultistreamPerformance(t *testing.T) {
+	// this test runs a tcp proxy in front of the http web server that adds a latency to each tcp packet.
+	// This is done to simulate network latency and ensure that multistream transfer is reasonably faster than singlestream.
+	// rawSize and latency need to be big enough so that data transfer takes a reasonable proportion of time comparing to writing to the disk.
+	ctx := context.Background()
```
dirkmc, line 24: I'd suggest we add t.Skip() at the top of this test as we will only use it occasionally.

dirkmc (approved):
LGTM 👍 

Nice work! 🙌

transport/httptransport/http_transport_perf_test.go
ischasny: My idea was to have that test running for each build so that we know that nothing occasionally broke concurrent transfers. For example if the code gets changed to download chunks serially, then all existing tests would just pass while the functionality would be broken. WDYT?

dirkmc: Ok seems fair, let's leave it as-is 👍

nonsense (approved):

LexLuthr (approved):

PR #1676, opened 2023-09-07
booster-http refactor - trustless only, improved compression, logging & e2e testing w/ trusted

rvagg:
This is a bit of a PoC at the moment; enables a mini Trustless Gateway implementation when you only want to serve CARs through `/ipfs/`. Uses Frisbii but I've copied in the two main modules that are used into this PR because of the libp2p dependency pain that gets stuck at ipni dependencies in Frisbii which are not required here. So cmd/booster-http/frisbii/* are (mostly) unmodified copies from github.com/ipld/frisbii - we only need the handler and the CAR streamer for this.

The transparent switch in implementations by subtle CLI argument changes is probably not optimal for users, but I've got a big integration test in here that pulls in the big [trustless pathing test fixture](https://github.com/ipld/ipld/blob/master/specs/transport/trustless-pathing/fixtures/unixfs_20m_variety.md) that exercises a whole lot of trustless path retrievals for a fixed 20Mb CAR (which we import across multiple deals here) and it gets run across both implementations in here with the same boost backend, just two different booster-http frontends.

The reason this _might_ be useful are that (1) it's super minimal and very lightweight; there's no intermediate caching of blocks while it builds up state, it just does a traversal straight from the backend blockstore and spits it out as a CAR, (2) it gives us a bit more freedom to experiment with alternative outputs, like what SPARK is after, and (3) gives us useful error output if there's a problem loading the root block, instead of the current empty CAR that we get.

- [x] Update docs

PR #1736, opened 2023-10-09
feat(booster-http): --compression-level flag for piece & frisbii, fix GzipHandler usage

rvagg:
This is a stacked PR on top of #1676; one of a couple I'm going to do to consolidate and address some things with the two primary HTTP handler paths.

This one does a few things:

* New `--compression-level` flag with a default of `1`. Allows users to turn it off if they're going to do this in a reverse proxy (recommended, really).
* Plumb through the same level to Frisbii as well as the piece handler
* Replace the direct usage of `gziphandler.GzipResponseWriter` in the piece HandlerFunc with a `gziphandler.MustNewGzipLevelHandler()` as a Handler, wrapping the HandlerFunc - this does two things: (1) lets us set a compression level, the default we're getting now is a basic deflate (`1`), and (2) lets us use the gzip pool that gziphandler manages at the level of Handler.
* Add a couple of extra headers for the piece handler, including a `Cache-Control`.
* A bunch of tests to make sure it's working - the existing test just sets a header and adds a gzip reader, which is already done internally (transparently) by the Go HTTP client; so the tests added here to a variety of on and off for compression.

I've set the default to `1` but I'd rather set it to `6` for better results, with a small amount of CPU impact. I chose `1` because that's all users are getting now.


willscott (approved):

rvagg:
test failing because of a minor compatibility problem with frisbii@0.3.0, going to cut a new version soon with some other updates for my next PR so I'll roll that in here and the test will turn green

rvagg:
released frisbii@v0.4.0, updated to that in #1676 and that should address the CI problems here

masih (approved):
I will review the final PR against main more closely. Based on a quick look this looks fine 👍

rvagg:
I've pulled this in to #1676, it car be reviewed there as part of the mega-changeset but is still a separate commit that can be viewed separately if desired.

PR #1741, opened 2023-10-09
Fix boostd-data go module path and remove replace directives

masih:
Use go workspaces to remove replace directives in root boost go module.

Fix the path to `boostd-data` go module to be a valid one.


ischasny (approved):
Looks good - can you please provide the context why the path change was required?

nonsense (approved):

masih:
Don't delete this branch until boostd data dependency points to a commit on main branch.

PR #1743, opened 2023-10-10
Fix static check issues

masih:
Address static check issues:
 * remove redundant import
 * lower case error string
 * Use `fmt.Sprintf` to construct string value instead of `fmt.Errorf.Error()`


LexLuthr (approved):

ischasny (approved):

nonsense (approved):

nonsense:
No need to add 3 people to review chores like this, I think one reviewer is enough for such PRs.

PR #1745, opened 2023-10-10
fix LID web UI

nonsense:
Fixed division by zero at two places, which can happen if Boost is installed on a new SP that has 0 deals.

Also small fix for debug level logging from boostd-data service.

We probably need an RC, such as Boost v2.1.1 after this is merged, @LexLuthr, WDYT?

TODO:
- [x] make sure LID Web UI is displayed correctly with 0 deals
- [x] make sure that boostd-data second process is used by boostd, and not an internal leveldb service for LID


LexLuthr:
UI if fixes now
Screenshot 2023-10-10 at 5 12 46 PM

LexLuthr:
I think we should release RC2 with this.

Issue #1733, opened 2023-10-03
New incoming offline deals are blocked when boost receives 200+ online deals.


cryptowhizzard:
### Checklist

- [X] This is **not** a question or a support request. If you have any boost related questions, please ask in the [discussion forum](https://github.com/filecoin-project/boost/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/boost/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/boost/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/boost/issues) and the [discussion forum](https://github.com/filecoin-project/boost/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/boost/releases), or the most recent RC(release canadiate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to boost.

### Boost component

- [X] boost daemon - storage providers
- [ ] boost client
- [ ] boost UI
- [X] boost data-transfer
- [ ] boost index-provider
- [ ] Other

### Boost Version

```text
Latest master october 2 2023
```

### Describe the Bug

When boost receives a lot of online deals ( 200+ ) the channel for making new offline deals seems to get blocked.

Both client and SP receive error:
Error: send proposal rpc: failed to read response: stream reset

I verified the numbers for storage in the config. They are all above 1000 ( lotus-miner as well as boost ).

### Logging Information

```text
Error: send proposal rpc: failed to read response: stream reset
```

### Repo Steps

1. Run '...'
2. Do '...'
3. See error '...'
...

cryptowhizzard:
This PR 1746 seems to improve a bit but the problem resides in provider_dealfilter.go.

Boost needs to output a lot of things about sectors and statuses to the external filter ( if configured ) and this is where the delay originates. Disabling everything except the deal parameters like client / provider / cid seems to resolve the problem in combination with this PR.

In my situation i am running a 30 PB+ machine and boost needs to collect a lot of info to send to that filter causing massive delays.

[...]

PR #1746, opened 2023-10-11
feat: run deal filters in parallel

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1733
This PR moves processesOnlineDeals and processOfflineDeals to their own goroutines.
Possible side effects:
1. Multiple CMD execution of custom filters in parallel (High CPU utilisation depending on the number of deal being processed)
2. The deal rejection will not be in order anymore. Say, we get deal A first and deal B later and start processing them. It is possible for deal A to be rejected for lack of resources and deal B to be accepted depending upon which goroutine finishes first. 

DO NOT MERGE

### DIFF
--- a/storagemarket/provider_loop.go
+++ b/storagemarket/provider_loop.go
@@ -278,7 +278,6 @@ func (p *Provider) processDeal(deal *types.ProviderDealState, rsp chan acceptDea
 	case p.processedDealChan <- processedDealReq{deal: deal, err: err, rsp: rsp}:
 	case <-p.ctx.Done():
 	}
-	return
 }
 
 func (p *Provider) processImportOfflineDealData(deal *types.ProviderDealState) *acceptError {
--- a/storagemarket/provider_loop.go
+++ b/storagemarket/provider_loop.go
@@ -406,6 +406,7 @@ func (p *Provider) run() {
 
 				// send an accept response
 				dealReq.rsp <- acceptDealResp{ri: &api.ProviderDealRejectionInfo{Accepted: true}}
+				continue
 			}
 
 			// Send new online and offline deals for processing.
--- a/storagemarket/provider_loop.go
+++ b/storagemarket/provider_loop.go
@@ -393,15 +393,6 @@ func (p *Provider) run() {
 			deal := dealReq.deal
 			p.dealLogger.Infow(deal.DealUuid, "processing deal acceptance request")
 
-			if deal.IsOffline && !dealReq.isImport {
-				// When the client proposes an offline deal, save the deal
-				// to the database but don't execute the deal. The deal
-				// will be executed when the Storage Provider imports the
-				// deal data.
-				go p.processDeal(deal, dealReq.rsp)
-				continue
-			}
-
 			if deal.IsOffline && dealReq.isImport {
 				// The Storage Provider is importing offline deal data, so tag
 				// funds for the deal and execute it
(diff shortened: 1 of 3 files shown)


LexLuthr:
Tested on devnet:
Sent 1000 deals in a for loop from bash with no sleep time.

```
real    15m16.104s
user    3m48.002s
sys     2m34.272s
```

Not a single deal timed out. Every deal was responded to correctly by either accepting it or rejecting it.

storagemarket/provider_loop.go
```diff
@@ -389,56 +415,39 @@ func (p *Provider) run() {
...
 
-				// The deal proposal was successful. Send an Accept response to the client.
-				dealReq.rsp <- acceptDealResp{ri: &api.ProviderDealRejectionInfo{Accepted: true}}
-				// Don't execute the deal now, wait for data import.
-				continue
-			}
+				// set up deal handler so that clients can subscribe to deal update events
+				dh, err := p.mkAndInsertDealHandler(deal.DealUuid)
+				if err != nil {
+					sendErrorResp(&acceptError{error: err, isSevereError: true, reason: "server error: starting deal thread"})
```
masih, line 434:
```suggestion
					sendErrorResp(&acceptError{error: err, isSevereError: true, reason: "server error: setting up deal handler"})
```

storagemarket/provider_loop.go
```diff
@@ -389,56 +415,39 @@ func (p *Provider) run() {
...
+					continue
+				}
 
-			// set up deal handler so that clients can subscribe to deal update events
-			dh, err := p.mkAndInsertDealHandler(deal.DealUuid)
-			if err != nil {
-				sendErrorResp(&acceptError{error: err, isSevereError: true, reason: "server error: starting deal thread"})
-				continue
+				// send an accept response
+				dealReq.rsp <- acceptDealResp{&api.ProviderDealRejectionInfo{Accepted: true}, nil}
```
masih, line 446: Specify keys in struct and avoid setting zero value fields (i.e. `err`)?

```suggestion
				dealReq.rsp <- acceptDealResp{ri: &api.ProviderDealRejectionInfo{Accepted: true}}
```

storagemarket/provider_loop.go
```diff
@@ -519,6 +528,68 @@ func (p *Provider) run() {
...
+
+			// start executing the deal
+			_, err = p.startDealThread(dh, deal)
+			if err != nil {
+				sendErrorResp(&acceptError{error: err, isSevereError: true, reason: "server error: starting deal thread"})
+				continue
+			}
+
+			// send an accept response
+			processedDeal.rsp <- acceptDealResp{&api.ProviderDealRejectionInfo{Accepted: true}, nil}
```
masih, line 591: * Ditto; see previous comment.
* Can we do something about duplicate code?

masih (approved):

storagemarket/provider_loop.go
LexLuthr: I have refactored to reduce the duplicate code and fix the RSP channel responses with specific keys.

nonsense:
@LexLuthr I see you posted numbers on how this perform with the new code (this PR). Could you also post numbers on how the code performs on current `main`, i.e. without these changes? Not necessarily 1000 deals if that's too slow, but some numbers for us to understand the impact of this PR. Thanks!

LexLuthr:
@nonsense I don't see much difference when there is no external filter running.

Main
real    0m53.846s
user    0m26.219s
sys     0m13.832s

Current branch
real    0m53.415s
user    0m26.272s
sys     0m12.452s


I have asked @cryptowhizzard to test it on his miner and check if there are any real world improvement in timeout rate.

We probably need to work with CIDGravity to move to a different execution mode than os.Exec to speed up the filter. Preferably an RPC call to a local process.

LexLuthr:
@cryptowhizzard Can you confirm if you are seeing some improvement with this PR?

nonsense (approved):

PR #1749, opened 2023-10-12
Create support ticket template

LexLuthr:
If the template looks good, please feel free to merge it

### DIFF
--- a/.github/ISSUE_TEMPLATE/support-ticket.yml
+++ b/.github/ISSUE_TEMPLATE/support-ticket.yml
@@ -22,7 +22,7 @@ body:
         required: false
       - label: booster-bitswap
         required: false
-      - label: LID Database (Ex: Yugabyte)
+      - label: LID Database - Yugabyte/LevelDB
         required: false
       - label: boostd-data
         required: false
--- a/.github/ISSUE_TEMPLATE/support-ticket.yml
+++ b/.github/ISSUE_TEMPLATE/support-ticket.yml
@@ -1,0 +1,85 @@
+name: Support Ticket
+description: Support ticket for requesting help from Boost team
+title: "[Support Ticket]: "
+labels: ["support", "triage"]
+body:
+- type: checkboxes
+  attributes:
+    label: Boost component
+    description: Please select the boost component you are filing a bug for
+    options:
+      - label: boost daemon - storage providers
+        required: false
+      - label: boost client
+        required: false
+      - label: boost UI
+        required: false
+      - label: boost data-transfer
+        required: false
+      - label: boost index-provider
+        required: false
+      - label: booster-http
+        required: false
+      - label: booster-bitswap
+        required: false
+      - label: LID
+        required: false
+      - label: boostd-data
+        required: false
+      - label: Other
+        required: false
+- type: textarea
+  id: version
+  attributes:
+    label: Boost Version
+    render: text
+    description: Enter the output of `boost --version` and `boostd --version` if applicable.
+    placeholder: |
+      e.g. 
+      Daemon: boostd version 0.0.1+git.b27fec5 
+      Client: boost version 0.0.1+git.b27fec5
+  validations:
+    required: true
+- type: textarea
+  id: Description
+  attributes:
+    label: Describe the problem
+    description: |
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/support-ticket.yml
```diff
@@ -0,0 +1,85 @@
...
+        required: false
+      - label: boost data-transfer
+        required: false
+      - label: boost index-provider
+        required: false
+      - label: booster-http
+        required: false
+      - label: booster-bitswap
+        required: false
+      - label: LID
```
brendalee, line 25: My main concern here is SPs may not know what LID is as it is a new concept. Open to suggestions on what to put here, maybe yugabytedb ? or database?

.github/ISSUE_TEMPLATE/support-ticket.yml
```diff
@@ -0,0 +1,85 @@
...
+    required: true
+- type: textarea
+  id: extraInfo
+  attributes:
+    label: Logging Information
+    render: text
+    description: |
+      Please provide debug logs of the problem, remember you can get set log level control for:
+      * boostd: use `boostd log list` to get all log systems available and set level by `boostd log set-level`.
+      If you don't provide detailed logs when you raise the issue it will almost certainly be the first request we make before furthur diagnosing the problem.
```
brendalee, line 63: I would remove the second sentence here.

.github/ISSUE_TEMPLATE/support-ticket.yml
masih: and what is the difference between boostd-data and LID?

.github/ISSUE_TEMPLATE/support-ticket.yml
```diff
@@ -0,0 +1,85 @@
+name: Support Ticket
+description: Support ticket for requesting help from Boost team
+title: "[Support Ticket]: "
+labels: ["support", "triage"]
+body:
+- type: checkboxes
+  attributes:
+    label: Boost component
+    description: Please select the boost component you are filing a bug for
```
masih, line 9: Just so I understand, what is the rationale for having this option at all? 

It will take very little time for us to label it once the issue is captured, not having it would reduce the barrier for capturing an issue, and we need to read through the issue to check it's labeled right in any case. No?

.github/ISSUE_TEMPLATE/support-ticket.yml
```diff
@@ -0,0 +1,85 @@
...
+      2. Do '...'
+      3. See error '...'
+      ...
+  validations:
+    required: false
+- type: checkboxes
+  id: terms
+  attributes:
+    label: Code of Conduct
+    description: By submitting this issue, you agree to follow our [Code of Conduct](https://example.com)
```
masih, line 82: What requires us to ask this from users on every issue capture?

Typically repos have a code of conduct which is linked to from README, and acceptance of it is implicit by the fact that they _are_ capturing an issue. Right?

.github/ISSUE_TEMPLATE/support-ticket.yml
masih: I concur.

.github/ISSUE_TEMPLATE/support-ticket.yml
nonsense: https://boost.filecoin.io/deployment/boostd-data

.github/ISSUE_TEMPLATE/support-ticket.yml
LexLuthr: Removed. it was left over from the Github template.

.github/ISSUE_TEMPLATE/support-ticket.yml
LexLuthr: There are multiple way to creating a support ticket. But, generally agreed upon industry standard is to collect some basic information to allow support engineers to provide an initial direction of investigation rather than asking same questions from the client/customer. 
If we don't even collect which component would is impacted, then this template would be no different than slack conversations. 
PS: There labels are not github labels.

brendalee:
@LexLuthr lgtm

ischasny (approved):

Issue #1734, opened 2023-10-05
Initial backend work on boostd to support multi minerID setup


ischasny:

ischasny:
Current status: 
1. The backend work is largely done. New configuration parameters have been added to setup multiple miners. The change has been propagated throughout the codebase. 
2. Devnet has been updated to support multiple miners. There is an outstanding bug with address of the second miner not being registered on chain, so it has to be done manually at the time of writing. 
3. UI work is partially done. The breakdown of the changes: 
```
 Storage Deals:
	* Add Provider column - DONE
	* Remove provider id at the top if one is running in the multiminer setup - DONE
	* When clicking on piece cid, add provider id in the Deals table
	* Legacy Deals - remove the link to it - DONE

Deal Proposals 
	* Add Provider Column - DONE
	* Remove provider id at the top if in multiminer mode - DONE

Retrievals:
	* same as above

Local Index Directory:
	* Remove the miner id at the top, remove the filtering from query - DONE

Sealing Pipeline:
	* Remove miner id at the top, on the GQL side aggregate data from all miners before returning - DONE

Published Deals:
	* No changes, just the miner id from the top needs to be removed - DONE

Network Indexer:
	* the same as above - DONE

Storage Space:
	* the same as above - DONE

Funds:
	* remove funds logs
	* remove tagged by boost
	* remove graphics and display all wallets in the table (not supported at the backend yet)
	* hide out “move collateral to escrow” button 

Deal Transfers:
	* No changes, just remove the miner id - DONE

Message Pool:
	* No changes, just remove the miner id - DONE

Settings:
	* wait for Mayank to merge his changes in

At the bottom of the left panel display a list of miner ids (not clickable)

```
4. Migration tooling - not started yet.

PR #1750, opened 2023-10-13
feat: multiminer support for boost deals

ischasny:
This is an in-progress PR for adding multi-miner support to the boost backend. As a part of this work we want to:
1. Change `storagemarket/provider.go` to support multiple miners for Boost storage deals.
2. Update `libp2p-graphsync` retrievals to support multiple miners for retrievals

_Legacy deal flow won't be updated to support multiple miners._

Fixes https://github.com/filecoin-project/boost/issues/1734


gql/resolver_maddr.go
```diff
@@ -1,7 +1,8 @@
 package gql
 
 func (r *resolver) MinerAddress() string {
-	return r.provider.Address.String()
+	// TODO: this function doesn;t seem to be used at all. Confirm.
+	return r.provider.Addresses[0].String()
```
nonsense, line 5: Not used in Go doesn't necessarily mean not used at all. I believe this is a reference to it from React:

```
const MinerAddressQuery = gql`
    query AppMinerAddressQuery {
        minerAddress
    }
`;
```

ischasny: good spot - thanks!

LexLuthr:
Closing as this feature is not planned anymore. We can use this code for Curio market.

PR #1751, opened 2023-10-13
use lotus v1.24.0-rc1

nonsense:


LexLuthr (approved):
We need to match the FFI version as well. Apart from that `make docsgen` is required to be run.

PR #1780, opened 2023-10-26
Group deals by provider when publish

ischasny:
With multiminer support a single batch can include deals from different providers. An onchain message must contain deals from the same provider only. Group deals by provider and publish separate messages to satisfy this requirement.

### DIFF
--- a/markets/storageadapter/dealpublisher.go
+++ b/markets/storageadapter/dealpublisher.go
@@ -321,8 +321,8 @@ func (p *DealPublisher) publishReady(ready []*pendingDeal) {
 	}
 
 	// Validate each deal to make sure it can be published
-	validatedByProvider := make(map[address.Address][]*pendingDeal, 0)
-	dealsByProvider := make(map[address.Address][]market.ClientDealProposal, 0)
+	validated := make(map[address.Address][]*pendingDeal, 0)
+	deals := make(map[address.Address][]market.ClientDealProposal, 0)
 	for _, pd := range ready {
 		// Validate the deal
 		if err := p.validateDeal(pd.deal); err != nil {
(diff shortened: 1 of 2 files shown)


markets/storageadapter/dealpublisher.go
```diff
@@ -321,26 +321,36 @@ func (p *DealPublisher) publishReady(ready []*pendingDeal) {
...
+		validatedByProvider[provider] = append(validatedByProvider[provider], pd)
+		dealsByProvider[provider] = append(dealsByProvider[provider], pd.deal)
+	}
+
+	for provider, deals := range dealsByProvider {
+		validated := validatedByProvider[provider]
+		deals := deals
+		go func() {
+			// Send the publish message
+			msgCid, err := p.publishDealProposals(deals)
```
LexLuthr, line 347: We need to handle this error

LexLuthr (commented):
We would need to put a requirement to use a single daemon node with multi-miner Boost storage. Otherwise, we cannot track PSD messages.
I would recommend checking PublishPendingDeals() in `gql/resolver_dealpublish.go` once to ensure ManualPSD is not broken by the new flow.

LexLuthr (approved):

markets/storageadapter/dealpublisher.go
LexLuthr: I missed that. Rest looks good.

nonsense (changes requested):
I don't think we should merge this, before the information about deals to be published is communicated well in the UI too. Although the code here probably works as expected, I don't think we communicate at all to the user what is happening behind the scene in the UI.

Screenshot 2023-10-26 at 16 13 12

markets/storageadapter/dealpublisher.go
```diff
@@ -321,26 +321,36 @@ func (p *DealPublisher) publishReady(ready []*pendingDeal) {
 	}
 
 	// Validate each deal to make sure it can be published
-	validated := make([]*pendingDeal, 0, len(ready))
-	deals := make([]market.ClientDealProposal, 0, len(ready))
+	validatedByProvider := make(map[address.Address][]*pendingDeal, 0)
```
nonsense, line 324: I prefer the current name. It is clear from the type that this is a map by provider.

markets/storageadapter/dealpublisher.go
```diff
@@ -321,26 +321,36 @@ func (p *DealPublisher) publishReady(ready []*pendingDeal) {
...
-	}
-
-	// Send the publish message
-	msgCid, err := p.publishDealProposals(deals)
-
-	// Signal that each deal has been published
-	for _, pd := range validated {
-		go onComplete(pd, msgCid, err)
+		provider := pd.deal.Proposal.Provider
+		if _, ok := dealsByProvider[pd.deal.Proposal.Provider]; !ok {
```
nonsense, line 334:
```suggestion
		if _, ok := deals[provider]; !ok {
```

nonsense:
Given that this PR is targeted at intermediate branch, feel free to merge, and we can sort out the UI later on.

markets/storageadapter/dealpublisher.go
```diff
@@ -321,26 +321,36 @@ func (p *DealPublisher) publishReady(ready []*pendingDeal) {
 	}
 
 	// Validate each deal to make sure it can be published
-	validated := make([]*pendingDeal, 0, len(ready))
-	deals := make([]market.ClientDealProposal, 0, len(ready))
+	validatedByProvider := make(map[address.Address][]*pendingDeal, 0)
+	dealsByProvider := make(map[address.Address][]market.ClientDealProposal, 0)
```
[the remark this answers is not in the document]
ischasny: fixed

PR #1791, opened 2023-10-31
boostd-data: add index command for debug purposes

nonsense:


extern/boostd-data/cmd/run.go
```diff
@@ -226,3 +233,99 @@ var yugabyteMigrateCmd = &cli.Command{
...
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:     "connect-string",
+			Usage:    "postgres connect string eg 'postgresql://postgres:postgres@localhost'",
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:     "filename",
+			Usage:    "",
```
LexLuthr, line 254:
```suggestion
			Usage:    "filename must be same as pieceCID",
```

extern/boostd-data/cmd/run.go
```diff
@@ -226,3 +233,99 @@ var yugabyteMigrateCmd = &cli.Command{
 		return err
 	},
 }
+
+var yugabyteAddIndexCmd = &cli.Command{
+	Name:   "yugabyte-add-index",
+	Usage:  "Add index via boostd-data to YugabyteDB",
```
LexLuthr, line 239:
```suggestion
	Usage:  "Add index via boostd-data to YugabyteDB (testing tool)",
```

LexLuthr (approved):
Just a couple of changes to clarify use case

extern/boostd-data/yugabyte/service.go
```diff
@@ -70,12 +72,10 @@ func NewStore(settings DBSettings, migrator *Migrator, opts ...StoreOpt) *Store
...
-		settings: settings,
-		cluster:  cluster,
-		migrator: migrator,
-		// The Cassandra driver has a 50k limit on batch statements. Keeping
-		// batch size small makes sure we're under the limit.
-		insertBatchSize: 5000,
+		settings:        settings,
+		cluster:         cluster,
+		migrator:        migrator,
+		insertBatchSize: settings.InsertBatchSize,
```
nonsense, line 78: @ischasny I fail to see where we are setting the default `InsertBatchSize` value. `Settings` is used by many commands, and it seems to me that it will be 0 for a large part of them, no?

extern/boostd-data/yugabyte/service.go
```diff
@@ -39,6 +38,8 @@ type DBSettings struct {
 	PayloadPiecesParallelism int
 	// CQL timeout in seconds
 	CQLTimeout int
+	// Number of records per insert batch
+	InsertBatchSize int
```
nonsense, line 42: I find this confusing... we have `InsertBatchSize` within Settings, but later it is used outside of settings, directly in `Store` - is it possible to be consistent somehow and have it at one place?

extern/boostd-data/yugabyte/service.go
ischasny: good point - fixed

extern/boostd-data/cmd/run.go
```diff
@@ -233,3 +240,113 @@ var yugabyteMigrateCmd = &cli.Command{
...
+			Usage:    "postgres connect string eg 'postgresql://postgres:postgres@localhost'",
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:     "filename",
+			Usage:    "filename must be same as pieceCID",
+			Required: true,
+		},
+		&cli.IntFlag{
+			Name:     "CQLTimeout",
```
ischasny, line 265: would be good to name the command consistently with the rest i.e. `cql-timeout`

nonsense: agree, but for now i'll stay consistent with the rest of the commands, and every one of the uses `CQLTimeout` at the moment.

LexLuthr (approved):

PR #1795, opened 2023-11-01
yugabyte/service: parallel batch inserts for car indices

nonsense:
This PR builds on top of the work @ischasny did with batch inserts for indices.

It basically splits the `recs` slice (which can get from 0 up to 50M-100M entries), into 32 chunks that get processed in parallel.

Considering that batch size is 10k, it makes sense to have concurrent go-routines working on inserting the index.

### DIFF
--- a/extern/boostd-data/yugabyte/service.go
+++ b/extern/boostd-data/yugabyte/service.go
@@ -537,28 +537,54 @@ func (s *Store) addMultihashesToPieces(ctx context.Context, pieceCid cid.Cid, re
 	insertPieceOffsetsQry := `INSERT INTO PayloadToPieces (PayloadMultihash, PieceCid) VALUES (?, ?)`
 	pieceCidBytes := pieceCid.Bytes()
 
-	var batch *gocql.Batch
-	for allIdx, rec := range recs {
-		if batch == nil {
-			batch = s.session.NewBatch(gocql.UnloggedBatch).WithContext(ctx)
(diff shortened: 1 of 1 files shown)


LexLuthr (approved):

extern/boostd-data/yugabyte/service.go
```diff
@@ -537,28 +537,54 @@ func (s *Store) addMultihashesToPieces(ctx context.Context, pieceCid cid.Cid, re
 	insertPieceOffsetsQry := `INSERT INTO PayloadToPieces (PayloadMultihash, PieceCid) VALUES (?, ?)`
 	pieceCidBytes := pieceCid.Bytes()
 
-	var batch *gocql.Batch
-	for allIdx, rec := range recs {
-		if batch == nil {
-			batch = s.session.NewBatch(gocql.UnloggedBatch).WithContext(ctx)
-			batch.Entries = make([]gocql.BatchEntry, 0, s.settings.InsertBatchSize)
+	threadBatch := len(recs) / 32 // split the slice into go-routine batches for ~32 workers
```
ischasny, line 540: Shall we take an opportunity and extract batching / parallelisation logic into a separate function as it used twice

ischasny: Otherwise lgtm

LexLuthr: I am merging this as is for now. We can shorten the code for next release.

PR #1801, opened 2023-11-03
fix: optimise add index performance

ischasny:
Slice recs into multiple smaller chunks when invoking AddIndex from boostd. Rpc data transfer has proven to take the longest time out of all indexing process.

Example for 56mil indexes:
```
Index Generation ~ 6 minutes (~56.5 mil)
Inserting Index into DB ~ 15 minutes
Breakdown of index insertion:
       dataTransfer ~ 8 minutes
       addMultihashesToPieces = 2 minutes
       addPieceInfos = 5m30s
       createPieceMetadata = 1s
```

Local performance test results against 1mil index records
```
1. 8 goroutines for data tranfer, 4 for insertion - 17 sec
2. 1 goroutine for data transfer, 1 for insertion - 50 sec 
3. 1 goroutine for data transfer, 32 for insertion - 27 sec (current main)
4. 8 coroutines for data transfer, 32 for insertion - 1.5min
```

Hence pick option 1 as reasonable defaults


nonsense:
@ischasny please post a comparison, before/after this PR, from the description it is not clear what the improvement is.

extern/boostd-data/ldb/db.go
```diff
@@ -114,7 +114,11 @@ func (db *DB) NextCursor(ctx context.Context) (uint64, string, error) {
 	}
 
 	cursor, _ := binary.Uvarint(b)
-	return cursor, fmt.Sprintf("%d", cursor) + "/", nil // adding "/" because of Query method in go-datastore
+	return cursor, KeyCursorPrefix(cursor), nil // adding "/" because of Query method in go-datastore
+}
+
+func KeyCursorPrefix(cursor uint64) string {
+	return fmt.Sprintf("%d", cursor) + "/"
```
nonsense, line 121: `Sprintf` is already formatting a string, so might as well put / in the initial formatting.

masih (approved):

piecedirectory/piecedirectory.go
```diff
@@ -292,12 +294,35 @@ func (ps *PieceDirectory) addIndexForPiece(ctx context.Context, pieceCid cid.Cid
 	if !errors.Is(err, io.EOF) {
 		return fmt.Errorf("generating index for piece %s: %w", pieceCid, err)
 	}
+	concurrency := AddIndexConcurrency
+	// in an unlikely case if there are less than 8 indexes
```
nonsense, line 298:
```suggestion
	// in an unlikely case if there are less than 8 records in the index
```

piecedirectory/piecedirectory.go
```diff
@@ -292,12 +294,35 @@ func (ps *PieceDirectory) addIndexForPiece(ctx context.Context, pieceCid cid.Cid
...
+	if concurrency > len(recs) {
+		concurrency = len(recs)
+	}
+
+	chunkSize := len(recs) / concurrency
```
nonsense, line 303: Chunk size generally refers to block size. Here it is really the `range` of the slice that each worker will process, so probably worth renaming `chunkSize` to something else, for example `rangeLen` or `rangeLength`.

piecedirectory/piecedirectory.go
```diff
@@ -292,12 +294,35 @@ func (ps *PieceDirectory) addIndexForPiece(ctx context.Context, pieceCid cid.Cid
...
+	eg := errgroup.Group{}
 
-	// Add mh => piece index to store: "which piece contains the multihash?"
-	// Add mh => offset index to store: "what is the offset of the multihash within the piece?"
-	log.Debugw("add index: store index in local index directory", "pieceCid", pieceCid)
-	if err := ps.store.AddIndex(ctx, pieceCid, recs, true); err != nil {
-		return fmt.Errorf("adding CAR index for piece %s: %w", pieceCid, err)
+	for i := 0; i < concurrency; i++ {
+		idx := i
```
nonsense, line 307: Nit, but you can just continue to use `i` here, as in `i := i`, no need to introduce another name.

cryptowhizzard:
This is a good improvement. Thanks!

PR #1807, opened 2023-11-07
fix: piece doctor stuck in checking loop due to very short PieceCheckPeriod

LexLuthr:
Minimum period is now 60 minutes and each piece check time is assumed 1 second


LexLuthr:
This fix is not working.

Issue #1798, opened 2023-11-02
The "Awaiting Publish Message" UI causes boost to crash


LexLuthr:
The crash is caused in Elapsed calculation when message is sent
```
2023/11/02 08:32:10 graphql: panic occurred: runtime error: invalid memory address or nil pointer dereference
goroutine 6735815 [running]:
github.com/graph-gophers/graphql-go/log.(*DefaultLogger).LogPanic(0x0?, {0x6a14f30?, 0xc015614480}, {0x4f77820, 0x8e0aab0})
        /home/nonsense/go/pkg/mod/github.com/graph-gophers/graphql-go@v1.3.0/log/log.go:21 +0x65
github.com/graph-gophers/graphql-go/internal/exec.execFieldSelection.func2.1()
        /home/nonsense/go/pkg/mod/github.com/graph-gophers/graphql-go@v1.3.0/internal/exec/exec.go:187 +0x89
panic({0x4f77820, 0x8e0aab0})
        /usr/local/go/src/runtime/panic.go:884 +0x213
github.com/filecoin-project/boost/lib/mpoolmonitor.(*MpoolMonitor).MsgExecElapsedEpochs(0xc00c9b86c0, {0x6a14f68, 0xc015614630}, {{0xc19c83cab0?, 0x1?}})
        /home/nonsense/boost/lib/mpoolmonitor/mpoolmonitor.go:196 +0x185
github.com/filecoin-project/boost/gql.(*dealResolver).message(0xc00eda4800, {0x6a14f68?, 0xc015614630?}, 0xc000b6f660?, {0x881855?, 0x50efec0?, 0x50efec0?})
        /home/nonsense/boost/gql/resolver.go:620 +0x588
github.com/filecoin-project/boost/gql.(*dealResolver).Message(0xc00eda4800, {0x6a14f68?, 0xc015614630?})
        /home/nonsense/boost/gql/resolver.go:574 +0x48
reflect.Value.call({0x540b9a0?, 0xc00eda4800?, 0x8205f6?}, {0x5487e55, 0x4}, {0xc00d3005d0, 0x1, 0x87c72e?})
        /usr/local/go/src/reflect/value.go:586 +0xb0b
reflect.Value.Call({0x540b9a0?, 0xc00eda4800?, 0xc000130720?}, {0xc00d3005d0?, 0x5469aa0?, 0x0?})
        /usr/local/go/src/reflect/value.go:370 +0xbc
[...]

PR #1817, opened 2023-11-15
fix: fix nil pointer panics in mpoolmonitor

ischasny:
Fixes: https://github.com/filecoin-project/boost/issues/1798


bobdubois:
fix tested and panic errors no longer seen in boostd.log

PR #1820, opened 2023-11-22
bump lotus to v1.25.0

nonsense:
- [ ] Test in devnet


extern/boostd-data/Dockerfile.test
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.18-alpine
+FROM golang:1.21-alpine
```
LexLuthr, line 1: I think default is v1.20 for Golang (based on Lotus). Shall we stick to that?

LexLuthr (approved):
LGTM!

extern/boostd-data/Dockerfile.test
nonsense: I don't think it matters much -- this docker stack is used solely for testing.

PR #1837, opened 2023-12-01
Update deal list page

LexLuthr:
This PR makes the following changes:
1. Add a new column for on chain deal ID (Boost and Legacy deals)
Screenshot 2023-12-01 at 5 48 30 PM
Screenshot 2023-12-01 at 5 47 46 PM
2. Create a new column "Sealing Status" and create an info button with description
Screenshot 2023-12-01 at 5 47 52 PM

3. Add info button to deal Status column
Screenshot 2023-12-01 at 5 48 00 PM

4. Fix Size for offline deals (we cannot fix for existing deals but new ones should display and retain correct size)
Screenshot 2023-12-04 at 2 41 46 PM
Screenshot 2023-12-04 at 2 45 52 PM

5. The onchainDealID and epochs do not have comma anymore
Screenshot 2023-12-01 at 5 48 13 PM

### DIFF
--- a/react/src/DealDetail.js
+++ b/react/src/DealDetail.js
@@ -547,8 +547,8 @@ export function DealStatusInfo(props) {
             <p>
                 <i>Indexing</i>
                 <p>
-                    Boost is indexing the deal in Local Index Directory(LID)
-                    and will be announcing the deal to the network so that clients know where to retrieve it.
+                    Boost is indexing the deal in the Local Index Directory(LID)
+                    and will be announcing the deal to the network so that clients know where to retrieve it from.
                 </p>
             </p>
             <p>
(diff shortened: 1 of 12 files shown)


react/src/DealDetail.js
```diff
@@ -545,15 +545,16 @@ export function DealStatusInfo(props) {
                 </p>
             </p>
             <p>
-                <i>Announcing</i>
+                <i>Indexing</i>
                 <p>
-                    Boost is announcing the deal to the network so that clients know where to retrieve it.
+                    Boost is indexing the deal in Local Index Directory(LID)
+                    and will be announcing the deal to the network so that clients know where to retrieve it.
```
nonsense, line 551:
```suggestion
                    Boost is indexing the deal in the Local Index Directory(LID)
                    and will be announcing the deal to the network so that clients know where to retrieve it from.
```

nonsense (approved):

PR #1853, opened 2023-12-20
Create 202312210049_add_sectorid_index.sql

RobQuistNL:
This speeds up a ton of queries as this is quite often called


RobQuistNL:
This might have been one of the major slowdowns in the deal making process, as it locks up the root DB which is used a lot.

RobQuistNL:
A previous deal request took 24 seconds
A current one (with this index) took 800ms

I'll be keeping an eye on the next incoming deal requests

RobQuistNL:
all subsequent deal requestst were handled < 2s (mostly <1s)

LexLuthr (approved):

PR #1855, opened 2023-12-21
Prefetch state claims to change O(n) calls to Lotus State to O(1)

RobQuistNL:
In my case (140.000 updates) shaves off 1.5 seconds per call (+ the load on the daemon) -> turns a 39 hour job into seconds

### DIFF
--- a/piecedirectory/doctor.go
+++ b/piecedirectory/doctor.go
@@ -17,6 +17,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
+	verifregtypes "github.com/filecoin-project/go-state-types/builtin/v9/verifreg"
 )
 
 var doclog = logging.Logger("piecedoc")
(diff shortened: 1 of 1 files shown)


LexLuthr:
@RobQuistNL Thank you for the PR. Looks good.

piecedirectory/doctor.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
+	verifregtypes "github.com/filecoin-project/go-state-types/builtin/v9/verifreg"
```
LexLuthr, line 20: Can you please check with v12 verifreg? We are trying to use latest actors as much as possible.

LexLuthr (commented):
One nit about the actor version. Rest is good.

piecedirectory/doctor.go
RobQuistNL: Are you sure that will work? The v9 one is the one used in the return type of the API call

LexLuthr: Ahhh! Looks like a Lotus side dependency. Merging.

LexLuthr (approved):

PR #1859, opened 2023-12-25
feat: index ad removal command

LexLuthr:
This command will send an ad to IPNI to remove the indexes for a particular deal. This command is to be used for removing deals not being cleaned up by automated system in Boost.


api/api.go
```diff
@@ -41,6 +41,8 @@ type Boost interface {
 	BoostDeal(ctx context.Context, dealUuid uuid.UUID) (*smtypes.ProviderDealState, error)                                                      //perm:admin
 	BoostDealBySignedProposalCid(ctx context.Context, proposalCid cid.Cid) (*smtypes.ProviderDealState, error)                                  //perm:admin
 	BoostDummyDeal(context.Context, smtypes.DealParams) (*ProviderDealRejectionInfo, error)                                                     //perm:admin
+	BoostIndexerAnnounceDealRemoved(ctx context.Context, pieceCid cid.Cid) (cid.Cid, error)                                                     //perm:admin
+	LegacyDealByProposalCid(ctx context.Context, propCid cid.Cid) (storagemarket.MinerDeal, error)                                              //perm:admin
```
nonsense, line 45: This is added under `Boost` MethodGroup, so maybe add `Boost` prefix?

api/proxy_gen.go
```diff
@@ -454,6 +458,17 @@ func (s *BoostStub) BoostIndexerAnnounceAllDeals(p0 context.Context) error {
 	return ErrNotSupported
 }
 
+func (s *BoostStruct) BoostIndexerAnnounceDealRemoved(p0 context.Context, p1 cid.Cid) (cid.Cid, error) {
+	if s.Internal.BoostIndexerAnnounceDealRemoved == nil {
+		return *new(cid.Cid), ErrNotSupported
```
nonsense, line 463: I think you can just use `cid.Undef` instead of `*new(cid.Cid)`

LexLuthr: This is auto generated by the `make gen`. We might have to adjust the types in api gen for this. I am not sure if it is worth the required changes.

api/api.go
LexLuthr: Updated. The reason to have it under Boost group is ensure compatibility after removing legacy markets.

nonsense (approved):

api/proxy_gen.go
nonsense: Yes, sorry, didn't notice this is the auto-generated part of the API.

PR #1867, opened 2024-01-10
Upgrade dependencies

masih:
Upgrade multiaddr dependency


LexLuthr (commented):
We just had a cbor update merged to main. I think you need to run `make gen` and `make docsgen` to fix the failing tests.

LexLuthr (approved):

PR #1873, opened 2024-01-16
Fixes POSDI error

cryptowhizzard:
POSDI index error fix for spade deals

### DIFF
--- a/piecedirectory/piecedirectory.go
+++ b/piecedirectory/piecedirectory.go
@@ -406,118 +406,139 @@ func (cr *countingReader) Read(p []byte) (n int, err error) {
 }
 
 func parsePieceWithDataSegmentIndex(pieceCid cid.Cid, unpaddedSize int64, r types.SectionReader) ([]model.Record, error) {
+    concurrency := runtime.NumCPU()
+    if concurrency < PodsiMinConcurrency {
+        concurrency = PodsiMinConcurrency
+    }
+    if concurrency > PodsiMaxConcurrency {
+        concurrency = PodsiMaxConcurrency
+    }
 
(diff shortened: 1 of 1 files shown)


nonsense:
@cryptowhizzard could you explain what the fix here is? Apart from DEBUG lines, I am not sure I see any changes to the current version of the code. Thanks!

cryptowhizzard:
That is a very good question. It's probably because we query the Filesize etc. I was debugging and looking why I got an EOF error running PODSI on the spade deals , but that is gone now.

So we are well on our way to finding the next bug , probably in RPC with lotus miner or something.

Anyway , I will ask Stuberman on slack to check if these deals do now pass and index with this and keep you updated.

piecedirectory/piecedirectory.go
```diff
@@ -406,119 +406,140 @@ func (cr *countingReader) Read(p []byte) (n int, err error) {
...
+    // Debug: Print the file size
+    file, ok := r.(io.ReadSeeker)
+    if ok {
+        fileSize, err := file.Seek(0, io.SeekEnd)
+        if err != nil {
+            log.Debugw("Failed to get file size", "error", err)
+        } else {
+            log.Debugw("File size", "size", fileSize)
+        }
+        _, err = file.Seek(0, io.SeekStart) // Return the file pointer to the start
```
LexLuthr, line 436: Shouldn't we set the seeker back to the start of the segment index offset same as line 422? The current code would work for 32 GiB files but might be problematic for smaller deals.

rvagg: yes, `dsis` is kind of important for the `parseDataSegmentIndex` usage below, `0` is not a good idea here

PR #1874, opened 2024-01-16
fix: unexpected EOF when parsing segments for PODSI deals

LexLuthr:
CC: @cryptowhizzard 

DO NOT MERGE

### DIFF
--- a/piecedirectory/piecedirectory.go
+++ b/piecedirectory/piecedirectory.go
@@ -406,139 +406,133 @@ func (cr *countingReader) Read(p []byte) (n int, err error) {
 }
 
 func parsePieceWithDataSegmentIndex(pieceCid cid.Cid, unpaddedSize int64, r types.SectionReader) ([]model.Record, error) {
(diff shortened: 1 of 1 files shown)


piecedirectory/piecedirectory.go
```diff
@@ -432,15 +442,18 @@ func parsePieceWithDataSegmentIndex(pieceCid cid.Cid, unpaddedSize int64, r type
 	panicked := false
 	indexData, err := parseDataSegmentIndex(pieceCid, bufio.NewReaderSize(cr, PodsiBuffesrSize), &panicked)
 	if err != nil {
+		log.Debugw("Failed to parse data segment index", "error", err)
```
ischasny, line 445: This additional logging isn't required because the error will be logged here already: https://github.com/filecoin-project/boost/blob/main/piecedirectory/piecedirectory.go#L321

cryptowhizzard:
Confirmed & working.

Issue #1709, opened 2023-09-22
[Epic] LID cleanup


LexLuthr:
_This is the parent tracking issue related cleanup of indexes and deals from LID within Boost._

# Problem Statement
Since the launch of Lotus, there have been no clean up policy in place for the metadata of expired or lost deals. We carried this behaviour into Boost and Boost v2 as well. After the launch of LID, this old metadata has become a nuisance causing the following problems.
1. Indexers still have old indexes for expired deals carried forwarded from lotus markets. 
2. These older indexes lead to a higher failure rate of retrievals by validation bots.
![image (9)](https://github.com/filecoin-project/boost/assets/88259624/69bb9da3-9018-48c3-9838-ae04271e6045)

3. When a new deals is lost (sector loss  for any reason), we still don't clean up the deal details even though it serves no purpose. This leads to accumulation of indexes which are just occupying space on the DB.
4. Inconsistencies in the state of a piece and indexes when same piece refers to old removed/expired deals and new healthy deals

# Change Proposal
Filecoin's main USP is Proof of spacetime. The argument that SPs can continue to serve the data even after sector expires without any incentive seems a bit premature. This epic propose the following changes to the existing deal flow and how LID cleans up the data.

- [ ] Clean up older deals from LID with a background job
    - [ ] The background job frequency must be configurable
    - [ ] The background job must also check deal expiration and clean up the expired deals. This should also take into account the sector state manager
    - [ ] An removal ad must be sent for each deal so indexers can be update about the new state of SP
- [ ] Clean up new deal from LID if sector reaches a final state other than `proving`
    - [ ] Boost must check that sector actually contains the deal when it reached the proving state for snap deals

PR #1881, opened 2024-02-28
feat: LID clean up

LexLuthr:
Fixes https://github.com/filecoin-project/boost/issues/1709

Implementation Details:
1. The process runs every 6 hours
2. It removes Completed/Slashed Boost deals, Completed Legacy Deals and TermMax reached Direct deals from LID
3. It also removed any dangling deals from LID which are not associated with any deals in SQL DB but are attributed to the miner ID

- [x] Account for FIP-45 as deal end epoch is no longer reliable


lib/pdcleaner/pdcleaner.go
```diff
@@ -0,0 +1,249 @@
...
+		case <-ticker.C:
+			log.Infof("Starting LID clean up")
+			err := p.CleanOnce()
+			if err != nil {
+				log.Errorf("Failed to cleanup LID: %s", err)
+				continue
+			}
+			log.Info("Finished cleaning up LID")
+		case <-p.ctx.Done():
+			return
```
masih, line 99: Log message at debug level to say that  clean up event loop has stopped?

lib/pdcleaner/pdcleaner.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return fmt.Errorf("getting complete boost deals: %w", err)
+	}
+	boostActiveDeals, err := p.dealsDB.ListActive(p.ctx)
+	if err != nil {
+		return fmt.Errorf("getting active boost deals: %w", err)
+	}
+
+	boostDeals = append(boostDeals, boostCompleteDeals...)
+	boostDeals = append(boostDeals, boostActiveDeals...)
```
masih, line 130: Seems like `boostDeals` size is known. Instantiate it with the right size after complete and active deal slices are determined? That should result in a more efficient allocation.

lib/pdcleaner/pdcleaner_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/ipld/go-car/v2"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPieceDirectoryCleaner(t *testing.T) {
+	req := require.New(t)
+	ctx := context.Background()
+
+	provAddr, err := address.NewIDAddress(1523)
+	require.NoError(t, err)
```
masih, line 37: Use `req` consistently or remove its allocation and stick with package level exported testify library?

masih (approved):

lib/pdcleaner/pdcleaner.go
```diff
@@ -0,0 +1,263 @@
...
+
+func NewPieceDirectoryCleaner(cfg *config.Boost) func(lc fx.Lifecycle, dealsDB *db.DealsDB, directDealsDB *db.DirectDealsDB, legacyDeals legacy.LegacyDealManager, pd *piecedirectory.PieceDirectory, full v1api.FullNode) PieceDirectoryCleanup {
+	return func(lc fx.Lifecycle, dealsDB *db.DealsDB, directDealsDB *db.DirectDealsDB, legacyDeals legacy.LegacyDealManager, pd *piecedirectory.PieceDirectory, full v1api.FullNode) PieceDirectoryCleanup {
+
+		pdc := newPDC(dealsDB, directDealsDB, legacyDeals, pd, full, cfg.LocalIndexDirectory.LidCleanupInterval)
+
+		ctx, cancel := context.WithCancel(context.Background())
+
+		lc.Append(fx.Hook{
+			OnStart: func(_ context.Context) error {
```
masih, line 51:
```suggestion
			OnStart: func(context.Context) error {
```

lib/pdcleaner/pdcleaner.go
```diff
@@ -0,0 +1,263 @@
...
+				}
+				mid, err := address.NewFromString(cfg.Wallets.Miner)
+				if err != nil {
+					return fmt.Errorf("failed to parse the miner ID %s: %w", cfg.Wallets.Miner, err)
+				}
+				pdc.miner = mid
+				go pdc.Start(ctx)
+				return nil
+			},
+			OnStop: func(_ context.Context) error {
```
masih, line 64:
```suggestion
			OnStop: func(context.Context) error {
```

node/config/types.go
```diff
@@ -289,6 +289,9 @@ type LocalIndexDirectoryConfig struct {
 	ServiceRPCTimeout Duration
 	// PieceDoctor runs a continuous background process to check each piece in LID for retrievability
 	EnablePieceDoctor bool
+	// Interval in hours after which LID clean up job should rerun. The cleanup entails removing indices and metadata
+	// for the expired/slashed deals. Disabled if set to 0.
+	LidCleanupInterval int
```
masih, line 294: I am curious why an `int` type? Would it be clearer to use `Duration` type considering it exists and that's what the ticker uses down the line?

lib/pdcleaner/pdcleaner.go
```diff
@@ -0,0 +1,263 @@
...
+				if err != nil {
+					return err
+				}
+				if len(dd) > 0 {
+					continue
+				}
+
+				err = p.pd.RemoveDealForPiece(p.ctx, piece, deal.DealUuid)
+				if err != nil {
+					// Don't return if cleaning up a deal results in error. Try them all.
```
masih, line 255: I am curious what happens to the dangling deals that fail to get removed?
Ditto re legacy deals above.

masih (approved):

lib/pdcleaner/pdcleaner.go
LexLuthr: This is to account for any DB errors. We can always retry in the later attempt. it seem inefficient to stop the cleanup for a single failure as keep that data for a little more time won't hurt the SP.

node/config/types.go
LexLuthr: I have updated to time.Duration(). I had 2 different approaches in mind and merged the 2 which resulted in ticker with duration and var with int.

lib/pdcleaner/pdcleaner.go
masih: Proceeding seems reasonable. Though I am curious where if ever those lingering deals get cleaned up. 

Context: addressing the cleanup retry for lingering deals can certainly be done in a separate PR. Just now I need your help to understand the systems dynamics in such failure cases.

LexLuthr: Imagine if there is issue with deleting a particular deal, it could be an ActorState or DB timeout or something else, we will not be cleaning up other deals which could have been cleaned without any issue. If such a problem comes very early in the execution then we keep holding the data we don't need. All the failures are logged and user can easily use CLI to try to remove that deal manually. This step can also help us debug the problem. 
The only exception would be we cannot CLI cannot cleanup deals with no corresponding identifier in SQL. We can handle that later if reported. 

I hope this makes sense.

lib/pdcleaner/pdcleaner.go
```diff
@@ -0,0 +1,308 @@
...
+				if err != nil {
+					return fmt.Errorf("checking if bitfield is set: %w", err)
+				}
+				// Each claim is created with ProveCommit message. So, a sector in claim cannot be unproven.
+				// it must be either Active(Proving, Faulty, Recovering) or terminated. If bitfield is not set
+				// then sector must have been terminated. This method will also account for future change in sector numbers
+				// of a claim. Even if the sector is changed then it must be Active as this change will require a
+				// ProveCommit message
+				if !present {
+					err = p.pd.RemoveDealForPiece(p.ctx, d.PieceCID, d.ID.String())
```
rvagg, line 237: I'm not clear on the sectors bitfield bit but the rest of this seems to me to be saying:

1. Find a list of "completed" direct deals (what's the criteria for "completed", does that mean that the deal has been handed off to the miner?)
2. Look up the allocation ID in the list of claims to see if it's been converted to a claim
3. Check that the sector has been activated, or at least is in an active state
4. if it's not active, then clean up the deal from the local piece directory

So, what are the cases this is trying to account for? There's two that I think I see:

1. "got the claim, but failed to activate the sector at the last hurdle"—which should be quite hard, but looking at the code not entirely impossible.
2. The sector was terminated but claim is dangling, as outlined in https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0045.md#sector-termination, which kind of sucks but I guess this is what makes the two-step process necessary?

LexLuthr: Sector bitfield comes from Actor state and account for all the sectors which are active for a miner. Active sectors include the ones which are not terminated or are not yet on-chain. 
1. Complete means that the sector it AP()ed in is now in proving state.

Accounting:
1. Any deals for which we don't activate a sector. We should clean up those.
2. Any deals which were claimed but now claim has expired and sector has been terminated should be cleaned up.
3. Any deal where sector has changed should not be cleaned up. (Future reseal).
4. Any deal which creates a claim but sector goes to faulty state instead of proving should not be cleaned up. This can work as faulty sectors are part of Active sector bitfield.

2 more things to test:
1. How does Boost behaves when sector reaching faulty state instead of proving.
2. Confirm if newly proving sector (without first wdPost) are part of the Active Bitfield. If not then update code to account for those. 


To sum it up, if I am not holding a deal, I should not hold the metadata either. We can expand this logic in future if more cases or edge cases are found.

rvagg: sounds good and reasonable

lib/pdcleaner/pdcleaner.go
```diff
@@ -218,33 +221,89 @@ func (p *pdcleaner) CleanOnce() error {
...
 		for _, d := range completeDirectDeals {
-			// AllocationID and ClaimID should match
 			cID := verifregtypes.ClaimId(d.AllocationID)
 			c, ok := claims[cID]
+			// If claim found
 			if ok {
-				present, err := activeSectors.IsSet(uint64(c.Sector))
+				// Claim Sector number and Deal Sector number should match(regardless of how DDO works)
+				// If they don't match and older sector is removed, then we can't use the metadata
+				// This check can be removed once Curio has resealing enabled, and it can provide
```
rvagg, line 250: 🤞 that you remember this! it'll be annoying if that lands and this is forgotten

lib/pdcleaner/pdcleaner.go
```diff
@@ -218,33 +221,89 @@ func (p *pdcleaner) CleanOnce() error {
...
 				}
+				continue
+			}
+
+			// If no claim found
+			clientAllocs, err := verifregState.GetAllocations(d.Client)
+			if err != nil {
+				return fmt.Errorf("getting allocations for client %s: %w", d.Client, err)
+			}
+			alloc, ok := clientAllocs[d.AllocationID]
```
rvagg, line 284: I think you can just use `verifregState.GetAllocation(d.Client, d.AllocationID)` here to make this much cheaper?

lib/pdcleaner/pdcleaner.go
```diff
@@ -218,33 +221,89 @@ func (p *pdcleaner) CleanOnce() error {
...
+				err = p.pd.RemoveDealForPiece(p.ctx, d.PieceCID, d.ID.String())
+				if err != nil {
+					// Don't return if cleaning up a deal results in error. Try them all.
+					log.Errorf("cleaning up direct deal %s for piece %s: %s", d.ID.String(), d.PieceCID, err.Error())
+				}
+				continue
+			}
+
+			// If claim is found then we should check if it is expired. If expired, clean up the deal.
+			if alloc.Expiration < head.Height() {
```
rvagg, line 299: these are the same logic, maybe just lump them in together, `if !ok || alloc.Expiration < head.Height() {`

lib/pdcleaner/pdcleaner.go
LexLuthr: It is not just this but the normal deal flow as well. Everything will be impacted by Curio's robustness.

lib/pdcleaner/pdcleaner.go
LexLuthr: Yes. Merging them.

PR #1886, opened 2024-03-07
fix: refactor LID clean up

LexLuthr:
Currently, LID clean up uses a very complex logic to figure out which deals/claims have not expired/slashed on chain to determine if the deal should be cleaned up.
This logic can be simplified into a check if sector containing the deal exists for miner or not. If not then on chain state of deal does not matter as we cannot serve retrievals for such deals.


LexLuthr:
@masih @rvagg I don't like they way test is setup. I was thinking about converting this into an itest instead of current way. Please let me know your thoughts on the same. If there is a way to emulate a Store, that would be the best. I could not find one.

cmd/boostd/index.go
```diff
@@ -66,13 +62,46 @@ var indexProvListMultihashesCmd = &cli.Command{
 		}
 		defer closer()
 
-		// get list of multihashes
-		mhs, err := napi.BoostIndexerListMultihashes(ctx, propCid)
+		if cctx.Args().Len() != 1 {
```
snadrus, line 65: what's wrong with a for-loop?

cmd/boostd/index.go
```diff
@@ -66,13 +62,46 @@ var indexProvListMultihashesCmd = &cli.Command{
...
+			return nil
+		}
+
+		mhs, err = napi.BoostIndexerListMultihashes(ctx, proposalCid.Bytes())
 		if err != nil {
 			return err
 		}
 
-		fmt.Printf("Found %d multihashes for deal with proposal cid %s:\n", len(mhs), propCid)
+		fmt.Printf("Found %d multihashes for deal with ID %s:\n", len(mhs), id)
```
snadrus, line 104: any reason this couldn't be JSON? It would be more appreciated for system tools. 
```json
{
"multihashes":  {
      "dealIdentifier1": [ "firstHash", "secondhash"],
      "dealIdentifier2": [ "thirdHash", "fourthHash"],
      "dealIdentifier3": [ ],
      }
}
```

it's as easy as 
```go
e := json.NewEncoder(os.Stdout)
e.SetIndent("  ",  "  ")
e.Encode(map[string]any{ "multihashes": map_Built_Iteratively})
```

itests/lid_cleanup_test.go
```diff
@@ -0,0 +1,306 @@
...
+		require.Fail(t, "direct data import rejected: %s", rej.Reason)
+	}
+	t.Log("Direct data import scheduled for execution")
+
+	// Wait for sector to start sealing
+	time.Sleep(2 * time.Second)
+
+	states := []lapi.SectorState{lapi.SectorState(sealing.Proving)}
+
+	// Exit if all sectors are now proving
```
snadrus, line 197: can you use require.Eventually() to wait for sealed sectors to land?

itests/lid_cleanup_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+	// Wait for wdPost
+	for {
+		head, err := f.FullNode.ChainHead(ctx)
+		require.NoError(t, err)
+		if head.Height() > st.Activation+abi.ChainEpoch(2880) {
+			break
+		}
+		t.Log("Waiting for first wdPost after committing sectors")
+		time.Sleep(3 * time.Second)
```
snadrus, line 240: require.Eventually()

lib/pdcleaner/pdcleaner.go
```diff
@@ -154,148 +148,56 @@ func (p *pdcleaner) CleanOnce() error {
...
-				// If deal is slashed or end epoch has passed. No other reason for deal to reach termination
-				toCheck := termOrSlash(md.Proposal.EndEpoch, md.State.SlashEpoch)
-				if toCheck < head.Height() {
-					err = p.pd.RemoveDealForPiece(p.ctx, d.ClientDealProposal.Proposal.PieceCID, d.ProposalCid.String())
-					if err != nil {
-						// Don't return if cleaning up a deal results in error. Try them all.
-						log.Errorf("cleaning up legacy deal %s for piece %s: %s", d.ProposalCid.String(), d.ClientDealProposal.Proposal.PieceCID.String(), err.Error())
-					}
-				}
+		present, err := activeSectors.IsSet(uint64(d.SectorNumber))
```
snadrus, line 178: should this happen in an:
```go
eg := errgroup.Group{}
eg.SetParallelism(10)
for /*.......*/ {
    eg.Go(func() error {
         //this stuff
   }
}
err := eg.Wait()
```

That way slow systems don't bog as much

snadrus (approved):

cmd/boostd/index.go
LexLuthr: The output can be really long like upto ~55 million multihashes. It would be ideal to do this one by one. This is a very rarely used cmd. Mostly used for debugging.

cmd/boostd/index.go
LexLuthr: We have a dedicated method in GraphQL API for this. The output there is a JSON structure with array. containing multihash and corresponding CID as well.

PR #1890, opened 2024-03-11
fix: allocate should not compare tmin and tmax with head

LexLuthr:


LexLuthr:
@TippyFlitsUK Please let us know if the defaults work in calibnet.

TippyFlitsUK:
These are all working as expected now 👍

snadrus (approved):

PR #1900, opened 2024-03-27
fix: extend-claim cli DC batching

LexLuthr:
This PR fixes the Datacap used per batch and reduces the batch size to 500.

### DIFF
--- a/cmd/boost/direct_deal.go
+++ b/cmd/boost/direct_deal.go
@@ -293,8 +293,8 @@ var clientExtendDealCmd = &cli.Command{
 	Name:  "extend-claim",
 	Usage: "extend claim expiration (TermMax)",
 	UsageText: `Extends claim expiration (TermMax).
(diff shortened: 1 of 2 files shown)


cmd/boost/direct_deal.go
```diff
@@ -293,8 +293,8 @@ var clientExtendDealCmd = &cli.Command{
 	Name:  "extend-claim",
 	Usage: "extend claim expiration (TermMax)",
 	UsageText: `Extends claim expiration (TermMax).
-If the client is original client then claim can be extended to Maximum 5 years and no Datacap is required.
-If the client id different then claim can be extended up to Maximum 5 years from now and Datacap is required.
+If the client is original client then claim can be extended to maximum 5 years and no Datacap is required.
```
snadrus, line 296:
```suggestion
If the client is the original client, then the claim can be extended up to a maximum of 5 years, and no Datacap is required.
```

snadrus (approved):

PR #1931, opened 2024-06-19
fix: single stream download for small files

LexLuthr:


ZenGround0:
Thanks!  I'll test this out with the default NChunks in devnet and let you know if it resolves the issue.

LexLuthr:
@ZenGround0 Pushed the last commit for setting price to 0 during devnet init.

transport/httptransport/http_transport.go
```diff
@@ -292,15 +292,15 @@ func (t *transfer) execute(ctx context.Context) error {
 
 	// Check if the control file exists and create it if it doesn't. Control file captures the number of chunks that the transfer has been started with.
 	// If the number of chunks changes half way through, the transfer should continue with the same chunking setting.
-	nChunks := t.nChunks
+	chunks := t.nChunks
```
snadrus, line 295: i kinda like nChunks as it indicates a number instead of the chunks themselves.

snadrus (approved):

transport/httptransport/http_transport.go
LexLuthr: It was clashing with the constant name.

LexLuthr: Renamed the constant

PR #1939, opened 2024-07-22
feat(gocql): add authentication support

strahe:
We are useing Yugabyte with [YCQL authentication](https://docs.yugabyte.com/preview/secure/enable-authentication/authentication-ycql/) enabled via the parameter `--use_cassandra_authentication` and [YSQL authentication](https://docs.yugabyte.com/preview/secure/enable-authentication/authentication-ysql/) enabled via `--ysql_enable_auth`. 

However, `boostd-data` is unable to utilize Cassandra authentication and can only connect using YSQL authorization through the connection string. 

This PR enables users to authenticate using [Cassandra credentials](https://github.com/yugabyte/gocql?tab=readme-ov-file#authentication).

### DIFF
--- a/cmd/migrate-lid/migrate_lid.go
+++ b/cmd/migrate-lid/migrate_lid.go
@@ -123,6 +123,14 @@ var migrateYugabyteDBCmd = &cli.Command{
 			Required: true,
 		},
 		&cli.StringFlag{
+			Name:  "username",
+			Usage: "yugabyte username to connect to over cassandra interface eg 'cassandra'",
+		},
+		&cli.StringFlag{
+			Name:  "password",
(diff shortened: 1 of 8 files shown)


LexLuthr:
@strahe This will need to be implemented on Curio side as well. Yb would be common for both. If Yb has auth enabled then we will need client auth on both sides.

strahe:
@LexLuthr Can you review this again? We really need this feature. As far as I know, YCQL is not used in Curio, so it does not affect Curio.

LexLuthr:
@strahe Have you tested this? Can you please add an auth test?

strahe:
@LexLuthr Tested and already in use in our production environment. Do you mean adding a test screenshot or a test case script?

LexLuthr:
Thanks for the PR

PR #1957, opened 2024-08-27
fix: boost allocate commands

strahe:
This PR fixes two issues with `boost allocation`:

1. When the number of entries in `--piece-file` exceeds the default batch size of 500, the `datacap` amount calculation is incorrect, leading to erroneous messages.

image

    
2. When sending multiple messages, the nonce value remains the same because the `datastore` in `SignAndPushToMpool` does not persist(is not shared), so only the first message can be successfully sent.

img_v3_02e5_bd602930-159f-4988-b951-2c25fa8b0dch

After:
img_v3_02e5_2efadf19-a862-46c9-94eb-b2f95ca92eeh

Tested in my calibnet environment.

### DIFF
--- a/cmd/boost/direct_deal.go
+++ b/cmd/boost/direct_deal.go
@@ -3,6 +3,8 @@ package main
 import (
 	"bufio"
 	"fmt"
+	"github.com/ipfs/go-datastore"
+	ds_sync "github.com/ipfs/go-datastore/sync"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


LexLuthr (changes requested):
Can you please also fix the datacap calculation for extend-claim command as well?

strahe:
> Can you please also fix the datacap calculation for extend-claim command as well?

The logic for calculating datacap in extend-claim should be correct.

https://github.com/filecoin-project/boost/blob/84f46db5f1e32cae27286d4588ea2d8ef4c2cc6a/cmd/boost/util/util.go#L370-L376

LexLuthr (approved):

PR #1967, opened 2024-09-26
feat: Curio market migration command

LexLuthr:


cmd/migrate-curio/migrate.go
```diff
@@ -0,0 +1,557 @@
...
+			// Mark deal added to harmonyDB
+			_, err = mdb.Exec(`INSERT INTO Deals (ID, DB, LID) VALUES (?, TRUE, FALSE) ON CONFLICT(ID) DO NOTHING`, deal.DealUuid.String())
+			if err != nil {
+				return fmt.Errorf("deal: %s: failed to mark deal migrated: %w", deal.DealUuid.String(), err)
+			}
+		}
+
+		if !b {
+			// Add LID details to pieceDeal in HarmonyDB
+			_, err = hdb.Exec(ctx, `SELECT process_piece_deal($1, $2, $3, $4, $5, $6, $7, $8, $9)`,
```
magik6k, line 289: I guess this will just reimport the piece, redoing LID/IPNI will be a separate small reindexing pipeline, right?

magik6k (approved):
Afaict this looks good

cmd/migrate-curio/migrate.go
LexLuthr: Yes. This is to ensure that there is no chance of missing metadata. Func process_piece_deal does not raise an exception if the piece is already present in the table. First time indexing task actually creates the indexes, we mark the piece as indexed and if another deal has the same piece then we skip the actual indexing next time around.

PR #2026, opened 2025-09-15
DDO with notifications spike

ZenGround0:
This is a spike branch for enabling manual testing.  Actual support for "DDO+" where smart contracts can receive notifications of sector prove commit will probably look similar.  But this PR doesn't do any of the hard cleanup stuff like sql migrations or thinking through exactly what the best api breaking change is.

### DIFF
--- a/cmd/boostd/import_direct_data.go
+++ b/cmd/boostd/import_direct_data.go
@@ -1,15 +1,18 @@
 package main
 
 import (
+	"encoding/hex"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	bcli "github.com/filecoin-project/boost/cli"
 	"github.com/filecoin-project/boost/storagemarket/types"
 	"github.com/filecoin-project/go-address"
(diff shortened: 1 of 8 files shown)


ZenGround0:
FYI this is currently failing with a problem in the esoteric polydawn cbor encoding library.  Probably related to the type in question partially implementing cbg (un)marshal.  Here is the error that Tippy hit when running in production: 

```
boostd import-direct --client-addr=t01011 --allocation-id=69984 --notify="t410fqpeeilhdvpabviipy77wsxs434vv4eclqrc5r4y:0xc0ffee" baga6ea4seaqnhnt3a6f4rrhvkcwcg6tenjufbcd22hadq73r5cdgmdaw5hwz4ma ~/4.car
Error: failed to execute direct data import: failed to insert direct deal entry to local db: failed to cbor dump DataActivationNotification: invalid state: value already consumed
```

When we pick this up first order of business is to fix that and probably tweak the test / add a new test.



FYI @warpfork your memory lives on and we miss you

PR #2075, opened 2026-04-23
update go 1.26, deps

LexLuthr:


go.mod
```diff
@@ -48,10 +48,10 @@ require (
 	github.com/graph-gophers/graphql-go v1.3.0
 	github.com/hannahhoward/go-pubsub v1.0.0
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/ipfs/boxo v0.36.0
+	github.com/ipfs/boxo v0.37.0
```
rvagg, line 51: 0.39 is out already, you could consider bumping to that, I don't know how painful it would be though

rvagg (approved):
`new()`, one of the best QoL improvements we've got lately

go.mod
LexLuthr: That requires pubsub upgrade which breaks Lotus and build fails.

PR #2132, opened 2026-09-19
feat: deprecate datacap for nv29 (FIP-0118 / Solstice)

### LINKED ISSUE #2106: [Solstice] datacap deprecation
FIP Draft PR: https://github.com/filecoin-project/FIPs/pull/1270

With nv29, the Solstice FIP proposes to end FIL+ and make all sectors 10x QAP by default. This task will track changes that boost may require.

In particular there will be a recommended method (https://github.com/filecoin-project/builtin-actors/issues/1763) for updating existing sectors to 10x QAP, and ideally this is a smooth experience.

snail8501:
## Deprecate datacap for nv29 (FIP-0118 / Solstice), plus two DoS fixes

This branch pins lotus to the Solstice line and makes Boost behave correctly once
the network reaches nv29, where datacap and verifreg are deprecated by FIP-0118.
It also carries two independent robustness fixes for client-controlled fields
that could crash `boostd`.

### 1. nv29 datacap deprecation (FIP-0118)

From nv29 no new allocations can be created and existing claims cannot be
extended, so every datacap entry point now fails fast with an explicit message
instead of bottoming out in an opaque "allocation not found" error.

- `boost direct-deal allocate` and `extend-claim` query `StateNetworkVersion`
  and refuse to build messages at nv29+.
- `list-allocations` is marked DEPRECATED in its usage string.
- `DirectDealsProvider.Accept` rejects DDO (FIL+ verified) deals at nv29+ via a
  new `isNv29OrAbove` helper.
- `DirectDealsProvider.Import` now returns as soon as `Accept` rejects a deal.

The `Import` early return also fixes a pre-existing nil pointer dereference:
`StateGetAllocation` returns `(nil, nil)` when an allocation does not exist
(which is why `Accept` has an explicit `allocation == nil` branch), but `Import`
ignored the rejection and dereferenced `allocation.Provider` on the next line.
Importing a direct deal with a non-existent allocation ID would panic. Reachable
only through the `perm:admin` `BoostDirectDeal` API, so operator-triggerable
rather than remotely exploitable.

### 2. Reject retrieval requests with a nil selector

The graphsync wire schema marks the request selector optional, so a peer can omit
it and it decodes to a nil `ipld.Node`. `dagcbor.Encode` dereferences it and
panics, taking down the daemon. Two guards:

- `handleRetrievalDeal` rejects the request before it is tracked as a transfer,
  so it is never registered as an active unpaid retrieval. The incoming-request
  hook then sees `intercept == false` and hands it to the legacy path.
- `requestValidator.acceptDeal` rejects it immediately before `dagcbor.Encode`,
  as defense in depth.

### 3. Reject zero-size deal transfers

`Transfer` is not covered by the client's signature and had no validation at all.
With `Transfer.Size == 0` the mismatch check in the HTTP transport is skipped (it
is guarded by `DealSize != 0`), `dealSize` falls through as 0 from the HEAD
response, `chunkSize` becomes 0, and `outputStats.Size() / chunkSize` panics in a
detached goroutine with no recover — crashing `boostd`. Fixed at three layers:

- `validateDealProposal` requires a positive `Transfer.Size` for online deals.
- `transfer.execute` rejects a non-positive `dealSize` before computing
  `chunkSize`.
- The transfer goroutine in `Execute` recovers panics and converts them into a
  transfer error, mirroring the existing `recover()` in
  `DirectDealsProvider.execDeal`.

### Dependencies

`lotus` is pinned to the `feat/solstice` branch
(`a1cca3114764f187a8780aa78055240f6c167aee`) via a `replace` directive; the
`require` line stays at `v1.36.2` so version tooling does not treat the module as
stale. `go-state-types` moves to a pseudo-version carrying `network.Version29`.
Routine bumps come along for the ride (boxo, go-libp2p-kad-dht, testify,
x/crypto, x/text).

The last commit regenerates `go.sum`, which had drifted when the dependency
versions changed: builds outside the workspace failed with `missing go.sum
entry`. Regenerated with `GOWORK=off go mod tidy` — workspace mode does not write
`go.sum`, which is how the drift went unnoticed.

### Testing

- `go mod tidy -diff` is clean and idempotent.
- Builds pass both in workspace mode and with `GOWORK=off` (the latter matters for
  downstream consumers and for Docker layers that copy only `go.mod`/`go.sum`).
- `go vet` clean on all touched packages.
- `go test ./transport/httptransport/` passes, including with `-race`.
- Verified by hand that a transfer with `DealSize=0` and a server advertising
  `Content-Length: 0` panics on `main` and returns a clean `deal size must be
  greater than zero` error on this branch.


storagemarket/direct_deals_provider.go
LexLuthr, line 612: This function need a pre/post NV29 branch. For deals which started sealing at NV28 but finish at NV29, this will cause the Boost to think deal failed because allocation can no longer be claimed.

transport/httptransport/http_transport.go
```diff
@@ -372,6 +383,15 @@ func (t *transfer) execute(ctx context.Context) error {
 		}
 	}
 
+	// A non-positive deal size would make the chunk size below zero and cause a
+	// divide-by-zero panic. The mismatch guard above is skipped when the deal
+	// size is zero, so reject it explicitly here.
```
LexLuthr:
```suggestion
	// Reject empty or invalid response sizes before calculating chunk sizes.
    // A zero size would make chunkSize zero, causing the division below to panic.
```

LexLuthr (changes requested):
More changes are required. Please do a deep dive and thorough testing. Add more tests. I am also not sure about freezing NV28 for itests. It does not add any benefit to keep those tests around post NV29. Some of these tests need actual behaviour modification instead of freezing them in place. Some of these issue were probably not caught because of that.
1. Boost deal defaults to verified. Let's flip that.
2. Remove extend-claim command post NV29.
3. A direct deal accepted before nv29 can finish sealing afterward without creating a claim. Piece doctor then clears its warning and skips checking whether its index or unsealed copy is missing.
4. Migration skips a direct deal when it cannot find its claim. For deals sealed after nv29, that can leave the deal record and retrieval mapping out of Curio even though the data was successfully sealed.
5. `boostd import-direct` needs the same early rejection.
6. A successfully sealed deal can display Proving(No claim found) after nv29.

storagemarket/direct_deals_provider.go
snail8501: Fixed. `resolveClaim` now dates the deal's **own sector** — `StateSectorGetInfo(...).Activation` against the nv29 upgrade height — and completes the deal when that sector activated at or after nv29, instead of reading the missing claim as a failure. Per-sector on purpose, not the chain's current version: a deal sealing from nv28 into nv29 has an activation before nv29, so it still owes its claim and still fails without one. Covered in `direct_deals_provider_test.go`.

transport/httptransport/http_transport.go
snail8501: Applied verbatim.

transport/httptransport/http_transport.go
```diff
@@ -372,6 +383,15 @@ func (t *transfer) execute(ctx context.Context) error {
...
+	if dealSize <= 0 {
+		return &httpError{
+			error: fmt.Errorf("deal size must be greater than zero, head: %d, dealInfo: %d", dealSize, t.dealInfo.DealSize),
+		}
```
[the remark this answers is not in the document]
snail8501: Added. Unknown deal size with a zero-length HEAD response is rejected with `deal size must be greater than zero` and explicitly not `panic`, so the check is pinned ahead of the division; the known-size case pins the order of the two size guards. With the `dealSize <= 0` branch disabled the first test fails with the actual divide-by-zero.

snail8501:
> More changes are required. Please do a deep dive and thorough testing. Add more tests. I am also not sure about freezing NV28 for itests. It does not add any benefit to keep those tests around post NV29. Some of these tests need actual behaviour modification instead of freezing them in place. Some of these issue were probably not caught because of that.
> 
> 1. Boost deal defaults to verified. Let's flip that.
> 2. Remove extend-claim command post NV29.
> 3. A direct deal accepted before nv29 can finish sealing afterward without creating a claim. Piece doctor then clears its warning and skips checking whether its index or unsealed copy is missing.
> 4. Migration skips a direct deal when it cannot find its claim. For deals sealed after nv29, that can leave the deal record and retrieval mapping out of Curio even though the data was successfully sealed.
> 5. `boostd import-direct` needs the same early rejection.
> 6. A successfully sealed deal can display Proving(No claim found) after nv29.

@LexLuthr All six are addressed. The nv28 freeze is gone — both itests now cross to nv29 while the sector seals instead of being pinned before it. --verified defaults to false, extend-claim is kept but gated (nv28 and nv29 have a compatibility window), the piece doctor no longer skips checks on a nv29 deal, migration no longer drops a deal whose claim is missing, import-direct rejects up front with the provider's own reason, and the UI no longer shows "(No claim found)" for a sector sealed at nv29+. All of it is covered by tests, and a new CI job runs the piece-doctor tests.

storagemarket/direct_deals_provider.go
```diff
@@ -119,6 +120,82 @@ func (ddp *DirectDealsProvider) Start(ctx context.Context) error {
...
+// value to compare against: callers test "activation >= height", so zero would
+// put every sector past the upgrade. Lotus uses a far-future epoch for an
+// unscheduled upgrade and a negative one for active-from-genesis.
+func Nv29UpgradeHeight(ctx context.Context, api v1api.FullNode) (abi.ChainEpoch, error) {
+	params, err := api.StateGetNetworkParams(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("getting network params: %w", err)
+	}
+
+	height := params.ForkUpgradeParams.UpgradeXxHeight
```
LexLuthr: Use the correct height with latest Lotus RC release

storagemarket/direct_deals_provider.go
```diff
@@ -587,6 +675,46 @@ func (ddp *DirectDealsProvider) watchSealingUpdates(entry *types.DirectDeal) *de
 	}
 }
 
+// resolveClaim reports whether a deal whose sector has reached a final sealing
+// state is finished. False with no error means a claim may still turn up; an
+// error is returned to the caller as-is.
+func (ddp *DirectDealsProvider) resolveClaim(entry *types.DirectDeal) (bool, *dealMakingError) {
```
LexLuthr: This function can no longer distinguish between a successful and failed snapDeal. You need to rethink the approach here.

cmd/migrate-curio/migrate.go
```diff
@@ -548,6 +549,38 @@ func migrateLegacyDeals(ctx context.Context, full v1api.FullNode, activeSectors
...
+func migratableDirectDeal(deal *types.DirectDeal, claim *verifreg9types.Claim, sealedAfterNv29, sectorAlive bool) (bool, string, error) {
+	// SectorID is zero until the piece reaches a sector.
+	if deal.Checkpoint < dealcheckpoints.AddedPiece {
+		return false, "the checkpoint is below add piece", nil
+	}
+
+	// The deal failed because its allocation could not be claimed. Past nv29
+	// there is no claim to make, so the failure is the expected one and the data
+	// is sealed and indexed: skipping it would strand it in Curio.
+	if deal.Err != "" && deal.Retry == types.DealRetryFatal && !sealedAfterNv29 {
```
LexLuthr: This basically means ignoring any Fatal error even unrelated to claims.

cmd/migrate-curio/migrate.go
```diff
@@ -561,42 +594,56 @@ func migrateDDODeals(ctx context.Context, full v1api.FullNode, activeSectors bit
...
+	}
+
+	// A query that fails leaves the deal undated rather than stopping the run:
+	// treating it as pre-nv29 skips one deal, which the next migration picks up.
+	sealedAfterNv29 := func(deal *types.DirectDeal) bool {
+		sealed, err := storagemarket.SealedAtOrAfterNv29(ctx, full, maddr, deal.SectorID)
+		if err != nil {
+			log.Warnw("could not date the deal against nv29; treating it as pre-nv29",
+				"deal", deal.ID.String(), "sector", deal.SectorID, "err", err)
+			return false
```
LexLuthr: You are basically skipping sector without telling the user. This should be an error not false. If there is an issue with migration, it should be surfaced.

LexLuthr (changes requested):
Following need to be NV29 aware.
1. sectorstatemgr.NewSectorStateMgr
2. Deal filters

storagemarket/direct_deals_provider.go
snail8501: Done. `lotus` is pinned to `v1.37.0-rc1` and the `feat/solstice` `replace` is gone, so `api.ForkUpgradeParams.UpgradeSolsticeHeight` is available in the released API. `nv29UpgradeHeight()` reads that height from the node: a zero height means a node too old to know about solstice and is an error, and negative heights — how devnets schedule forks — are allowed.

storagemarket/direct_deals_provider.go
snail8501: Agreed, and the activation dating is gone. Onboarding now requires a non-zero spacetime weight *and* the sealer reporting the deal's piece, instead of comparing an epoch. Without a claim, the sealer's own piece list decides: present completes the deal, absent fails it as `ErrPieceNotOnboarded`. A sealer that reports no pieces at all leaves the piece unverifiable rather than either completing or failing it, and a final sealing state that destroys the data is fatal as `ErrSectorSealingFailed` — so a failed `snapDeal` is caught by the piece list or by the sealing state, not by the epoch.

cmd/migrate-curio/migrate.go
snail8501: Fixed. `migratableDirectDeal` excuses exactly one fatal error, `ErrNoClaimFound`, and only for a piece onboarded at or after nv29. Every other fatal error skips the deal as before — `ErrPieceNotOnboarded` and `ErrPieceUnverifiable` included, since they say the data could not be established rather than that the claim is gone.

cmd/migrate-curio/migrate.go
snail8501: Fixed. `dateDealsAgainstNv29()` returns an error and `migrateDDODeals` aborts the run on it. The previous version read an unreachable chain as pre-nv29 and silently skipped the deals this change exists to rescue. It is now one `ChainHead` plus one `StateMinerSectors` for the whole run, rather than a chain read per deal.

snail8501:
@LexLuthr Both points from the second review are in 8e4b6a4.

**sectorstatemgr** — `refreshState` now reads both spacetime weights, via `SectorCarriesData()`, and needs no network-version check: the weights themselves are the correct signal. From nv29 every piece's spacetime lands in `VerifiedDealWeight` and `DealWeight` stays at zero, so reading `DealWeight` alone reported every post-nv29 sector as holding nothing. That is what made those sectors look empty both to the deal filters and to the "sectors with deals" list. Covered by `sectorstatemgr/sector_weights_test.go`.

**Deal filters** — `DealFilterParams` now carries `NetworkVersion`, and `BasicDealFilter` reads `ConsiderVerifiedStorageDeals` and `ConsiderUnverifiedStorageDeals` as a single switch at nv29, where no deal is verified. Read separately, a miner with verified deals left on and unverified deals off rejected every deal. A once-per-process warning names both configured values. External filters receive the version through the storage JSON document, now 2.3.0; the retrieval document deliberately stays at 2.2.0 rather than advertising a field it does not have, and an unset version is refused rather than read as nv0.

One correction to my earlier replies, now that 8e4b6a4 has replaced that approach: my answer on `direct_deals_provider.go:715` said the deal's own sector is dated by `Activation`. That is no longer what the code does — after the point you raised at line 681, an epoch comparison cannot separate a successful snap from a failed one, so a claimless deal is settled from the sealer's piece list instead. The same applies to the migration and UI items in my 09-22 summary: the conclusions still hold, but the mechanism is piece-based rather than activation-based.

